PR #13, opened 2016-10-14
Fix overflow for width==32

phil-opp:
If `width == 32` the left shift overflows, which causes a panic in debug mode. In order to fix this, we increase the mask size to `u64`.

[wrapping_shl](https://doc.rust-lang.org/nightly/std/primitive.u32.html#method.wrapping_shl) doesn't work here since it “removes any high-order bits of rhs that would cause the shift to exceed the bitwidth of the type”. So `1.wrapping_shl(32)` becomes `1.wrapping_shl(0)`, which leads to the wrong mask.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -227,8 +227,8 @@ pub fn gen_register_r(r: &Register, d: &Defaults) -> Vec<Tokens> {
             }
         } else {
             let width_ty = width.to_ty();
-            let mask: u32 = (1 << width) - 1;
-            let mask = Lit::Int(u64::from(mask), IntTy::Unsuffixed);
+            let mask: u64 = (1 << width) - 1;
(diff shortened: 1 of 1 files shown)


japaric:
Good catch! Thank you @phil-opp.

@homunkulus r+

homunkulus:
:pushpin: Commit 010f3c4 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit 010f3c4 with merge 010f3c4...

homunkulus:
:sunny: Test successful - [status-travis](https://travis-ci.org/japaric/svd2rust/builds/167663985)
Approved by: japaric
Pushing 010f3c4d0c97d00f6b977ddde7d216cc4d1ef087 to master...

PR #21, opened 2016-11-03
Enumerated values

brandonedens:
Add capability to emit rust interface to hardware memory map using enumerated values defined in the SVD file.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -49,8 +49,8 @@ pub fn gen_peripheral(p: &Peripheral, d: &Defaults) -> Vec<Tokens> {
         }
 
         let comment = &format!("0x{:02x} - {}",
-                     register.address_offset,
-                     respace(&register.description))[..];
+                               register.address_offset,
+                               respace(&register.description))[..];
 
         let field_ty = Ident::new(register.name.to_pascal_case());
         let field_name = Ident::new(register.name.to_snake_case());
@@ -62,8 +62,7 @@ pub fn gen_peripheral(p: &Peripheral, d: &Defaults) -> Vec<Tokens> {
         offset = register.address_offset +
                  register.size
             .or(d.size)
-            .expect(&format!("{:#?} has no `size` field", register)) /
-                 8;
+            .expect(&format!("{:#?} has no `size` field", register)) / 8;
     }
 
     let p_name = Ident::new(p.name.to_pascal_case());
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,10 +10,25 @@ use std::io;
 use std::io::Write;
 
 use quote::Tokens;
-use syn::*;
+use syn::{Ident, IntTy, Lit};
 
 use inflections::Inflect;
-use svd::{Access, Defaults, Peripheral, Register};
+use svd::{Access, Defaults, Field, Peripheral, Register};
+
+/// Trait that sanitizes name avoiding rust keywords and the like.
+trait SanitizeName {
+    /// Sanitize a name; avoiding Rust keywords and the like.
+    fn sanitize(&self) -> String;
+}
+
+impl SanitizeName for String {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -196,6 +210,41 @@ pub fn gen_register(r: &Register, d: &Defaults) -> Vec<Tokens> {
...
+            .map(|x| Ident::new(x.name.sanitize().to_pascal_case()))
+            .collect::<Vec<_>>()
+            .into_iter();
+        let vals = enum_vals.values
+            .iter()
+            .map(|x| x.value)
+            .collect::<Vec<_>>()
+            .into_iter();
+
+        if field.description.is_some() {
```
japaric: Should this use `if let`?

src/lib.rs
```diff
@@ -196,6 +210,41 @@ pub fn gen_register(r: &Register, d: &Defaults) -> Vec<Tokens> {
...
+            let comment = field.description.as_ref();
+            items.push(quote! {
+                #[doc = #comment]
```
japaric: Hmm, what this generate if `comment` is `None`?

src/lib.rs
```diff
@@ -253,16 +302,44 @@ pub fn gen_register_r(r: &Register, d: &Defaults) -> Vec<Tokens> {
...
+
+                quote! {
+                    pub fn #name(&self) -> #ty {
+                        const MASK: #bits_ty = #mask;
+                        const OFFSET: u8 = #offset;
+
+                        match ((self.bits >> OFFSET) & MASK) {
+                            _ => ::core::option::Option::None,
+                            #(#vals => ::core::option::Option::Some(#keys)),*
+                        }.unwrap()
```
japaric: `unwrap`? :scream_cat: 

This should return `Option` IMO. Most people will be using this like `reg.bit() == Some(Variant)` or via `if let Some(variant) = reg.bit()` rather than unwrapping what `reg.bit()` returns.

brandonedens: That unwrap() is there is to trip a run-time failure if the bitfield contains an invalid value; as in a value that is not one of the enumerated values which hopefully should never happen if the program / hardware is correct and the ARM manufacturer wrote a proper SVD. We could change that to _ => panic!("")

japaric: > hopefully should never happen
> if the program / hardware is correct
> if ... the ARM manufacturer wrote a proper SVD

Right, but it *could* happen. So, at least, we should give a more informative message.

Personally, I'm more worried about LLVM not being able to optimize away these panics. The Option alternative doesn't have that problem.

homunkulus:
:umbrella: The latest upstream changes (presumably 57e9e72) made this pull request unmergeable. Please resolve the merge conflicts.

japaric:
Thinking again about this: the enum could contain an extra, hidden variant to represent the "unexpected" register value. So, basically:

``` rust
enum State {
    /// A
    ValidStateA,
    /// B
    ValidStateB,
    /// C
    ValidStateC,
    #[doc(hidden)]
    __InvalidState__,
}
```

Then the `read` method would look more "normal"; no `Option` in the return value:

``` rust
// From the POV of the user, this can only return: ValidState{A,B,C}
fn read(&self) -> State { .. }
```

The downside is that the user would be able to create `__InvalidState__`.

If we are paranoid about the user misusing `__InvalidState__`, we could add a debug assertion to `write` to prevent the user from passing `__InvalidState__`. Likewise, we could add a debug assertion to the  `PartialEq` implementation as well.

In any case, I'm much more concerned about the ergonomics of this. If the user has to import a bunch of enums to be able to write/modify a register then the gain in type safe will come at the price of losing ergonomics. The trade off has to be assessed carefully. Or some way to avoid the loss in ergonomics must be devised.

japaric:
New attempt in #52. Closing this PR in favor of that one.

PR #22, opened 2016-11-04
Support derived_from through type aliases

phil-opp:
This PR generates a type alias for derived types. For example:

```rust
pub type I2c2 = ::i2c1::I2c1;
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,7 +16,18 @@ use inflections::Inflect;
 use svd::{Access, Defaults, Peripheral, Register};
 
 pub fn gen_peripheral(p: &Peripheral, d: &Defaults) -> Vec<Tokens> {
-    assert!(p.derived_from.is_none(), "DerivedFrom not supported");
+    if let Some(ref derived_from) = p.derived_from {
+        let p_name = Ident::new(p.name.to_pascal_case());
+        let derived_name = format!("::{}::{}",
+                                   derived_from.to_snake_case(),
+                                   derived_from.to_pascal_case());
+        let derived_ty = Ident::new(derived_name);
+
+        let type_alias = quote! {
+            pub type #p_name = #derived_ty;
+        };
+        return vec![type_alias];
+    }
 
     let mut items = vec![];
     let mut fields = vec![];


japaric:
Hmm, I'm not sure this is the right approach to support `derived_from`.

IIRC, `derived_from` is sometimes used to model "struct inheritance" such that the `derived_from` value is the base struct and more fields can be added to the peripheral that uses it.

Also, this path `::i2c1::I2c1` won't always work because the `i2c1` module may not be in the root of the crate and it may not even be named `i2c1` (it could be simply named `i2c`). All this because `svd2rust` doesn't, itself, build a tree of modules (on purpose, for maximum flexibility) nor picks the name of the modules it generates.

Finally, I don't see much value in having two different names for the exact same peripheral. Having two variables like `i2c1: &I2c` and `i2c2: &I2c` reflect better what's really going on: both are "instances" of the same peripheral.

phil-opp:
I'm not sure about this either. I just thought that it might be useful.

> IIRC, derived_from is sometimes used to model "struct inheritance" such that the derived_from value is the base struct and more fields can be added to the peripheral that uses it.

Hmm, I haven't seen any instance of additional fields so far. Would it be better to create a newtype wrapper with a public inner field? Or just copy the whole struct with all of its methods?

> Also, this path ::i2c1::I2c1 won't always work because the i2c1 module may not be in the root of the crate and it may not even be named i2c1 (it could be simply named i2c).

Good point. I completely forgot that I added the module generation myself :D.

> Finally, I don't see much value in having two different names for the exact same peripheral. Having two variables like i2c1: &I2c and i2c2: &I2c reflect better what's really going on: both are "instances" of the same peripheral.

I agree. However, these names come directly from the SVD file. And I prefer e.g.

```rust
struct i2cs {
    pub i2c1: &'static mut I2c1,
    pub i2c2: &'static mut I2c2,
}
```

over

```rust
struct i2cs {
    pub i2c1: &'static mut I2c1,
    pub i2c2: &'static mut I2c1,
}
```

The second case looks like a typo at first sight.

japaric:
> Would it be better to create a newtype wrapper with a public inner field? Or just copy the whole struct with all of its methods?

I'd probably go for:

``` rust
struct DerivedPeripheral {
    base: BasePeripheral,
    pub reg1: Reg1,
    ..
}

impl Deref for DerivedPeripheral {
    type Target = BasePeripheral;

    ..
}

impl DerefMut for DerivedPeripheral { .. }
```

even though it's considered an anti-pattern, I think. But can't say for sure without looking at an actual SVD that contains this kind of inheritance.

> And I prefer e.g.

If I ever wrote that, I'd prefer:

``` rust
struct i2cs {
    pub i2c1: &'static mut I2c,
    pub i2c2: &'static mut I2c,
}
```

but I always do renames like `I2c1` to `I2c` right after calling `svd2rust`.

phil-opp:
> I always do renames like I2c1 to I2c right after calling svd2rust

I try to auto-generate crates for various boards from the svd files, so I want to avoid any manual steps.

---

Closing, since this PR doesn't really makes sense without a module structure.

PR #31, opened 2016-11-20
Make write() take a mutable reference, to match modify() signature.

wesen:


japaric:
This `write`, that takes `&self`, is a method only available in *write only* registers. I *thought* this *was* memory safe (i.e. shouldn't require `unsafe`) because you can think of it as a "stateless" `static` variable so you *can't* cause data races with it (or so I thought). But I'm not sure anymore. There is some discussion in japaric/f3#59. I'd like to settle when it's safe (perhaps never?) to manipulate memory mapped registers before doing any change here as any change in that regard would be a breaking change.

homunkulus:
:umbrella: The latest upstream changes (presumably 4522caa) made this pull request unmergeable. Please resolve the merge conflicts.

wesen:
I think this is moot, and just going to close it.

PR #36, opened 2016-11-29
gen_register_w: always cast the reset value to the correct type

istankovic:
It may happen that a register (e.g. the WDT CTRL register of Atmel ATSAMD21E17)
does not have the reset value specified.  In such cases the default
value 0u32 will be used.  However, if the register width is not 32 (as
is the case with the above mentioned WDT CTRL), rustc will complain.

So always cast the default value to the type that corresponds to the
register's width.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -365,13 +365,14 @@ pub fn gen_register(r: &Register, d: &Defaults) -> Vec<Tokens> {
         .expect(&format!("{:#?} has no `size` field", r))
         .to_ty();
     let access = r.access.unwrap_or_else(|| {
-        let fields = r.fields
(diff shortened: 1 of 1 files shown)


japaric:
@istankovic Thanks!

@homunkulus r+

homunkulus:
:pushpin: Commit 0d63b0e has been approved by `japaric`

homunkulus:
:hourglass: Testing commit 0d63b0e with merge 0d63b0e...

homunkulus:
:broken_heart: Test failed - [status-appveyor](https://ci.appveyor.com/project/japaric/svd2rust/build/1.0.19)

PR #38, opened 2016-12-08
Implement naive `derivedFrom` for peripherals

grossws:
This implementation resolves `derivedFrom` reference and produces `merged_peripheral` which has:
- `name` and `base_address` from target peripheral,
- `derived_from` set to None, 
- all other fields are taken from base peripheral if not present in target one.


src/main.rs
```diff
@@ -41,20 +41,40 @@ fn main() {
...
+        .collect::<Vec<_>>()
+        .join("\n\n")
+}
+
+fn merge(p: &svd::Peripheral, bp: &svd::Peripheral) -> svd::Peripheral {
+    svd::Peripheral {
+        name: p.name.clone(),
+        base_address: p.base_address,
+        derived_from: None,
+        group_name: p.group_name.clone().or(bp.group_name.clone()),
```
japaric: Could you change the `or` to `or_else` to avoid an unnecessary clone when e.g. `p.group_name` is `Some`.

src/main.rs
```diff
@@ -41,20 +41,40 @@ fn main() {
...
+        description: p.description.clone().or(bp.description.clone()),
+        interrupt: p.interrupt.clone().or(bp.interrupt.clone()),
+        registers: p.registers.clone().or(bp.registers.clone()),
```
japaric: Shouldn't this merge both `Vec`s?

grossws: I think, we shouldn't merge them since [documentation](http://www.keil.com/pack/doc/CMSIS/SVD/html/elem_peripherals.html#elem_peripheral) says that values are inherited:
> Specify the name of a peripheral from which this peripheral will be derived. Values are inherited from the derived peripheral. Specified elements override inherited values.

grossws:
@japaric, I'll try to refactor and add a test to `ci/run.sh` this evening.

src/main.rs
japaric: Alright. We should check that the implementations don't actually use deriveFrom to implemen inheritance ...

japaric:
Also, if the deriveFrom peripheral is equivalent to the base one then we should just create a type alias of the base peripheral (`type Derived = Base`) instead of creating a new struct.

grossws:
It's a good idea but requires refactoring of code generation (and user should be aware that he should also generate all structs for base peripheral), so I've implemented it in just quick & dirty way to avoid modifying `src/lib.ru` since branch `auto` was about to merge. But I will consider it for next iteration if `src/lib.rs` isn't going to change much to avoid maintaining quite a big patch.

Maybe `svd2rust` should have some kind of *standalone* mode when it generates all consts with base addresses and all structs in one pass to give user consistent result.

phil-opp:
> @phil-opp I think at least one of you also wanted/needed this feature. Have any thoughts on this approach?

Hmm, I'm not sure… The advantage of a new struct is that `svd2rust` outputs something usable for all peripherals.  For example, when I do a `svd2rust -i STM32F30x.svd gpioi` I get a struct that I can use directly, even if `gpioi` is derived from `gpiod`. On the other hand, it's not really a good idea to have different types for the identical `gpioi` and `gpiod` peripherals, since it makes writing generic code much harder (e.g. a method that operates on any `gpio` port). 


> Maybe svd2rust should have some kind of standalone mode when it generates all consts with base addresses and all structs in one pass to give user consistent result.

I tried to create something like that in [embed-rs/svd_board](https://github.com/embed-rs/svd_board) (based on svd2rust). It auto-generates a `Hardware` struct that contains all registers from the svd file. There's also a `hw` function that returns a `Hardware` struct filled with the correct pointers.

grossws:
> On the other hand, it's not really a good idea to have different types for the identical gpioi and gpiod peripherals, since it makes writing generic code much harder (e.g. a method that operates on any gpio port).

It seems to be simple to extract common functionality when we are speaking of GPIOs which are very alike although there should be different structs. For example `STM32F40x.svd` have different reset values for `GPIOA`, `GPIOB` and all other `GPIOx` `MODER`s. For such cases one could provide common interface via trait.

It becomes a lot more difficult for common trait extraction if we are talking of timers. For example `STM32F407xx` have:
- two 16-bit advanced timers (`TIM1`, `TIM8`) with complementary outputs, up/down/up+down counting, DMA requests and 4 CC (capture/compare) channels,
- two 32-bit general-purpose (`TIM2`, `TIM5`) with up/down/up+down counting, DMA req and 4 CC channels; but `TIM2` special internal trigger output from `TIM8` and USB OTG and `TIM5` has special timer input remapping;
- two 16-bit general-purpose (`TIM3`, `TIM4`) with up/down/up+down counting, 4 CC channels;
- two 16-bit general-purpose (`TIM9`, `TIM12`) with up counting and 2 CC channels but without DMA req (and they do differ by maximum clock frequency);
- four another 16-bit general-purpose (`TIM10`, `TIM11`, `TIM13`, `TIM14`) with up counting and 1 CC channel and without DMA req; also they have different max frequencies for input clock;
- two 16-bit basic timers (`TIM6`, `TIM7`) with up counting, DMA req and without CC channels at all.

I truly cannot imagine set of automatically generated traits to support at least half of these functions only from SVD file even for very narrow class like STM32-only. And it's totally out of SVD scope.

And in case of not generating (or writting) such traits we will anyway generate different structs for say `TIM1` and `TIM6` but `TIM1` obviously could be used in any place where `TIM6` can be. So there should be trait for basic timer, I guess. But what traits could be generic and flexible enough to describe that it can be used as source to start ADC?

Also I think that such traits should be defined in separate crate and their implementations should be defined in repos like [japaric/f3](https://github.com/japaric/f3).

src/main.rs
grossws: As I see in `STM32F30x.svd` `derivedFrom` is used only in two variants:
- full copy of base peripheral (like `GPIO`s) with different `name` and `baseAddress`);
- copy with overrided interrupts (like `DMA2`).

And merging `interrupts` in second case makes no sense.

japaric:
> The advantage of a new struct is that svd2rust outputs something usable for all peripherals.

I'd be OK with e.g. `svd2rust -i $F gpioc` just outputting `type Gpioc = Gpiob` because right now it just errors :-).

> it's not really a good idea to have different types for the identical gpioi and gpiod peripherals

I agree 100%.

> I tried to create something like that in embed-rs/svd_board ..

I'd like to have an option to generate code for ALL the peripherals in svd2rust in the future, when it's able to handle both derivedFrom and register arrays.

>  For example STM32F40x.svd have different reset values for GPIOA, GPIOB

Those should have different types.

Two peripherals should only have the same type if the only different between them is name and baseAddress.

> For such cases one could provide common interface via trait.

I think generating such traits is out of scope for svd2rust.

src/main.rs
japaric: Would it make sense to add an `assert` to check that one of `base.registers` or `derived.registers` is empty? The assertion should fail if both `base.registers` and `derived.registers` are not empty `Vec`s because that could indicate an attempt at inheritance.

> And merging interrupts in second case makes no sense.

(yeah, I imagine. That's why I only asked about `registers` :-)

grossws: I've never seen `SVD` with such a trick. I rechecked it with `SVD`s from [posborne/cmsis-svd](https://github.com/posborne/cmsis-svd/tree/c2acb8a89e00b497711251ac46d43cb805f265c1/data) and none of them overrides registers when `derivedFrom` is used.

Script which I used to check it:
```bash
for f in $(find -name '*.svd') ; do 
  fs=${f##*/}
  cat $f | xmlstarlet sel -t -m '//peripheral[@derivedFrom]' \
    -i 'count(registers/register) > 0' \
    -o "$fs: " -v 'name' -o ' ' -c 'count(registers/register)' -n
done
```

So, such assert seems to be good precaution against potential misuse and in case such `SVD`s really exists `svd2rust` user would file an issue.

grossws:
> I'd be OK with e.g. `svd2rust -i $F gpioc` just outputting `type Gpioc = Gpiob` because right now it just errors :-).

As I see user usually generates code for `GPIO`s and wraps it to different modules to separate different registers with same names. E.g. `Moder` for `GPIOA` and `GPIOB` having different reset values. 
In your example user should import `Gpiob` to such module manually because `svd2rust` doesn't know how user wraps its output.

japaric:
> As I see user usually generates code for GPIOs and wraps it to different modules to separate different registers with same names.

You are right. Then let's punt this type alias stuff until when svd2rust has a "generate everything" feature. Let's just "duplicate" the structs of equivalent peripherals for now. As the users are mainly using this manually or via a shell script they can just *avoid* generating duplicated structs and do the type aliasing manually.

src/main.rs
grossws: Added an `assert!`

japaric:
Thanks, @grossws!

@homunkulus r+

homunkulus:
:pushpin: Commit e94b369 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit e94b369 with merge e94b369...

homunkulus:
:sunny: Test successful - [status-appveyor](https://ci.appveyor.com/project/japaric/svd2rust/build/1.0.22), [status-travis](https://travis-ci.org/japaric/svd2rust/builds/184041107)
Approved by: japaric
Pushing e94b369d73b9d8fdf9b8e0b54cf9fe1fe2c9291b to master...

PR #53, opened 2017-01-29
enum based API (alternative write API)

japaric:
This PR is like #52 but with an alternative write API.

The generated write API looks like this:

``` rust
#[derive(Clone, Copy)]
#[repr(C)]
pub struct DirW {
    bits: u32,
}

pub struct DirWPin0<'a> {
    register: &'a mut DirW,
}

impl<'a> DirWPin0<'a> {
    pub fn input(self) -> &'a mut DirW {
        const MASK: u32 = 1;
        const OFFSET: u8 = 0u8;
        self.register.bits &= !((MASK as u32) << OFFSET);
        self.register.bits |= 0u32 << OFFSET;
        self.register
    }

    pub fn output(self) -> &'a mut DirW {
        const MASK: u32 = 1;
        const OFFSET: u8 = 0u8;
        self.register.bits &= !((MASK as u32) << OFFSET);
        self.register.bits |= 1u32 << OFFSET;
        self.register
    }
}

impl DirW {
    #[doc = "Bit 0 - Pin 0."]
    pub fn pin0(&mut self) -> DirWPin0 {
        DirWPin0 { register: self }
    }
    
    ..
}
```

Instead of exposing a method that takes an enum to modify a bitfield of a
register (`register.bitfield(Enum::VariantA)`), like in #52, this API exposes a
method that returns a proxy struct whose methods let you pick a value for the
bitfield (`register.bitfield().variant_a()`) *and* this method returns back the
register so you don't lose the chain-ability:
`register.bitfield1().variant_a().bitfield2().variant_b()`. 

[Here's](https://github.com/japaric/nrf51822/pull/5/files#diff-f5c08210e017893c364670505e0738f9) the blinky example from #52 ported to this alternative write API.

cc @brandonedens

PR #56, opened 2017-02-06
allow dead_code and missing_docs in generated output

crawford:
Fixes https://github.com/japaric/svd2rust/issues/18


japaric:
Thanks for the PR, @crawford.

If the generated code produces dead_code warnings, that indicates a bug in svd2rust or that svd2rust is being misused (e.g. you are not exposing the API as public).

And I'd rather fix the missing_docs warnings by actually documenting better the API rather than just sprinkling allow(missing_docs) here and there.

The docs generation will be revised in #53 so let's revisit this later.

crawford:
Okay, I'll have to double check my setup to figure out why I get all of those warnings.

Looking forward to #53.

PR #72, opened 2017-04-18
Feature: --rustfmt output.

archaelus:
Adds a new flag ‘-r’ ‘--rustfmt’ that pretty-prints the generated rust output with the rustfmt library. This makes the output human readable and easy to compile.


archaelus:
Presumably if merged, there would need to be README/doc updates. I'm using it as `svd2rust -r -i somefile.svd > $halcrate/src/lib.rs` which then builds fine with cargo build.

japaric:
Thanks for the PR, @archaelus.

I feel that adding rustfmt *right now* is more of a disadvantage than an advantage. For one, rustfmt is still undergoing development and new options are being added to it as part of the RFC process to define the default / official Rust style. That means that if we depend on rustfmt right now we would have to bump svd2rust everytime rustfmt has a new release; otherwise, people would still have to run (a more updated) rustfmt on the svd2rust output. OTOH, the advantage of this PR is that you don't have to install or run rustfmt but people are likely to already have rustfmt installed as they probably use it for others projects and, in the future, it's likely that rustfmt will become part of the "Rust platform" distribution so it will be installed by default along with rustc and the RLS. As for not having to rustfmt yourself, I don't see much advantage to that since it's easy to pipe svd2rust to rustfmt so you only save yourself some keystrokes on a not too common operation.

Thanks again but I won't take this PR. We can reevaluate once rustfmt hits version 1.0 and the default style has been decided.

archaelus:
That sounds fair and reasonable - thanks for explaining.

To make a pitch for something like this as a future addition to svd2rust, it was surprising to me initially that there was an expected post-processing step - rust code that you would commit to a hardware crate is probably a default that many people will expect. One other point in favour of integration is that if you need any changes from rustfmt's defaults (using it as a library I needed to tweak a couple of things like line length not being a hard error), then integration is going to save people having to rediscover that independently and setup a rustfmt config file.

PR #79, opened 2017-04-29
If <resetValue> is not present at any level, assume it is zero

### LINKED ISSUE #75: Reset value should not be required
The SVD spec does not require reset values to be present. This happens in real data, e.g. on TM4C series it looks like all writable registers are always initialized to 0.

whitequark:
CMSIS-SVD specification does not require <resetValue> to be present,
so failing when it isn't is wrong. In general, it seems safe to
assume that it's just zero.

Fixes #75.

### DIFF
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -494,11 +494,8 @@ pub fn register(
 
         let rv = register.reset_value
             .or(defs.reset_value)
-            .map(|rv| util::unsuffixed(u64(rv)))
-            .ok_or_else(|| {
-                            format!("Register {} has no reset value",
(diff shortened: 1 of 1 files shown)


Emilgardis:
This is incorrect.

http://www.keil.com/pack/doc/CMSIS/SVD/html/elem_registers.html#elem_register

> If the elements `<size>`, `<access>`, `<resetValue>`, and `<resetMask>` have not been specified on a higher level, then these elements are mandatory on register level.

whitequark:
@Emilgardis Thanks, I missed that sentence.

PR #104, opened 2017-06-05
full test suite

japaric:
let's see if this times out

PR #106, opened 2017-06-05
make sure that the register size is always a power of 2 >= 8

### LINKED ISSUE #102: register of size 1?
MB9AF10xN.svd

Contents:

``` xml
        <register>
          <name>WDG_RIS</name>
          <description>Hardware Watchdog Timer Interrupt Status Register</description>
          <addressOffset>0x10</addressOffset>
          <size>1</size>
          <access>read-only</access>
          <resetValue>0xFF</resetValue>
          <resetMask>0x00</resetMask>
          <!-- FIELDS -->
          <fields>
            <!-- FIELD "RIS" -->
            <field>
              <name>RIS</name>
              <description>Hardware watchdog interrupt status bit</description>
              <lsb>0</lsb>
              <msb>0</msb>
              <access>read-only</access>
            </field>
          </fields>
        </register>
```

Produces:

``` rust
        impl R {
            # [ doc = r" Value of the register as raw bits" ]
            # [ inline ( always ) ]
            pub fn bits(&self) -> bool {
                self.bits
            }
            # [ doc = "Bit 0 - Hardware watchdog interrupt status bit" ]
            # [ inline ( always ) ]
            pub fn ris(&self) -> RISR {
                let bits = {
                    const MASK: bool = true;
                    const OFFSET: u8 = 0;
                    ((self.bits >> OFFSET) & MASK as bool) != 0
                };
                RISR { bits }
            }
        }
```

Compiler error:

``` console
error[E0369]: binary operation `>>` cannot be applied to type `bool`
    --> src/lib.rs:6496:22
     |
6496 |                     ((self.bits >> OFFSET) & MASK as bool) != 0
     |                      ^^^^^^^^^^^^^^^^^^^^^
     |
     = note: an implementation of `std::ops::Shr` might be missing for `bool`

error: aborting due to previous error(s)
```

The problem I see that the SVD file declares the register has having a size of 1. I'm not sure if that allowed by the SVD format (the website appears to be down ATM) but I'd expect it to only allow sizes multiple of 8 (~~bytes~~ bits). Note that we are talking about *registers*; bitfields can have any size (smaller than the size of their register).

Anyhow the current implementation maps a size of 1 to a `bool` so the `bits` field of `R` is `bool` which can be shifted to the right by `OFFSET` thus the compiler error.

cc @whitequark fallout from #84 

EDIT I meant *bits* not *bytes*.

japaric:
fixes #102

cc @whitequark
waiting for #104 to make sure this doesn't break other stuff

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -445,96 +445,96 @@ main() {
 
         Spansion)
             # OK
-            test_svd MB9AF12xK
-            test_svd MB9AF12xL
-            test_svd MB9AF42xK
-            test_svd MB9AF42xL
-            test_svd MB9BF12xJ
-            test_svd MB9BF12xS
-            test_svd MB9BF12xT
-            test_svd MB9BF16xx
-            test_svd MB9BF32xS
-            test_svd MB9BF32xT
-            test_svd MB9BF36xx
-            test_svd MB9BF42xS
-            test_svd MB9BF42xT
-            test_svd MB9BF46xx
-            test_svd MB9BF52xS
-            test_svd MB9BF52xT
-            test_svd MB9BF56xx
+            # test_svd MB9AF12xK
+            # test_svd MB9AF12xL
+            # test_svd MB9AF42xK
+            # test_svd MB9AF42xL
+            # test_svd MB9BF12xJ
+            # test_svd MB9BF12xS
+            # test_svd MB9BF12xT
+            # test_svd MB9BF16xx
+            # test_svd MB9BF32xS
+            # test_svd MB9BF32xT
+            # test_svd MB9BF36xx
+            # test_svd MB9BF42xS
+            # test_svd MB9BF42xT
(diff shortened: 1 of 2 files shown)


japaric:
@homunkulus r+

homunkulus:
:pushpin: Commit f09b542 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit f09b542 with merge f09b542...

homunkulus:
:sunny: Test successful - [status-appveyor](https://ci.appveyor.com/project/japaric/svd2rust/build/1.0.96), [status-travis](https://travis-ci.org/japaric/svd2rust/builds/239791040?utm_source=github_status&utm_medium=notification)
Approved by: japaric
Pushing f09b542a4cb40339f87b2a88c16dcf3f2f2ee601 to master...

PR #121, opened 2017-07-05
Register array in RegisterBlock

### LINKED ISSUE #114: Merge registers with the same type into an array
The code generated for the GPIO peripheral of nrf51 micros looks like this:

``` rust
pub struct Gpio {
    pub out: Out,
    pub outset: Outset,
    pub outclr: Outclr,
    pub in_: In,
    pub dir: Dir,
    pub dirset: Dirset,
    pub dirclr: Dirclr,
    pub pin_cnf0: PinCnf,
    pub pin_cnf1: PinCnf,
    pub pin_cnf2: PinCnf,
    pub pin_cnf3: PinCnf,
    pub pin_cnf4: PinCnf,
    pub pin_cnf5: PinCnf,
    pub pin_cnf6: PinCnf,
    pub pin_cnf7: PinCnf,
    pub pin_cnf8: PinCnf,
    pub pin_cnf9: PinCnf,
    pub pin_cnf10: PinCnf,
    pub pin_cnf11: PinCnf,
    pub pin_cnf12: PinCnf,
    pub pin_cnf13: PinCnf,
    pub pin_cnf14: PinCnf,
    pub pin_cnf15: PinCnf,
    pub pin_cnf16: PinCnf,
    pub pin_cnf17: PinCnf,
    pub pin_cnf18: PinCnf,
    pub pin_cnf19: PinCnf,
    pub pin_cnf20: PinCnf,
    pub pin_cnf21: PinCnf,
    pub pin_cnf22: PinCnf,
    pub pin_cnf23: PinCnf,
    pub pin_cnf24: PinCnf,
    pub pin_cnf25: PinCnf,
    pub pin_cnf26: PinCnf,
    pub pin_cnf27: PinCnf,
    pub pin_cnf28: PinCnf,
    pub pin_cnf29: PinCnf,
    pub pin_cnf30: PinCnf,
    pub pin_cnf31: PinCnf,
    // some fields omitted
}
```

Instead, it should generate:

``` rust
pub struct Gpio {
    pub out: Out,
    pub outset: Outset,
    pub outclr: Outclr,
    pub in_: In,
    pub dir: Dir,
    pub dirset: Dirset,
    pub dirclr: Dirclr,
    pub pin_cnf: [PinCnf; 32],
    // some fields omitted
}
```
The SVD file contains information about PinCnf being an array but the current svd2rust implementation simply *unrolls* the array into several struct fields instead of creating a single struct field that represents the array of registers.

kjetilkjeka:
I guess this pull request is dropping a bit out of the sky, but I realized I needed the feature sooner rather than later so decided to just go for it. When it's merged it should fix #114 

Some refactoring was done to allow for integration register arrays, the most significant are
- expand function only takes 1 register instead of slice. This means that the function has to be called for every register that should be expanded. This was done so a register can be conditionaly expanded depending on if it's possible to convert to a Rust array or not.
- register_to_rust method was factored out from register_block. Mainly so register block wouldn't grow very large with lots of nested control flow.
- ExpandedRegister was removed, the functions only use Register struct now. This was because if not it would require conversion functions when registers was not expanded. The struct would also need changing as it would only have to support arrays.
- Since the ExpandedRegisters was dropped the expand function now writes ```%s(i)``` (where i is the index) to the names of the expanded registers. This is used later on to resolve name and type.
- regex is a dependecy, this is used to replace ```%s(i)``` with the index. Adding this dependecy might not be desired? and if not I can reimplement it with some logic that iterates over the string.

I've tested with [S32K144.svd](https://github.com/kjetilkjeka/s32k144.rs/blob/master/src/S32K144.svd) and it seems to work perfectly with that file.

As mentioned in #114 i do hope that the method of using svd array information instead of unrolling arrays after is acceptable.

### DIFF
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -466,7 +466,25 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
     let mut i = 0;
     // offset from the base address, in bytes
     let mut offset = 0;
-    for register in util::sort_by_offset(util::expand(registers)) {
+    let mut registers_expanded = vec![];
+
+    for register in registers {
+        match *register {
+            Register::Single(ref _into) => registers_expanded.push(register.clone()),
+            Register::Array(ref _into, ref array_info) => {
+                let array_convertable = false; // TODO: write this condition
+
+                if array_convertable {
+                    registers_expanded.push(register.clone());
+                } else {
+                    registers_expanded.append(&mut util::expand(register));
+                }
+            },
+        }
+    }
+
+    
+    for register in util::sort_by_offset(registers_expanded) {
         let pad = if let Some(pad) = register.address_offset.checked_sub(offset) {
             pad
         } else {
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -2,7 +2,6 @@ use std::collections::HashMap;
 use std::io::{self, Write};
 
 use cast::u64;
-use either::Either;
 use quote::Tokens;
 use svd::{Access, BitRange, Defaults, Device, EnumeratedValues, Field,
           Peripheral, Register, Usage, WriteConstraint};
@@ -468,7 +467,7 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
     // offset from the base address, in bytes
     let mut offset = 0;
     for register in util::sort_by_offset(util::expand(registers)) {
-        let pad = if let Some(pad) = register.offset.checked_sub(offset) {
+        let pad = if let Some(pad) = register.address_offset.checked_sub(offset) {
             pad
         } else {
             writeln!(
@@ -476,7 +475,7 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
                 "WARNING {} overlaps with another register at offset {}. \
                  Ignoring.",
                 register.name,
-                register.offset
+                register.address_offset
             ).ok();
             continue;
         };
@@ -492,27 +491,27 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
 
         let comment = &format!(
             "0x{:02x} - {}",
-            register.offset,
-            util::respace(&register.info.description)
+            register.address_offset,
+            util::respace(&register.description)
         )
             [..];
 
-        let rty = match register.ty {
-            Either::Left(ref ty) => Ident::from(&**ty),
-            Either::Right(ref ty) => Ident::from(&***ty),
-        };
-        let reg_name = Ident::new(&*register.name.to_sanitized_snake_case());
-        fields.push(quote! {
-            #[doc = #comment]
-            pub #reg_name : #rty,
-        });
+        let rty = Ident::new(register.name.split('[').next().unwrap().to_sanitized_upper_case());
+        let reg_name = Ident::new(register.name.to_sanitized_snake_case());
+        
+        fields.push(
+            quote! {
+                #[doc = #comment]
+                pub #reg_name : #rty,
+            });
 
-        offset = register.offset +
-                 register
-                     .info
-                     .size
-                     .or(defs.size)
-                     .ok_or_else(
+        offset =
+            register.address_offset
+            +
+            register
+            .size
+            .or(defs.size)
+            .ok_or_else(
             || {
                 format!("Register {} has no `size` field", register.name)
             },
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -494,6 +494,8 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
                 }
             ),
             Register::Array(ref info, ref array_info) => {
+                let sequential_adresses = register_size == array_info.dim_increment*BITS_PER_BYTE;
+                
                 let index_convertible = if let Some(ref indexes) = array_info.dim_index {
                     let mut index_iter = indexes.iter();
                     let mut previous = 0;
@@ -529,9 +531,7 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
                     false
                 };
 
-                let packed_registers = array_info.dim_increment == 4;
-                
-                let array_convertible = index_convertible && packed_registers;
+                let array_convertible = index_convertible && sequential_adresses;
 
                 if array_convertible {
                     registers_expanded.push(
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -472,7 +472,7 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
         match *register {
             Register::Single(ref _into) => registers_expanded.push(register.clone()),
             Register::Array(ref _into, ref array_info) => {
-                let array_convertable = if let Some(ref indexes) = array_info.dim_index {
+                let index_convertible = if let Some(ref indexes) = array_info.dim_index {
                     let mut index_iter = indexes.iter();
                     let mut previous = 0;
 
@@ -483,15 +483,15 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
                                                                         })
                     };
 
-                    let mut array_convertable_temp = match get_number_option(index_iter.next()) {
+                    let mut index_convertible_temp = match get_number_option(index_iter.next()) {
                         Some(0) => true,
                         _ => false,
                     };
                     
                     for element in index_iter {
-                        if !array_convertable_temp { break; }
+                        if !index_convertible_temp { break; }
 
-                        array_convertable_temp = match get_number_option(Some(element)) {
+                        index_convertible_temp = match get_number_option(Some(element)) {
                             Some(i) => i == previous+1,
                             _ => false,
                         };
@@ -502,13 +502,16 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
                         };
                     }
                     
-                    array_convertable_temp
+                    index_convertible_temp
                 } else {
                     false
                 };
-                
 
-                if array_convertable {
+                let packed_registers = array_info.dim_increment == 4;
+                
+                let array_convertible = index_convertible && packed_registers;
+
+                if array_convertible {
                     registers_expanded.push(register.clone());
                 } else {
                     registers_expanded.append(&mut util::expand(register));
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -467,7 +467,7 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
     let mut i = 0;
     // offset from the base address, in bytes
     let mut offset = 0;
-    for register in util::expand(registers) {
+    for register in util::sort_by_offset(util::expand(registers)) {
         let pad = if let Some(pad) = register.offset.checked_sub(offset) {
             pad
         } else {
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -568,7 +568,7 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
         let comment = &format!(
             "0x{:02x} - {}",
             register.offset,
-            register.description,
+            util::respace(&register.description),
         )
             [..];
         
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -472,7 +472,41 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
         match *register {
             Register::Single(ref _into) => registers_expanded.push(register.clone()),
             Register::Array(ref _into, ref array_info) => {
-                let array_convertable = false; // TODO: write this condition
+                let array_convertable = if let Some(ref indexes) = array_info.dim_index {
+                    let mut index_iter = indexes.iter();
+                    let mut previous = 0;
+
+                    let get_number_option = |element_option: Option<&String>| -> Option<usize> {element_option
+                                                                        .and_then(|element| match element.parse::<usize>() {
+                                                                            Ok(i) => Some(i),
+                                                                            _ => None,
+                                                                        })
+                    };
+
+                    let mut array_convertable_temp = match get_number_option(index_iter.next()) {
+                        Some(0) => true,
+                        _ => false,
+                    };
+                    
+                    for element in index_iter {
+                        if !array_convertable_temp { break; }
+
+                        array_convertable_temp = match get_number_option(Some(element)) {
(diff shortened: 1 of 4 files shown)


kjetilkjeka:
Also have a look at line 399 at generate.rs

I guess that is only valid as long as there are 4 byte (32 bit) registers. However I'm not sure how it can be generalized over register lengths?

src/generate.rs
```diff
@@ -356,16 +355,71 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
...
+    for register in registers {
+        match *register {
+            Register::Single(ref _into) => registers_expanded.push(register.clone()),
+            Register::Array(ref _into, ref array_info) => {
+                let index_convertible = if let Some(ref indexes) = array_info.dim_index {
+                    let mut index_iter = indexes.iter();
+                    let mut previous = 0;
+
+                    let get_number_option = |element_option: Option<&String>| -> Option<usize> {element_option
+                                                                        .and_then(|element| match element.parse::<usize>() {
```
japaric: this `match element.parse()` is equivalent to `element.parse().ok()`

src/generate.rs
```diff
@@ -356,16 +355,71 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
...
+                                                                            Ok(i) => Some(i),
+                                                                            _ => None,
+                                                                        })
+                    };
+
+                    let mut index_convertible_temp = match get_number_option(index_iter.next()) {
```
japaric: this match is equivalent to `LHS == Some(0)`

src/generate.rs
```diff
@@ -356,16 +355,71 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
...
+                        Some(0) => true,
+                        _ => false,
+                    };
+                    
+                    for element in index_iter {
+                        if !index_convertible_temp { break; }
+
+                        index_convertible_temp = match get_number_option(Some(element)) {
```
japaric: this `get_number_option(Some(element))` would be simpler as `element.parse().ok()`

japaric: the `match` can probably be replaced by `LHS == Some(previous + 1)`

src/generate.rs
```diff
@@ -356,16 +355,71 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
...
     let mut offset = 0;
-    for register in util::expand(registers) {
-        let pad = if let Some(pad) = register.offset.checked_sub(offset) {
+    let mut registers_expanded = vec![];
+
+    for register in registers {
+        match *register {
+            Register::Single(ref _into) => registers_expanded.push(register.clone()),
+            Register::Array(ref _into, ref array_info) => {
+                let index_convertible = if let Some(ref indexes) = array_info.dim_index {
```
japaric: Could you add a comment about what this does? It wasn't entirely clear from reading the contents. Is it checking that `indexes == 0..info.dim`? In that case this may be simpler to write this as `indexes.iter().eq(0..info.dim)`.

src/generate.rs
```diff
@@ -356,16 +355,71 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
...
+                            _ => {break;},
+                        };
+                    }
+                    
+                    index_convertible_temp
+                } else {
+                    false
+                };
+
+                let packed_registers = array_info.dim_increment == 4;
```
japaric: nit: given that this is a simpler check short circuiting on this equality may save some work

src/generate.rs
```diff
@@ -356,16 +355,71 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
     let mut i = 0;
     // offset from the base address, in bytes
     let mut offset = 0;
-    for register in util::expand(registers) {
-        let pad = if let Some(pad) = register.offset.checked_sub(offset) {
+    let mut registers_expanded = vec![];
+
+    for register in registers {
+        match *register {
```
japaric: could you add a note here indicating that we'll only map a register array to an array type if there are no holes and each register in the array is 32 bits wide? Why is the 32 bits check required, by the way? Other registers arrays will be simply unrolled / expanded.

src/util.rs
```diff
@@ -136,88 +136,120 @@ pub fn respace(s: &str) -> String {
...
+                "0x{:02x} - {}",
+                register.address_offset,
+                respace(&register.description)
+            )
+                [..];
+
+            let array_index_match = Regex::new(r"%s\((?P<i>[ a-zA-Z0-9]*)\)").unwrap();
+            //println!("{}", register.name.as_str());
+            let rty = Ident::new(array_index_match.replace_all(register.name.as_str(), "").to_sanitized_upper_case());
+            let reg_name = Ident::new(array_index_match.replace_all(register.name.as_str(), "$i").to_sanitized_snake_case());
```
japaric: As I asked elsewhere wouldn't it be better to replace the "%s" during `util::expand`?

src/util.rs
```diff
@@ -136,88 +136,120 @@ pub fn respace(s: &str) -> String {
...
         }
     }
 
-    out.sort_by_key(|x| x.offset);
-
     out
 }
 
+
+pub fn sort_by_offset(mut registers: Vec<Register>) -> Vec<Register> {
```
japaric: I don't think this routine is complex enough that should be a function.

src/util.rs
```diff
@@ -136,88 +136,120 @@ pub fn respace(s: &str) -> String {
...
+    }
 }
 
+
 /// Takes a list of "registers", some of which may actually be register arrays,
-/// and turns it into a new *sorted* (by address offset) list of registers where
-/// the register arrays have been expanded.
-pub fn expand(registers: &[Register]) -> Vec<ExpandedRegister> {
+/// and turn in into alist of registers where the register arrays have been expanded.
+pub fn expand(register: &Register) -> Vec<Register> {
```
japaric: Given that this will now only be used with `Register::Array` it may be better not to pass `&Register` but rather the contens of the `Register::Array` variant

src/util.rs
```diff
@@ -136,88 +136,120 @@ pub fn respace(s: &str) -> String {
...
-            }
-            Register::Array(ref info, ref array_info) => {
-                let has_brackets = info.name.contains("[%s]");
+                );
 
-                let ty = if has_brackets {
-                    info.name.replace("[%s]", "")
+            for (idx, i) in indices.iter().zip(0..) {
+                let name = if has_brackets {
+                    info.name.replace("[%s]", format!("%s({})", idx).as_str())
```
japaric: why do we want to keep the "%s" in this stage? Is it going to be used afterwards? Why not just directly replace it with the formatted `idx` at this point?

src/util.rs
```diff
@@ -136,88 +136,120 @@ pub fn respace(s: &str) -> String {
...
                 } else {
-                    info.name.replace("%s", "")
+                    info.name.replace("%s", format!("%s({})", idx).as_str())
```
japaric: same question as above

src/generate.rs
```diff
@@ -381,36 +435,31 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
...
+            Register::Array(ref _into, ref array_info) => register.address_offset
+                +
+                register
+                .size
+                .or(defs.size)
+                .ok_or_else(
+                    || {
+                        format!("Register {} has no `size` field", register.name)
+                    },
+                )? * array_info.dim / 8,
```
japaric: nit: I know that it's unrelated to this PR but could you replace the `8` by a `BITS_PER_BYTE` constant?

src/util.rs
japaric: OK. I see why "%s(N)" is being used: to have expanded registers use the same type. This ... feels hackish; I would prefer if this was encoded in the type system rather than using a regex replace. ExpandedRegisters was used for this reason (to make all the expanded registers have the same type).

kjetilkjeka:
I agree that %s(N) is hacky. But the ExpandedRegister didn't work either as there was no support for array. I made a new struct extending syn::Field (by composition) adding size, offset and description. This makes perfect sense in my head as we're converting from svd::Register to rust fields (which syn::Field represents). Another bonus is the ability to use to_tokens instead of formating manually. Hope you agree with this reasoning? This made some refactoring necessary inside util but I think it makes more sense as it is now.

>>I guess that is only valid as long as there are 4 byte (32 bit) registers. However I'm not sure how it can be generalized over register lengths?

>I did not understand this constraint. Why are 4 byte registers special here? Shouldn't the check be register.size == dim_increment? That would check if each array element has reserved space apart from the rergiste

You're not suppose to understand that, was just me beeing stupid. Should be fixed now.

Most of your comments were considering the conditions for expanding. I've cleaned them up a whole lot. I think it's readable to the point making more comment than the one outside the function unnecessary. But since I wrote the code 2 hours ago I'm quite biased, say if you still want me to add a comment there.

I think I've addressed all the issues you mentioned except "Given that this will now only be used with Register::Array it may be better not to pass &Register but rather the contens of the Register::Array variant". I was a bit reluctant to change this out because:
-The similar function ```pub fn convert_svd_register(register: &svd::Register) -> syn::Field``` is used with svd::Register making my mental model of how to use these functions more consistent
- That expanding a Single register is allowed and gives exactly one registers makes sense (atleast in my head)
- One more argument gives the ability to make logic errors by passing non corresponding info and array_info
Anyway It's not a big deal but just think about it, if you still want it changed after, no problem.

homunkulus:
:umbrella: The latest upstream changes (presumably c7aed0a) made this pull request unmergeable. Please resolve the merge conflicts.

kjetilkjeka:
Rebase onto japaric/master to resolve merge conflicts and include bare_metal features

homunkulus:
:umbrella: The latest upstream changes (presumably 82ebef2) made this pull request unmergeable. Please resolve the merge conflicts.

kjetilkjeka:
Rebased onto 0.11.1 to make mergeable

kjetilkjeka:
@japaric What's the status on svd2rust at them moment? Now that the bare metal changes are adopted will this PR be next in line or should I expect others merged before it?

kjetilkjeka:
@japaric I guess you didn't have the time to look at this PR last time we talked about it.

Do you have an updated estimate of when you will be able to review this PR?

src/generate.rs
```diff
@@ -461,69 +460,135 @@ pub fn peripheral(
     Ok(())
 }
 
+struct RegisterBlockField {
+    field: syn::Field,
+    description: Ident,
```
japaric: This is not a identifier; it should be a `String`.

src/generate.rs
```diff
@@ -461,69 +460,135 @@ pub fn peripheral(
...
+                    #name : [u8; #pad],
+                });
             i += 1;
         }
 
         let comment = &format!(
             "0x{:02x} - {}",
             register.offset,
-            util::respace(&register.info.description)
+            register.description,
```
japaric: Please revert this change or the documentation will have random newlines in it. As in: keep the `util::respace`.

kjetilkjeka:
Thank you for the review @japaric

> I've requested two changes: one is required to have svd2rust not regress in terms of documentation rendering. Please fix them and I'll merge this.

This should be fixed now.

> Do you know if any of the devices in the test suite is exercising this change? If none is could you add your SVD file to the test suite?

MK02F12810.svd should exercise this change (amongst other) for this register.

        <register>
          <dim>8</dim>
          <dimIncrement>0x4</dimIncrement>
          <dimIndex>0,1,2,3,4,5,6,7</dimIndex>
          <name>TAGVDW0S%s</name>
          <description>Cache Tag Storage</description>
          <addressOffset>0x100</addressOffset>
          <size>32</size>
          <access>read-write</access>
          <resetValue>0</resetValue>
          <resetMask>0xFFFFFFFF</resetMask>
          <fields>
            <field>
              <name>valid</name>
              <description>1-bit valid for cache entry</description>
              <bitOffset>0</bitOffset>
              <bitWidth>1</bitWidth>
              <access>read-write</access>
            </field>
            <field>
              <name>tag</name>
              <description>14-bit tag for cache entry</description>
              <bitOffset>5</bitOffset>
              <bitWidth>14</bitWidth>
              <access>read-write</access>
            </field>
          </fields>
        </register>
        <register>

The diff for this register (between kjetilkjeka/svd2rust/master and kjetilkjeka/svd2rust/field_array) looks like the following:

    12303c12303,12317
    <         pub tagvdw0s: [TAGVDW0S; 8],
    ---
    >         pub tagvdw0s0: TAGVDW0S,
    >         #[doc = "0x104 - Cache Tag Storage"]
    >         pub tagvdw0s1: TAGVDW0S,
    >         #[doc = "0x108 - Cache Tag Storage"]
    >         pub tagvdw0s2: TAGVDW0S,
    >         #[doc = "0x10c - Cache Tag Storage"]
    >         pub tagvdw0s3: TAGVDW0S,
    >         #[doc = "0x110 - Cache Tag Storage"]
    >         pub tagvdw0s4: TAGVDW0S,
    >         #[doc = "0x114 - Cache Tag Storage"]
    >         pub tagvdw0s5: TAGVDW0S,
    >         #[doc = "0x118 - Cache Tag Storage"]
    >         pub tagvdw0s6: TAGVDW0S,
    >         #[doc = "0x11c - Cache Tag Storage"]
    >         pub tagvdw0s7: TAGVDW0S,

This means that the doc string is the same as it used to be for tagvdw0s0. "#[doc = 0x100 - Cache Tag Storage]"

I've also manually inspected the other registers in this file where the new code path is exercised and they are equivalent to the previous example.

The previous example is also equivalent to the .svd file I tested during devlopment [S32K144.svd](https://github.com/kjetilkjeka/s32k144.rs/blob/master/src/S32K144.svd)

I can, of course, manually inspect more files? But I've also tested it with the different (allowed) uses of dimIncrement and dimIndex manually. So I believe the next step should be to confirm that all the .svd files in the test suite will be compiled without error.

japaric:
Thanks, @kjetilkjeka.

@homunkulus r+

homunkulus:
:pushpin: Commit 52cf267 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit 52cf267 with merge 1f3542f...

homunkulus:
:sunny: Test successful - [status-appveyor](https://ci.appveyor.com/project/japaric/svd2rust/build/1.0.122), [status-travis](https://travis-ci.org/japaric/svd2rust/builds/277468628?utm_source=github_status&utm_medium=notification)
Approved by: japaric
Pushing 1f3542f635c1512ca3155b1e0312021be2c08ee7 to master...

PR #134, opened 2017-08-03
Implemented new functions to format literals as seperated hex idents

therealprof:
The two main reasons behind this change are ensuring that the numbers
(mostly addresses, masks and default values) line up with the datasheet
for easier searching and verification since those are pretty much always
expressed in hexadecimal. The other reason being clippy now complains
about long numeric literals (without seperators) being hard to read -- I
agree.

### DIFF
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -916,7 +916,7 @@ pub fn fields(
                 sc: Ident::new(&*sc),
                 mask: util::hex_or_bool((((1 as u64) << width) - 1) as u32, width),
                 name: &f.name,
-                offset: util::hex(f.bit_range.offset),
+                offset: util::unsuffixed(u64::from(f.bit_range.offset)),
                 ty: width.to_ty()?,
                 write_constraint: f.write_constraint.as_ref(),
             })
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -5,7 +5,7 @@ use cast::u64;
 use quote::{Tokens, ToTokens};
 use svd::{Access, BitRange, Defaults, Device, EnumeratedValues, Field,
           Peripheral, Register, Usage, WriteConstraint};
-use syn::{self, Ident, Lit};
+use syn::{self, Ident};
 
 use errors::*;
 use util::{self, ToSanitizedSnakeCase, ToSanitizedUpperCase, U32Ext, BITS_PER_BYTE};
@@ -424,7 +424,7 @@ pub fn peripheral(
 ) -> Result<()> {
     let name = Ident::new(&*p.name.to_uppercase());
     let name_pc = Ident::new(&*p.name.to_sanitized_upper_case());
-    let address = util::unsuffixed(u64(p.base_address));
+    let address = util::hex(p.base_address);
     let description = util::respace(p.description.as_ref().unwrap_or(&p.name));
 
     items.push(quote! {
@@ -756,7 +756,7 @@ pub fn register(
         let rv = register
             .reset_value
             .or(defs.reset_value)
-            .map(|rv| util::unsuffixed(u64(rv)))
+            .map(|rv| util::hex(rv))
             .ok_or_else(|| {
                 format!("Register {} has no reset value",
                                     register.name)
@@ -868,9 +868,9 @@ pub fn fields(
         access: Option<Access>,
         description: String,
         evs: &'a [EnumeratedValues],
-        mask: Lit,
+        mask: Tokens,
         name: &'a str,
-        offset: Lit,
+        offset: Tokens,
         pc_r: Ident,
         pc_w: Ident,
         sc: Ident,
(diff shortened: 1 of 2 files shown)


therealprof:
@japaric Mind having a look?

src/generate.rs
```diff
@@ -79,7 +79,7 @@ pub fn device(
     });
 
     if let Some(cpu) = d.cpu.as_ref() {
-        let bits = util::unsuffixed(cpu.nvic_priority_bits as u64);
+        let bits = util::hex(cpu.nvic_priority_bits);
```
japaric: I don't think writing this one in hex is going to make it more readable as the value is always in the range from 1 to 8.

src/generate.rs
```diff
@@ -849,9 +849,9 @@ pub fn fields(
                 access: f.access,
                 evs: &f.enumerated_values,
                 sc: Ident::new(&*sc),
-                mask: util::unsuffixed_or_bool((1 << width) - 1, width),
+                mask: util::hex_or_bool((((1 as u64) << width) - 1) as u32, width),
                 name: &f.name,
-                offset: util::unsuffixed(u64(f.bit_range.offset)),
+                offset: util::hex(f.bit_range.offset),
```
japaric: This value will in the range 0..32. TBH, I personally would find it easier to read it in decimal since it appears in the RHS of expressions like `value << offset`.

src/util.rs
```diff
@@ -251,6 +251,32 @@ pub fn access_of(register: &Register) -> Access {
         )
 }
 
+/// Turns `n` into an unsuffixed separated hex ident
+pub fn hex(n: u32) -> Ident {
```
japaric: nit: `Ident` is for (parsed) identifiers. I see that there's no `Lit` variant for hexadecimal so we can't use that. I think returning `Tokens` would be the most correct choice after `Lit`. Could you change this to `Tokens`? (I think you would have to do something like `let mut ts = Tokens::new(); ts.append("0xdead_beef"); ts`

jonas-schievink:
The general idea makes sense to me, as do @japaric's comments. I don't know enough about ARM to have any real insight here :)

homunkulus:
:umbrella: The latest upstream changes (presumably #121) made this pull request unmergeable. Please resolve the merge conflicts.

src/generate.rs
therealprof: I prefer to have everything base 2/16 (as implied by the word "bits" here) related written in hex but I do agree it makes little sense for small numbers. I can get rid of that.

src/generate.rs
therealprof: I agree, will change that back.

src/util.rs
therealprof: I will try that.

src/generate.rs
therealprof: @japaric Actually I can't change that back because `offset` now takes a Token while `util::unsuffixed()` produces a Lit. If you want I can add some more code to produce a decimal token to be fed into offset.

japaric: Or, maybe, `unsuffixed` could return `Tokens`.

therealprof:
@japaric Except for the one change requested which I commented above the other requests have been addressed now.

src/generate.rs
therealprof: Okay, it's done.

src/util.rs
```diff
@@ -306,18 +306,18 @@ pub fn hex_or_bool(n: u32, width: u32) -> Tokens {
     }
 }
 
-/// Turns `n` into an unsuffixed literal
-pub fn unsuffixed(n: u64) -> Lit {
-    Lit::Int(n, IntTy::Unsuffixed)
+/// Turns `n` into an unsuffixed token
+pub fn unsuffixed(n: u64) -> Tokens {
+    let mut t = Tokens::new();
+    t.append (format!("{}", n));
```
jonas-schievink: nit: stray whitespace between `append` and `(`

therealprof: You'll find a lot more nits if you run rustfmt over it, which is preventing me from doing it all the time. 😉

japaric:
Thanks, @therealprof.

@homunkulus r+

homunkulus:
:pushpin: Commit b5756e8 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit b5756e8 with merge b5756e8...

homunkulus:
:broken_heart: Test failed - [status-travis](https://travis-ci.org/japaric/svd2rust/builds/278964263?utm_source=github_status&utm_medium=notification)

homunkulus:
:sunny: Test successful - [status-appveyor](https://ci.appveyor.com/project/japaric/svd2rust/build/1.0.126), [status-travis](https://travis-ci.org/japaric/svd2rust/builds/278964263?utm_source=github_status&utm_medium=notification)
Approved by: japaric
Pushing b5756e8dc787fae2d497e06d488cefb8c5f1c8d6 to master...

PR #153, opened 2017-10-17
Allow optional 'dim_index' if register name contains [%s]

### LINKED ISSUE #146: Conversion failing on XMC4100
I am running svd2rust on an XMC4100.svd file and it is panicing with the following

> 'C:\Users\lucas\.cargo\registry\src\github.com-1ecc
6299db9ec823\svd-parser-0.5.2\src\lib.rs:576 parse::u32(t)', src\libcore\opti
on.rs:839:4

If I followed it correctly it is for a check on 'resetMask'.

I ran the SVD against an XML validator for the SVD schema and it stated everything was ok.

Any suggestions on what i could test to track down the issue?

### LINKED ISSUE #150: SVD Registers with no dimIndex
According to the CMSIS standard:
https://www.keil.com/pack/doc/CMSIS/SVD/html/elem_special.html 

dimElementGroups with "[%s]" in their name are not suppose to use the dimIndex tag. I have an SVD that currently is doing this but getting an error due to the tag missing. The SVD file is loaded in #146 but will only get that far if my PR in japaric/svd#40 is used.

Requesting the the dimIndex missing be allowed if name contains [%s]

lucasbrendel:
Based on SVD spec if the register name contains '[%s]' in its name then the dim_index field is not required.

I think this is a fix for that situation.

This fixes #150 and should then be able to close #146 as i am able to successfully generate after applying these changes.

### DIFF
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -537,21 +537,29 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
             Register::Array(ref info, ref array_info) => {
                 let sequential_addresses = register_size == array_info.dim_increment*BITS_PER_BYTE;
 
-                let numeral_indexes = array_info.dim_index.clone()
+                let numeral_indexes = if !register.name.contains("[%s]") {
+                    array_info.dim_index.clone()
                     .ok_or_else( || format!("Register {} has no `dim_index` field", register.name))?
                     .iter()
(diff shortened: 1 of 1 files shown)


homunkulus:
:umbrella: The latest upstream changes (presumably #158) made this pull request unmergeable. Please resolve the merge conflicts.

japaric:
Closing in favor of #168

PR #156, opened 2017-11-14
Generate interrupt dispatch code.

dvc94ch:
Useful for architectures that prefer or mandate dispatching interrupts in
software. Adds #[derive(Debug)] for printing readable interrupt names.

### DIFF
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -200,7 +200,8 @@ pub fn interrupt(
     let mut interrupts = interrupts.into_iter().map(|(_, v)| v).collect::<Vec<_>>();
     interrupts.sort_by_key(|i| i.value);
 
-    let mut arms = vec![];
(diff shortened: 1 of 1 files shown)


japaric:
Hey @dvc94ch. Thanks for the PR.

I think it would be more idiomatic to return `Option` or `Result` from `Interrupt::from` (or maybe even implement `TryFrom`) rather than panic! on invalid inputs.

japaric:
Thanks for making the change, @dvc94ch. (I didn't get a notification so I didn't notice until now).

This LGTM now but the generated code needs to contain some feature gates because the `TryFrom` API is unstable. There are also some import errors:

```
error[E0405]: cannot find trait `TryFrom` in this scope
```

dvc94ch:
Superseded by #185

PR #163, opened 2017-12-23
generate a CPU constant that maps to cpu.name

japaric:


japaric:
We no longer need this for japaric/cortex-m-rtfm#60.

@Emilgardis You +1-ed this PR. Do you have a use case for this? If so I can still land the PR -- it seems uncontroversial.

Emilgardis:
I have no real use case. However, having this information available is as you say, uncontroversial.

japaric:
I'm going to punt on this until there's a use case for it. (I'd prefer not to maintain more things in the output)

PR #168, opened 2018-01-19
Handle missing dimIndex on registers

### LINKED ISSUE #150: SVD Registers with no dimIndex

kevinmehall:
This supersedes @lucasbrendel's similar PR #153 and fixes #150.

Per the [SVD spec](http://arm-software.github.io/CMSIS_5/SVD/html/elem_special.html#dimElementGroup_gr), dimIndex is not required:

> By default, \<dimIndex\> is a value starting with 0.
> Remark: Do not define \<dimIndex\> when using the placeholder [%s] in \<name\> or \<displayName\>.

I interpret this as the element being optional with "%s" and not allowed with "[%s]" (this change does not reject the latter). Atmel SAMD21 SVDs use "%s" with no dimIndex for 0-indexed arrays.

This also tests that dimIndex is sequential and numeric in a single pass without allocating temporary vectors, and fixes the typo adresses -> addresses.

### DIFF
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -564,32 +564,17 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
                 size: register_size,
             }),
             Register::Array(ref info, ref array_info) => {
-                let sequential_adresses = register_size == array_info.dim_increment * BITS_PER_BYTE;
+                let sequential_addresses = register_size == array_info.dim_increment * BITS_PER_BYTE;
 
-                let numeral_indexes = array_info
-                    .dim_index
-                    .clone()
-                    .ok_or_else(|| {
-                        format!("Register {} has no `dim_index` field", register.name)
-                    })?
-                    .iter()
-                    .all(|element| element.parse::<usize>().is_ok());
-
-                let sequential_indexes = if numeral_indexes && sequential_adresses {
-                    array_info
-                        .dim_index
-                        .clone()
-                        .unwrap()
+                // if dimIndex exists, test if it is a sequence of numbers from 0 to dim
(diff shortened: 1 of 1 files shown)


jamesmunns:
Not sure how I missed this, but this PR also addresses the issues I had mentioned in https://github.com/japaric/svd2rust/pull/177.

:+1: from me!

jamesmunns:
@therealprof you may want to test your svds with this branch.

therealprof:
@jamesmunns For the ATSAMD the result is pretty much the same (as far as I can tell from memory it's indentical) as with your PR. The result in this particular case is still broken but I blame Atmel here.

kevinmehall:
@therealprof Yeah, the SAMD SVDs will still need some more features implemented including `derivedFrom` on registers, register clusters, and building unions for registers with duplicate addresses. None of which is actually in violation of the SVD spec as far as I've seen, but yes, they definitely could have done it more elegantly.

japaric:
@homunkulus r+

japaric:
bors retry

japaric:
bors-ng doesn't seem to have a retry command implemented, then ...

bors r+

japaric:
[All tests passed](https://travis-ci.org/japaric/svd2rust/builds/345753607). Merging manually.

PR #175, opened 2018-02-07
Enable code generation in the absence of register reset values.

### LINKED ISSUE #174: Handle SVD files which don't define register reset values
Atmel's ARM chips omit the `<resetValue>` element for register descriptions. `<resetValue>` is used to generate a `write` function which doesn't have to read the initial value from a register in order to preserve it while modifying individual fields.

ecstatic-morse:
The second commit also generates `reset` for write-only registers. Previously it was only generated for read-write registers. I'm not sure if the previous behavior was intended. This resolves #174.

### DIFF
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -759,19 +759,6 @@ pub fn register(
     }
 
     if access == Access::WriteOnly || access == Access::ReadWrite {
-        reg_impl_items.push(quote! {
-            /// Writes to the register
-            #[inline]
-            pub fn write<F>(&self, f: F)
-            where
-                F: FnOnce(&mut W) -> &mut W
-            {
-                let mut w = W::reset_value();
-                f(&mut w);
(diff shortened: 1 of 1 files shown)


ecstatic-morse:
Sorry, removed offending commit. 
I could also update the Travis CI tests to run on Atmel targets, but you might want to only uncomment a fraction of them.

japaric:
@ecstatic-morse could you uncomment some of the Atmel SVD files to test this change?

Emilgardis:
cc @ecstatic-morse are you still working on this?

I was misinformed with why d6d4b82 was needed. Could you enable tests for ~3 Atmel SVD files?

Disasm:
I hope we can close this PR in favor of https://github.com/rust-embedded/svd2rust/pull/258
Thank you! Feel free to open new PR if you want to continue working on this.

PR #179, opened 2018-02-25
Do not transform peripheral names

docbrown:
Currently, `svd2rust` always transforms peripheral names to uppercase. For peripherals that are not acronyms, like SysTick, the resulting type name looks a little odd. This patch simply sanitizes peripheral names, but leaves them in their original case. For vendor SVDs with weirdly-formatted names, an XML transform or .patch could be used to fix them, much like device-specific crates already do to add `enumeratedValues` and such.

This is a breaking change that could lead to API breakage in existing crates that assume the uppercase transformation is always done.

### DIFF
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -8,7 +8,7 @@ use svd::{Access, BitRange, Defaults, Device, EnumeratedValues, Field, Periphera
 use syn::{self, Ident};
 
 use errors::*;
-use util::{self, ToSanitizedSnakeCase, ToSanitizedUpperCase, U32Ext, BITS_PER_BYTE};
+use util::{self, Sanitize, ToSanitizedSnakeCase, ToSanitizedUpperCase, U32Ext, BITS_PER_BYTE};
 use Target;
 
 /// Whole device generation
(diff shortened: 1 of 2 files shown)


docbrown:
I can undo the change to the `Peripherals` fields since that is guaranteed to break existing crates, but `Peripherals::take().sys_tick` looks more Rustic than `Peripherals::take().SYSTICK` or `Peripherals::take().SysTick` to me.

Emilgardis:
This does in a way make it look more _rusty_, but I'm not sure if this is what we want as the change would be "purely" cosmetic. Uppercase also reminds me of static values, which matches the idea of having singleton peripherals.

mathk:
Ping from triage: @rust-embedded/tools can someone tell if that can be accepted ?

mathk:
Ping from triage: @docbrown need a merge/rebase.

japaric:
This is a breaking change since it may change the name of peripheral types so first it needs to be approved by the team and then it must wait until we are ready for the next minor release before it can land in tree.

This sounds good to me; iirc, we already leave register and bitfield names as they are so this would make things more consistent. AFAIS, most vendors use uppercase for peripheral names so most people won't see any change though.

I will leave the rest of the team comment. (Changing label to S-waiting-on-team)

thejpster:
If we're going to break things, I'd like someone to suggest how we can distinguish modules, PAC objects and HAL implementations other than only by using different capitalisations of the same letters (see `uarte`, `Uarte` and `UARTE`).

PR #180, opened 2018-02-26
Reorganize svd2rust

jamesmunns:
I have made a few minor cosmetic changes:

* Break `generate.rs` into a module, with each of the major components living as a file inside of it
* Each major component has a function called `render()`, which all return `Result<Vec<Tokens>>` (as before some did that, and some took a `&mut Vec<Tokens>` as an argument they only pushed to)
* I moved some `util` components into modules, when they were only used by that module

I expect NO functional changes to the output. Please let me know if I can run a full regression test locally.

In the future, I think some more work could be done to break up `generate::register::render()`, as it is a very large function, but I wanted to make this PR in manageable pieces.

Please let me know if you would like me to squash my changes. I have kept them separate in case you would like to inspect them.


jamesmunns:
@japaric or @emilgardis could I trouble you to review and possibly merge this? Or at least provide some positive or negative feedback (sorry, I know you are both busy)

japaric:
This LGTM. Eventually we want to add unit tests at each codegen level (enumeratedValue, register, peripheral, etc.) so it makes sense to split the generate.rs file this.

bors r+

japaric:
> Build failed

One build timed out. Testing locally.

japaric:
> Testing locally.

Test passed. Merging manually.

PR #149, opened 2017-09-30
Svd clusters

brandonedens:
Add support for generating rust software to support the <cluster> tag.

Issue #16, opened 2016-10-20
"overloaded" registers


japaric:
STM32F30x.svd
Peripheral: TIM2

The SVD file specifies that **two** registers exist at offset 0x1C: `CCMR2_Input` and `CCMR2_Output`. The thing is that this register can be used *differently* depending on whether the TIM2 peripheral is configured in Output Compare Mode or in Input Compare Mode. By differently, I mean that the bits of this register have different meaning depending on the compare mode.

I'm not sure how to model this. We could use a `union` here but that's not stable. For now, I think, I'm just going to pick one register and ignore the other but emit a warning.

cwoodall:
One approach would be to make both registers available but disable any constraints that would make read or write access to the register safe.

etrombly:
Having a similar issue with an STM32F103. The MODE register of a GPIO configures it for input or output, the CNF register handles the configuration (open-drain, push-pull, etc) but the settings are different depending on the mode.

silven:
I think I've run into this as well.

Trying to use the generated file from: https://github.com/nicholastmosher/lpc176x5x-rs/blob/master/src/uart.rs which has a register called "DLL", but because it overlaps with the "RBR" register at offset 0x0, it cannot be used at all.

The SVD file used (https://raw.githubusercontent.com/posborne/cmsis-svd/aa4721af946a253d18c8737b01d23e9c88a42e84/data/NXP/LPC176x5x_v0.2.svd) annotates this with "alternateRegister".

So, "+1" I guess.

protomors:
To @etrombly. I solved this problem by changing field definitions in SVD file. CNF and MODE bits for the same pin are adjacent so instead of separate 2-bit fields I declared them as one 4-bit field. And then created enum for them according to the "Table 20. Port bit configuration table" in the reference manual. So now I can select mode with single call like "output10pp" (Output, push-pull, max speed 10MHz) or "input_floating".

But there are cases where such solution is not possible. I think that in situation like example with timers it will be best to generate API for both registers (maybe just print a warning about it during generation). And then programmer can use appropriate API depending of the mode.

pftbest:
[SVD files](https://github.com/pftbest/msp430g2553/blob/master/msp430g2553.svd) I am generating for MSP430 are also affected by this issue, but since the overlapping registers are not in the same peripheral, the svd2rust does not complain about them.

japaric:
@pftbest 

> but since the overlapping registers are not in the same peripheral,

[...]

PR #182, opened 2018-02-27
Add initial Cluster support

jamesmunns:
Add cluster support to svd2rust. I am calling it initial, because I have only tested it against the `nRF52` [svd file](https://github.com/jamesmunns/nrf52/blob/master/nrf52.svd).

Note, this is based on https://github.com/japaric/svd2rust/pull/180, so I would prefer that PR be merged first. If that PR is rejected, I can instead rebase these changes on `japaric/master` (just let me know).

Edit: All changes relevant to clusters are squashed into the last commit of this PR, so if you want to see the relevant changes, check out https://github.com/japaric/svd2rust/pull/182/commits/6de6de760c535e6f60eed4acb41e30db1458f33c.

This is based on the work of @brandonedens earlier https://github.com/japaric/svd2rust/pull/149, and I believe would close https://github.com/japaric/svd2rust/issues/107 and related issues.

CC @japaric @emilgardis @therealprof

Edit 2: Here are some handy references:
* [SVD snippet with clusters](https://gist.github.com/jamesmunns/c4e53fe5bd74dca81fdbff6bb1798ddd)
* [code generated before](https://gist.github.com/jamesmunns/d854a6c2665cca59edb88143d82a19c6)
* [code generated after](https://gist.github.com/jamesmunns/7558667e34c33124c60b3aaaf679a196)
* [diff of code generated](https://gist.github.com/jamesmunns/03fd0d7d3595cd0816dbfa8daca0a553)


src/generate/peripheral.rs
```diff
@@ -186,40 +145,237 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
...
+
 /// Takes a svd::Register which may be a register array, and turn in into
 /// a list of syn::Field where the register arrays have been expanded.
-fn expand_svd_register(register: &Register) -> Vec<syn::Field> {
-    let name_to_ty = |name: &String| -> syn::Ty {
+fn expand_svd_register(register: &Register, name: Option<&str>) -> Vec<syn::Field> {
+    let name_to_ty = |name: &String, ns: Option<&str>| -> syn::Ty {
+        let ident = if let Some(ns) = ns {
+            Cow::Owned(
+                String::from("self::") + &ns.to_sanitized_snake_case() + "::"
```
Emilgardis: I don't understand why this change is needed, care to explain?

src/generate/peripheral.rs
```diff
@@ -186,40 +145,237 @@ fn register_block(registers: &[Register], defs: &Defaults) -> Result<Tokens> {
...
+    let name = Ident::new(match name {
+        Some(name) => name.to_sanitized_upper_case(),
+        None => "RegisterBlock".into(),
+    });
+
     Ok(quote! {
         /// Register block
         #[repr(C)]
-        pub struct RegisterBlock {
+        pub struct #name {
```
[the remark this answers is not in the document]
jamesmunns: At least in my testing with nrf52, this did not cause any problems with rtfm. This is an artifact of #149, and I admit I do not fully understand it. I can try testing without this change if you would like.

src/generate/peripheral.rs
jamesmunns: This change allows for generation of names like this: https://gist.github.com/jamesmunns/03fd0d7d3595cd0816dbfa8daca0a553#file-ficr-diff-L58

Since clusters are nested namespaces, it is necessary for the outer (cluster) structures to refer to types inside of the nested namespaces.

jamesmunns:
I have also now tested this against `ATSAMD21E15A.svd`, and that has revealed that there are some additional problems that need to be fixed. One of these issues is fixed in d275d6f.

Another problem is that the ATSAMD svd has clusters with repeating names. This is currently not handled by my changes. See this snippet:

```xml
       <cluster>
        <name>USART</name>
        <description>USART Mode</description>
        <alternateCluster>I2CM</alternateCluster>
        <headerStructName>SercomUsart</headerStructName>
        <addressOffset>0x0</addressOffset>
        <register>
          <name>BAUD</name>
          <description>USART Baud Rate</description>
          <alternateGroup>DEFAULT_MODE</alternateGroup>
          <addressOffset>0x0C</addressOffset>
          <size>16</size>
          <fields>
            <field>
              <name>BAUD</name>
              <description>Baud Rate Value</description>
              <bitOffset>0</bitOffset>
              <bitWidth>16</bitWidth>
            </field>
          </fields>
        </register>
        <register>
          <name>BAUD</name>
          <description>USART Baud Rate</description>
          <alternateGroup>FRAC_MODE</alternateGroup>
          <addressOffset>0x0C</addressOffset>
          <size>16</size>
          <fields>
            <field>
              <name>BAUD</name>
              <description>Baud Rate Value</description>
              <bitOffset>0</bitOffset>
              <bitWidth>13</bitWidth>
            </field>
            <field>
              <name>FP</name>
              <description>Fractional Part</description>
              <bitOffset>13</bitOffset>
              <bitWidth>3</bitWidth>
            </field>
          </fields>
        </register>
        <register>
```

jamesmunns:
@Emilgardis - apologies for basing this on my "refactoring" branch. It took me a while to understand what components had to do with which parts of the generation process, and the refactoring branch was me organizing things based on my learning while trying to tackle the cluster feature. As previously mentioned, I would be happy to rebase this on top of master (but I am hoping the other PR is merged).

Additionally I agree that I need to improve the situation with comments. I will plan to add those in the near future.

src/generate/peripheral.rs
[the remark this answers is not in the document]
Emilgardis: I think I'm thinking of the older RTFM where (I think) RegisterBlock was directly used. Now we just deref "into" it so this shouldn't change anything.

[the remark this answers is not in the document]
jamesmunns: @Emilgardis I actually just went and figured out what this does. It controls the name of structs that are clusters, such as https://gist.github.com/jamesmunns/03fd0d7d3595cd0816dbfa8daca0a553#file-ficr-diff-L56. Without this, the other structs inside of this scope would all have the name `RegisterBlock`, which would be a problem.

At this point, it is implicit in the code that `name` is always `Some(...)` in a cluster, and `None` when not in a cluster. I will add comments that improve that (now that I understand this :smile:)

jamesmunns:
I think https://github.com/japaric/svd2rust/pull/182#issuecomment-368916284 is not technically a blocker to merging this, though clusters that contain multiple registers at the same offset (like `ATSAMD21E15A.svd` shown above) will not (yet) be able to take advantage of the cluster support, until there is support for union'd types (I think covered by issue #16).

therealprof:
Don't worry about the FUBAR ATSAMD too much, that'll only make your life miserable but doesn't bring any benefit to the cause. The chat I had with a Microchip guy yesterday at the Embedded World was very revealing, too.

src/util.rs
```diff
@@ -238,3 +239,25 @@ impl U32Ext for u32 {
         })
     }
 }
+
+/// Return only the clusters from the slice of either register or clusters.
+pub fn only_clusters(ercs: &[Either<Register, Cluster>]) -> Vec<&Cluster> {
```
japaric: off-topic: this looks like something that could be provided by the either crate. It looks a bit like [`Iterator::partition`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition) but that method returns two `Vec`s of the same type.

Emilgardis: I opened an issue in bluss/either#20 regarding this

jamesmunns:
Hey, quick status update on this:
- Currently register offsets inside of clusters are not being calculated correctly
- I have a very ugly brute force fix for it, and I am working on cleaning it up a bit
- My laptop fan just died, so I need to get that fixed before I can do any serious compilation or testing, so I expect a delay of a day or two.

I just wanted to follow up in case anyone other than me was trying to use this branch :). If you want to try out my ugly fixes, I can push those to another branch on my fork, just @ me.

therealprof:
@jamesmunns I just tried a few SVD and it seems to work great. No regressions with simple ones and nrf51 gets a few more components by using this.

Few nits:
- I'm getting lint errors due to gratuitous use of `use vcell::VolatileCell;` where not needed.
- The formatting after `rustfmt` is slightly different, even worse than it was before, e.g. docstrings right after opening braces: `pub enum SRCR {#  [ doc = "Internal 32KiHz` ...

jamesmunns:
@therealprof Be careful with the current branch. In the generated output, the `#[doc = ""]` comments have the correct offsets, but extra `_reservedxx` fields are being generated if your clusters do not have an explicit `<size>` field. You can verify this in GDB by looking at the addresses of the fields.

Regarding formatting, yeah, its pretty rough. I wonder what we are doing that makes cargo fmt upset. Probably worth an improvement issue.

Regarding VolatileCell, I can probably take that out of the generated code, and we can just `use vcell::VolatileCell;` once at the top of the file (in fact, I think we already do).

therealprof:
@jamesmunns I've just done some basic regression testing and worked for my curiosity (since you mentioned you require that for the nrf52 I was curious to see whether the nrf51 also has some additional hidden components -- yes it does!). No concrete use case at the moment.

jamesmunns:
@therealprof I don't expect this to break anything that already worked before, but the "new stuff" (e.g. the peripherals/registers that were cluster based) I would expect not to work in some cases.

Were you testing the new peripherals? Or your existing nrf51-hal code?

therealprof:
No, just checking that a swapped out version would compile. I haven't even noticed that (albeit very few) peripherals were missing so obviously I didn't those so far.

src/util.rs
jamesmunns: At least right now, I won't be implementing this, because I can't figure out how to get itertools to figure out the vec of refs. Open to PRs in the future :)

jamesmunns:
Fixed size rendering (as far as I can tell).

bors r+

jamesmunns:
Looks like this failed due to the excess "VolatileCell" mentioned above: https://travis-ci.org/japaric/svd2rust/jobs/349051903 - I'll try and find a clever fix, otherwise I'll hit it with an `#[allow(unused)]`.

jamesmunns:
@japaric I'm a little confused - CI passed but Bors timed out? Is this okay to merge?

jamesmunns:
Merging as tests have passed.

japaric:
@jamesmunns I increased the [timeout](https://github.com/japaric/svd2rust/blob/631ab3e5ffaec4bb276d930b265a2a8178e7d82f/bors.toml#L5) to 2 hours the other day. If it still takes longer than that to run the tests feel free to increase it.

istankovic:
@therealprof:
> Don't worry about the FUBAR ATSAMD too much, that'll only make your life miserable but doesn't bring any benefit to the cause. The chat I had with a Microchip guy yesterday at the Embedded World was very revealing, too.

Care to share the details? Also, I'd be curious to know whether you consider previously-Atmel chips FUBAR or just the SVDs. Thing is, I have an ATSAMD21 board that I'd like to bring up and play with the peripherals and the register cluster issue has been blocking, as far as I understand.

therealprof:
@istankovic For whatever reason the engineers at (formerly) Atmel decided that it would be a great idea to have lots of totally asynchronously running clocks driving the individual peripherals. While (in theory) this yields a ton of freedom tweaking all the clocks to run at just the perfect speed for the job, it actually means that you can never be sure that data arrives at the right place at the right time due to clock drift so whenever you have peripherals working in lockstep (as one pretty much always does) you have to sprinkle costly busy wait loops all over the map in your peripheral access code. Not to mention that getting the clocks set up in the first place is a major PITA and can easily drive one mad because the chips will simply lock up somewhere in the initialisation code and you've no idea why... That combined with the fact that the documentation is in large portions conflicting, incorrect or even absent and Atmel seems to consider examples to be for wussies only is a bit of a buzzkill.

I was asking the Microchip guy at the Embedded World for new advancements in the BLE/802.15.4 area to which he frankly replied that they somewhat underestimated the demand for that so they're still playing catchup in their design teams. And when I nudged him a bit about my experience with the ATSAM chips and the crazy clocks he just nodded and said that this probably wasn't the brightest hour of the engineering team.

The SVDs are just another sad part of the story...

Wez Furlong is working on ATSAMD21 support and if you want to look at something "working" (although definitely not pretty) you might want to have a look at my https://github.com/therealprof/atsamd20e15a which implements something "shiny" for the lowlier but otherwise mostly compatible brother.

istankovic:
@therealprof thanks for the elaboration. Yes, regarding the Atmel parts, clock configuration can sometimes really be painful and I spent many hours debugging non-functioning peripherals. However, I'd say that once one gets used to it, it isn't such a big deal and it has its advantages (lower power, speed optimization etc.) Regarding the documentation, I'm not sure about the current state, but back when I was actively working with Atmel parts, AVRs upto Cortex-Ms, I considered Atmel docs to be one of the better in the industry, honestly. One could actually build a system solely by following a datasheet and perhaps consulting a few app notes, without needing to spend a great deal of time searching the net for answers/advice. Maybe that changed.

Thanks for the pointer, I'll check out your repo.

therealprof:
@istankovic 

> it isn't such a big deal and it has its advantages (lower power, speed optimization etc.)

That's what they claim, I haven't seen that in real life though.

> Atmel parts, AVRs upto Cortex-Ms, I considered Atmel docs to be one of the better in the industry, honestly.

The AVR documentation is indeed really good; I've worked a lot with ATTiny, ATMega and AT90 in the past and those are well documented in stark contrast to the ATSAMD parts; haven't bothered checking other ATSAM parts though.

PR #190, opened 2018-03-11
First pass at implementing derivedFrom.

### LINKED ISSUE #181: Missing peripherals in the generated file
Using HEAD svd2rust on Atmel's SAMD10's svd file, quite a few peripherals are not generated, and no errors are issued.

I made a [repo](https://github.com/nsabovic/atsamd10d13am) which contains the [svd file](https://raw.githubusercontent.com/nsabovic/atsamd10d13am/master/svd/ATSAMD10D13AM.svd).

Running the latest version of svd2rust as of 2/25/18, v0.12.0-8-g65482c2, the following peripherals which are in svd are not generated: pac1, pac2, port_iobus, rtc, sercom0, sercom1, sercom2, tc1, tc2, wdt. The svd passes SVDConv with few warnings, and svd2rust also outputs a few warnings like:

    WARNING count_dith4 overlaps with another register at offset 52. Ignoring.

nsabovic:
Trying to fix #181, this is the first pass at implementing derivedFrom. These also the very first lines I've ever written in Rust, so be gentle.

This patch requires japaric/svd#49.

With this patch, I get a bit further in generating the peripherals for SAM D10.


jamesmunns:
Hey @nsabovic, sorry there hasn't been any response yet. I'll try to review and test this over the next few days, and also follow https://github.com/japaric/svd/pull/49 to see if anything happens there.

Thanks for contributing!

src/generate/peripheral.rs
```diff
@@ -11,24 +11,35 @@ use util::{self, ToSanitizedSnakeCase, ToSanitizedUpperCase, BITS_PER_BYTE};
...
+    let derive_regs = p_derivedfrom.is_some() && p_original.registers.is_none();
 
     let name_sc = Ident::new(&*p.name.to_sanitized_snake_case());
-    let (base, derived) = if let Some(base) = p.derived_from.as_ref() {
-        // TODO Verify that base exists
-        // TODO We don't handle inheritance style `derivedFrom`, we should raise
-        // an error in that case
-        (Ident::new(&*base.to_sanitized_snake_case()), true)
+    let base = if derive_regs {
+        Ident::new(&*p_derivedfrom.unwrap().name.to_sanitized_snake_case());
```
wez: Need to remove the `;` from the end of the line here and in the else branch below because they're causing `base` to have the type `()` rather than the `Ident` type that you're constructing here, and that causes things to explode because the generated code comes out like this:

```
impl Deref for SERCOM4 {
    type Target = ( ) :: RegisterBlock ;
...
``

**Comment by wez** [comment]
@nsabovic are you planning to do something similar for registers?

**Comment by wez** [comment]
FWIW, I'd vote for inlining the helper from https://github.com/japaric/svd/pull/49 into this PR and it will be easier to land and integrate just a single PR; PR dependencies are a PITA :)

```
fn derive_peripheral(p: &Peripheral, other: &Peripheral) -> Peripheral {
    let mut derived = p.clone();
    if derived.group_name.is_none() && other.group_name.is_some() {
        derived.group_name = other.group_name.clone();
    }
    if derived.description.is_none() && other.description.is_some() {
        derived.description = other.description.clone();
    }
    if derived.registers.is_none() && other.registers.is_some() {
        derived.registers = other.registers.clone();
    }
    if derived.interrupt.is_empty() {
        derived.interrupt = other.interrupt.clone();
    }
    derived
}
```

**Comment by wez** [comment]
Over in https://github.com/japaric/svd/pull/50 I've added parsing out the `<register derivedFrom=` attribute, and together with this PR, I have some code on top that expands derived registers: https://github.com/wez/svd2rust/tree/derivedFrom.

That resolves the bad sizing issue that @kevinmehall mentioned in https://github.com/japaric/svd2rust/issues/181#issuecomment-368392498.  Actually, I think all three of those points have now been addressed and we just need to review, clean things up and land them.

**Comment by nsabovic** [comment]
@wez thanks for catching that 👍 

Registers, clusters and fields can all derive from others, would be nice to make a more general solution. My chip's SVD (SAM D10) doesn't have `<register derivedFrom="">` in the SVD, so I spaced out on that part of the spec completely...

...but it does have `alternateCluster` and `alternateRegister`, so I'd love to have those—the API with unions, does it require using `unsafe` all over the place? I was thinking, since my chip has an alternate register for, say PER register in dither5 mode, to have in the `impl` of PER register a function called `as_dither5()`, which would give you a dither5 view of the same register, while also taking a reference so you can't access it in both modes at the same time. Seems rustier, and also capitalizes on the fact that there is a default register, and the alternate register. What do you think?

**Comment by mathk** [comment]
Ping from triage: This need to be review. Also merge/rebase is needed.

**Comment by Disasm** [comment]
Ping from triage: @nsabovic could you rebase your changes?

**Comment by ryankurte** [comment]
bors try

File: src/generate/peripheral.rs [hunk 2]
```diff
@@ -12,26 +12,38 @@ use util::{self, ToSanitizedSnakeCase, ToSanitizedUpperCase, BITS_PER_BYTE};
...
     defaults: &Defaults,
     nightly: bool,
 ) -> Result<Vec<Tokens>> {
     let mut out = vec![];
 
+    let p_derivedfrom = p_original.derived_from.as_ref().and_then(|s| {
+        all_peripherals.iter().find(|x| x.name == *s)
+    });
+
+    let p_merged = p_derivedfrom.map(|x| p_original.derive_from(x));
```
ryankurte: where does the `Peripheral::derive_from(&mut self, p: Peripheral)` method come from? (this might have got disappeared during rebase..?)

michalfita:
Does anyone has working branch with stuff for supporting `derivedFrom`? I tried @wez's and @nsabovic's and both fail to build.

therealprof:
@nsabovic Sorry for the long delay. Are you still interested in working on this? If so would you mind bringing the PR into a working state?

wez:
https://github.com/rust-embedded/svd2rust/pull/256 is the successor to this PR

nsabovic:
I am not really set up to look at this right now. If you need it urgently, I'm not the best bet.

adamgreig:
Closing in preference of now merged #319 which is built on this and #256.

PR #199, opened 2018-03-27
Improve ci

Emilgardis:
This enables rustfmt, makes seeing what errors happened during `cargo check` easier and some minor other changes.

### DIFF
--- a/ci/svd2rust-regress/src/main.rs
+++ b/ci/svd2rust-regress/src/main.rs
@@ -6,16 +6,18 @@ extern crate reqwest;
 #[macro_use]
 extern crate structopt;
 
-mod tests;
 mod errors;
 mod svd_test;
+mod tests;
 
-use std::path::PathBuf;
-use structopt::StructOpt;
 use rayon::prelude::*;
+use std::path::PathBuf;
+use std::process::{exit, Command};
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::process::exit;
 use std::time::Instant;
+use std::fs::File;
+use std::io::Read;
+use structopt::StructOpt;
 
 #[derive(StructOpt, Debug)]
 #[structopt(name = "svd2rust-regress")]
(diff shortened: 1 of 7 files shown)


ci/svd2rust-regress/src/main.rs
```diff
@@ -93,6 +108,33 @@ fn main() {
...
+    let default_rustfmt: Option<PathBuf> = if let Some(v) = Command::new("rustup")
+        .args(&["which", "rustfmt"])
+        .output()
+        .ok()
+        .map(|v| v.stdout)
+    {
+        Some(String::from_utf8_lossy(&v).into_owned().trim().into())
+    } else {
+        None
+    }; // FIXME: capture error and assure exit is 0
```
jamesmunns: @Emilgardis did you want to clean this up before merging?

ci/svd2rust-regress/src/tests.rs
```diff
@@ -76,7 +76,7 @@ use self::RunWhen::*;
 /// that are not valid in Rust ident
 const BLACKLIST_CHARS: &'static [char] = &['(', ')', '[', ']'];
 
-/// Lovingly stolen from `svd2rust`. Probably could be `Cow`, but I'm lazy
```
jamesmunns: :smile:

ci/svd2rust-regress/src/main.rs
```diff
@@ -133,8 +175,9 @@ fn main() {
     tests.par_iter().for_each(|t| {
         let start = Instant::now();
 
-        match svd_test::test(t, &bin_path) {
+        match svd_test::test(t, &bin_path, rustfmt_bin_path) {
             Ok(()) => {
+                // TODO: If verbosity is > 1, print every logged stderr
```
jamesmunns: TODO check

ci/svd2rust-regress/src/svd_test.rs
```diff
@@ -50,25 +50,27 @@ impl CommandHelper for Output {
...
             file_helper(&out_payload, out)?;
-        }
+        };
 
         if let Some(err) = stderr {
             let err_payload = String::from_utf8_lossy(&self.stderr);
             file_helper(&err_payload, err)?;
+        };
+
+        if cant_fail && !self.status.success() {
```
jamesmunns: Ah, I hadn't thought about putting this check at the end. Good call :+1:

ci/svd2rust-regress/src/main.rs
Emilgardis: This one I feel can be implemented later.

ci/svd2rust-regress/src/main.rs
Emilgardis: I'm not sure what I want to do about this because if this fails it would simply mean that rustup is missing.
I think a panic if opt.format is true and opt.rustfmt_bin_path would suffice

Emilgardis:
Seems like i686-apple-darwin never ran

Manually merging as this should work as expected

PR #201, opened 2018-04-22
Expose internals to lib

dfrankland:
This makes it nice and easy for `cargo build` scripts to generate the source for SVD files, similar to the way that `bindgen` works.

For an example take a look at <https://github.com/dfrankland/mk20d7/blob/master/build.rs>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -410,6 +410,25 @@
 
 // NOTE This file is for documentation only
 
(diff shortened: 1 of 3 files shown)


Emilgardis:
Thank you!

bors r+

Emilgardis:
halting due to riscv-rust/riscv-rt#3 blocking ci

bors r-

PR #203, opened 2018-04-26
cortex-m: generate code that compiles on stable

japaric:
This is still missing some stuff:

- svd2rust must now generate an interrupts.x linker script and a build script that puts the linker script somewhere the linker can find it. See japaric/stm32f103xx#24 for a sample of what those files look like.

- we have to decide how to organize the svd2rust output. It probably makes sense to generate 3 files on $PWD (lib.rs, build.rs, interrupts.x) for Cortex-M and a single lib.rs for the other targets.

- I probably broke the other targets; need to test them

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,27 +12,38 @@ matrix:
   include:
     # Nightly, for testing
     - env: TARGET=x86_64-unknown-linux-gnu VENDOR=Atmel
-      rust: nightly
+      rust: beta
+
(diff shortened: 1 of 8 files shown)


japaric:
@jamesmunns heads up. This PR breaks svd2rust-regress because svd2rust now emits three files in the current directory instead of dumping the generated code to stdout.

japaric:
All the vendor tests passed locally so merging manually.

PR #212, opened 2018-05-09
Change inflection on unions

Emilgardis:
PascalCase doesn't work when we have overloaded registers named
`FOO_11` and `FOO1_1`.

This happens in
* Freescale MKL81Z7
* Freescale MKL82Z7
* Freescale MK81F25615
* Freescale MK82F25615
on `LTC0_PK*` registers.

I'm not sure if `{ident}Union` is the best choice anymore as `FOO_BARUnion` doesn't seem as good as it possibly could be.

cc @jamesmunns @wez

### DIFF
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -7,7 +7,7 @@ use svd::{Cluster, ClusterInfo, Defaults, Peripheral, Register};
 use syn::{self, Ident};
 
 use errors::*;
-use util::{self, ToSanitizedPascalCase, ToSanitizedSnakeCase, ToSanitizedUpperCase, BITS_PER_BYTE};
+use util::{self, ToSanitizedSnakeCase, ToSanitizedUpperCase, BITS_PER_BYTE};
 
 use generate::register;
 
@@ -490,7 +490,7 @@ fn register_or_cluster_block_nightly(
         if region.fields.len() > 1 && !block_is_union {
             let (type_name, name) = match region.ident.clone() {
                 Some(prefix) => {
-                    (Ident::new(format!("{}Union", prefix.to_sanitized_pascal_case())),
+                    (Ident::new(format!("{}_UNION", prefix.to_sanitized_upper_case())),
                     Ident::new(prefix))
                 }
                 // If we can't find a name, fall back to the region index as a


wez:
Bummer.  I'm not super opinionated on the naming here.  Should we embrace uppercase and go for `FOO_BAR_UNION`?

Emilgardis:
I switched it over to `{}_UNION`, @jamesmunns you talked briefly about the naming in #192 
I ran this against svd2rust-regress and it's all clear, however I'm not sure if this naming is ideal

jamesmunns:
Hey, ACKing that I've seen this, and while I don't have strong opinions on naming (other than it should be consistent).

I think I need some time to grok what is going on here, and see if I can think of anything clever. I probably will be able to get to this by Monday (feel free to re-ping me if I have forgotten :) )

Emilgardis:
ping @jamesmunns

jamesmunns:
Argh. I knew there was something I was forgetting. Thanks @Emilgardis.

I'm not crazy about naming the unions as upper case, as that is typically used for constants, and is a bit of a wierd interface for users. I'd rather have the naming "weird" for these edge cases, and less weird for normal usage, rather than making it equally weird for everyone.

I don't want to hold up something that works, because I don't have particularly strong feelings about naming either.

I guess my one bike shed thought would be to introduce something like `Pascal_Snake_Case` for unions, which would look like this:

```text
<name>LTC0_PKA1_1</name> => Ltc0_Pka1_1
<name>LTC0_PKA_11</name> => Ltc0_Pka_11
```

Or replacing `_` with a word, such as `Sub`:

```text
<name>LTC0_PKA1_1</name> => Ltc0SubPka1Sub1
<name>LTC0_PKA_11</name> => Ltc0SubPkaSub11
```

But honestly this is even bad, as it introduces a new, weird naming convention, or inserts a lot of noise into words.

If no one really cares, lets go with UPPER_CASE for now, and have a follow up discussion before we move union support into the stable part of svd2rust

Emilgardis:
@jamesmunns re unions, we may want to move away from them, see #218

I'll wait a bit and then r+, let's see if anyone has a better solution

Emilgardis:
passed locally with svd2rust-regress, merging manually

PR #214, opened 2018-05-10
Provide library access to code generation

tones111:
This change implements the minimal API described in #207 to allow downstream build.rs scripts to generate the svd2rust output.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -439,10 +439,12 @@ mod generate;
 mod util;
 
 pub use util::Target;
+use util::build_rs;
 
 pub struct Generation {
     pub lib_rs: String,
     pub device_x: String,
+    pub build_rs: String,
 }
 
 type Result<T> = std::result::Result<T, SvdError>;
@@ -472,6 +474,7 @@ pub fn generate(xml: &str, target: &Target, nightly: bool) -> Result<Generation>
     Ok(Generation {
         lib_rs: lib_rs,
         device_x: device_x,
+        build_rs: build_rs().to_string(),
     })
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -439,12 +439,21 @@ mod generate;
 mod util;
 
 pub use util::Target;
-use util::build_rs;
 
 pub struct Generation {
     pub lib_rs: String,
+    pub device_specific: Option<DeviceSpecific>,
+
+    // Reserve the right to add more fields to this struct
+    _extensible: (),
+}
+
+pub struct DeviceSpecific {
     pub device_x: String,
     pub build_rs: String,
+
+    // Reserve the right to add more fields to this struct
+    _extensible: (),
 }
 
 type Result<T> = std::result::Result<T, SvdError>;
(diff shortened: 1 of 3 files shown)


tones111:
Rebased and updated for v0.13.0

Emilgardis:
cc @japaric I want to get an ok from you before merging

tones111:
Rebased for v0.13.1.  Instead of returning a tuple a struct with named fields is used to remove ambiguity over the tuple ordering.

tones111:
ping @Emilgardis @japaric 

I'm still interested in landing this PR.  rebased against the current master branch.

jacobrosenthal:
I found myself hacking this in as well. A lof of crates have shell scripts that attempt to cargo install binaries at specific versions. Until thats more doable, making form, svd2rust etc all library usable means we can lock to the correct version in the cargo manifest.

jacobrosenthal:
After I cleaned my version up, it became the naive solution exposing internals tried here and rejected. oops. Sorry for the noise. 

Ive successfully utilized this PR now, though it does seem to be missing the build_rs file generation.

jacobrosenthal:
One simple thing thats missing here is the build_rs output. I added it to my commit here and could pr:
https://github.com/jacobrosenthal/svd2rust/commit/0bcc4db32dedd0d9975c2bfd1a5ca6c00adcf09d

tones111:
My ultimate goal for this effort is to finish the implementation for https://github.com/adamgreig/stm32-rs/pull/5

Wouldn't generating the build.rs file prevent the crate from providing its own?  In the case of stm32-rs, the crate would be able to cover a large number of devices and the user selects the target with a feature flag.  In that case the build.rs file has a bunch of other work to accomplish like extracting the correct svd file from one of the archives in the repo.

jacobrosenthal:
From looking at the tool, it seems like the code path for cortex m targets at the very least wants the (currently hardcoded) build.rs to load the memory.x 

If you dont want it you could not save it. 

Also you should check out my script (which I wasnt intending to run as a build.rs) where I run a pinned programmatic version of rustfmt instead of using command
https://github.com/jacobrosenthal/efm32-rs/blob/upstream/tools/src/bin/gen.rs#L37-L43

src/lib.rs
```diff
@@ -422,7 +422,82 @@
...
+mod util;
+
+pub use util::Target;
+
+pub struct Generation {
+    pub lib_rs: String,
+    pub device_specific: Option<DeviceSpecific>,
+
+    #[allow(dead_code)]
+    reserved: (),
```
Emilgardis: To match the style of svd-parser, I'd want this to be changed to

```
// Reserve the right to add more fields to this struct
_extensible: (),
```

That way we don't use an unnecessary allow and clarify (in simple terms) why we do this

src/lib.rs
```diff
@@ -422,7 +422,82 @@
...
+}
+
+pub struct DeviceSpecific {
+    pub device_x: String,
+    pub build_rs: String,
+
+    #[allow(dead_code)]
+    reserved: (),
```
Emilgardis: Same as above

jacobrosenthal:
I tested this for my use cases and it is working well. Well done @tones111

jacobrosenthal:
Any last requests @therealprof or @japaric can we get a merge (and a point release)

Emilgardis:
try timed out, seems the version of the backtrace crate causing the previous ci fails has been yanked however.

therealprof:
bors try

PR #216, opened 2018-05-11
registers_with_uniq_names

astro:
Here is a proposal to solve one more issue I found in the [LPC13xx .svd file](https://github.com/posborne/cmsis-svd/blob/master/data/NXP/LPC13xx_svd_v1.svd): registers with duplicate names.

This PR does not simply drop the duplicate identifiers but tries to rename them by appending `2`, or `3`, and so on.

### DIFF
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -732,11 +732,15 @@ fn cluster_block(
     let reg_block = register_or_cluster_block(&c.children, defaults, Some(&mod_name), nightly)?;
 
     // Generate definition for each of the registers.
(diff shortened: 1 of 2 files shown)


Emilgardis:
Adding `_x` to a register seems to me to be a sub-optimal solution. What registers have duplicate names? Could this be solved with `<alternateGroup>` (which has not been implemented as of v0.13.1)

mathk:
Ping from triage: @astro  are you still looking for improvement idea ?

astro:
@mathk That would be helpful.

burrbull:
If this is still actual, please rebase.

PR #227, opened 2018-06-22
Derive common traits for write enums like for read enums

chrysn:
The enums are just values so this should do no harm, and it helps
building interfaces around the peripherals that do validation or
otherwise pass those values around.

Closes: https://github.com/japaric/svd2rust/issues/135


mathk:
bors try

mathk:
Ping from triage: @rust-embedded/tools Can that be merge or we need more approval ?

japaric:
bors r=ryankurte

PR #229, opened 2018-07-14
Remove #[deny(warnings)] from generated code

cgm616:
This declaration causes errors when using cargo doc, since some doc comments include lines that give doc warnings. This doesn't prevent the crate from building, so it's easy for these doc comment warnings to slip through to releases. However, because of this line (which can't be disabled later on), doc comment warnings are treated as errors and prevent documentation for the generated crate from building, as well as any other crate that depends on the generated crate.

For example, the `nrf51` crate is generated with `svd2rust`. This line causes the docs in that crate to error when building. The `nrf51-hal` crate depends on `nrf51`, and so it's docs don't build either. The `microbit` crate depends on `nrf51-hal`, so it's docs don't build either. Finally, any other crate that depends on any of those can't build docs. This makes it extraordinarily difficult to program using any of these crates.

For more detailed information about the above case, check out therealprof/nrf51#3.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,18 +1,18 @@
 build:stable:
   image: rust:stretch
   script:
-    - cargo build
+    - RUSTFLAGS="-D warnings" cargo build
 
 build:nightly:
   image: rustlang/rust:nightly
   script:
-    - cargo build
+    - RUSTFLAGS="-D warnings" cargo build
 
 regress:nightly:
   image: rustlang/rust:nightly
   script:
-    - cargo build --release
+    - RUSTFLAGS="-D warnings" cargo build --release
     - rustup component add rustfmt-preview
     - cd ci/svd2rust-regress
     - rm -rf ./output
-    - cargo run --release -- --long-test --format --verbose
+    - RUSTFLAGS="-D warnings" cargo run --release -- --long-test --format --verbose
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ env:
   global:
     - CRATE_NAME=svd2rust
     - TARGET=x86_64-unknown-linux-gnu
+    - RUSTFLAGS="-D warnings"
 
 matrix:
   include:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -31,11 +31,11 @@ install:
 
 test_script:
   - if [%APPVEYOR_REPO_TAG%]==[false] (
-      cargo build --target %TARGET%
+      RUSTFLAGS="-D warnings" cargo build --target %TARGET%
     )
 
 before_deploy:
-  - cargo rustc --target %TARGET% --release --bin svd2rust -- -C lto
+  - cargo rustc --target %TARGET% --release --bin svd2rust -- -C lto -D warnings
   - ps: ci\before_deploy.ps1
 
 deploy:
--- a/src/generate/device.rs
+++ b/src/generate/device.rs
@@ -9,7 +9,12 @@ use Target;
 use generate::{interrupt, peripheral};
 
 /// Whole device generation
-pub fn render(d: &Device, target: &Target, nightly: bool, device_x: &mut String) -> Result<Vec<Tokens>> {
+pub fn render(
+    d: &Device,
+    target: &Target,
+    nightly: bool,
+    device_x: &mut String,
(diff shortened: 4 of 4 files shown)


cgm616:
Since removing the `#[deny(warnings)]` attribute removes wanted functionality in some cases, it can be substituted with either of the following invocations of cargo:

```
> RUSTFLAGS="-D warnings" cargo build
> cargo rustc -- -D warnings
```

adamgreig:
Like the linked issue says, perhaps we could (additionally) escape `[` and `]` in generated doc strings to avoid the warning altogether.

cgm616:
I agree that escaping in doc strings is a smart improvement to make, if only to reduce the number of warnings when compiling documentation.

kjetilkjeka:
I'm not sure about this, please correct me if I'm wrong. However, I was under the impression that `#[deny(warnings)]` is problematic in libraries due to adding warnings in rustc is not considered a breaking change. Meaning that it's impossible to guarantee that the library will continue building for newer versions of rustc.

cgm616:
@kjetilkjeka You are correct; see here: https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md.

jamesmunns:
Hey all,

This seems like an overall reasonable change. I believe the "deny warnings" was generally an attempt to catch subtly-incorrect code generation, though @japaric  will have to have the final word on its historical usage.

If @japaric doesn't object, I would suggest modifying the CI to take advantage of one of the options @cgm616 suggested in https://github.com/japaric/svd2rust/pull/229#issuecomment-405033399, so that we can at least retain this at CI time, and remove it from regular usage.

Thanks for the comments and contribution!

Disasm:
I think we have two ways of resolving this:
1. Fix CI scripts according to the discussion. @cgm616, could you update your PR?
2. Close PR and leave everything as is. New warnings can be fixed on a regular issue/PR basis.

cc @japaric @jamesmunns

hannobraun:
@Disasm 
> 2\. Close PR and leave everything as is. New warnings can be fixed on a regular issue/PR basis.

New warnings would have to be fixed in every single svd2rust-generated crate. `#[deny(warnings)]` is an antipattern and needs to go, no doubt about it.

japaric:
Note that almost all (\*) warnings that come from crates.io (\*\*) dependencies are ignored (silenced) in Cargo builds, regardless of whether that dependency has `#![deny(warnings)]` in it or not. This is done exactly because of Rust stability promise ("building a crate with a newer stable release should always work").

(\*) the exception are warnings related to soundness issues. In that case you will always get the warning but `#[deny(warnings)]` will not turn the warning into an error.

(\*\*) this doesn't apply to path dependencies since you can modify the source and fix the warning.

What happened with `cargo doc` (more specifically with `rustdoc`) in this particular case was a bug that affected nightly and was fixed in nightly. Bugs that affect the statement in my first paragraph are considered high priority (release blocking) bugs and will be fixed before they reach stable.

With that out of the way, if you want to remove `#[deny(warnings)]` from the generated code go for it. I have no strong opinions on the matter and would prefer not to participate in such discussion.

therealprof:
@cgm616 Please fix the conflicts.

cgm616:
Just updated the CI scripts; I think I got all of them.

Disasm:
Thanks! Howerver, you removed `#[deny(warnings)]` from svd2rust generated code, not from svd2rust itself: you need to use `-D warnings` in places, where CI checks the code generated by svd2rust. Here is one of them: https://github.com/rust-embedded/svd2rust/blob/116a125d45c4c6dd7229c44c203fa218310c8c63/ci/script.sh#L13-L19
You may need to git-rebase your changes to cover all of the places.

Generated code is also cargo-checked in the regression tests:
https://github.com/rust-embedded/svd2rust/blob/116a125d45c4c6dd7229c44c203fa218310c8c63/ci/svd2rust-regress/src/svd_test.rs#L164-L169
They probably also need to be patched, but I'm not sure.

therealprof:
@cgm616 Sorry for the delay. Would you still be interested in addressing this? Please update the PR in this case.

therealprof:
@cgm616 Ping? Probably the best to close this and start over.

cgm616:
Sorry about this, I've been super busy lately. I'd still like to do this, so I'll close this PR and in the next week or so I'll try and open another.

Issue #232, opened 2018-09-03
Escape `[` and `]` in svd description before it is used in the doc attribute


Nicoretti:
Hi @all,

I generated some peripherals using **svd2rust** and stumbled upon an issue with **missing escape sequences** for characters which should be escaped **in the docstrings**.

**Issue:**
Missing escape sequence for `[` and `]` in generated docstrings.

**Impact:**
Documentation generation will fail.

**Steps to reproduce:**
1. Download [Keil.STM32F4xx_DFP.2.13.0.pack](https://keilpack.azureedge.net/pack/Keil.STM32F4xx_DFP.2.13.0.pack)
   The .pack file actually is a zip file
2. Unzip the .pack file and find the get STM32F41x.svd file
    (CMSIS/SVD/STM32F41x.svd)
3. Run the following svd2rust command
    `svd2rust -i STM32F41x.svd --target "cortex-m"`
4. Add the generated .rs file to some crate
5. Try to generate the documentation for that crate

**Summary**
If an .svd file contains a description with the characters `[` and `]`
e.g.:
```     
    <interrupt>
         <name>EXTI9_5</name>
         <description>EXTI Line[9:5] interrupts</description>
         <value>23</value>
     </interrupt>
```
**svd2rust** will generate a docstring missing necessary escapes for `[` and `]`:

```    
    #[doc = "23 - EXTI Line[9:5] interrupts"]
    EXTI9_5,
```
this will prevent `cargo doc` from generating documentation for the generated file:
```
error: `[9:5]` cannot be resolved, ignoring it...
   --> peripherals/src/lib.rs:259:23
    |
259 |     #[doc = "23 - EXTI Line[9:5] interrupts"]
    |                       ^^^ cannot be resolved, ignoring
    |
note: lint level defined here
   --> peripherals/src/lib.rs:3:12
    |
3   | #![ deny ( warnings ) ]
    |            ^^^^^^^^
    = note: #[deny(intra_doc_link_resolution_failure)] implied by #[deny(warnings)]
    = help: to escape `[` and `]` characters, just add '\' before them like `\[` or `\]`
```

**Attention:**
The suggested cargo/rustc fix will work, but be aware that in order to achiev `\[` you need to write `\\[` I guess this is due to the fact that we need create an escape sequence for the "docstring"  and not for the string itself.

A quick search brought up this spots in the code where such an error might happen: 
```
src/generate/interrupt.rs|59| #[doc = #description]
[...]

PR #233, opened 2018-09-03
Add escape sequences for special characters in docstrings

Nicoretti:
Cargo will fail to generate documentation if `[` and `[` are not
escaped in docstrings. Therefore an additional transformation
for escape sequences is done before the docstring is written to file.

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -31,7 +31,10 @@ pub fn render(
         rsize.next_power_of_two()
     };
     let rty = rsize.to_ty()?;
-    let description = util::respace(&register.description);
+    let description = util::normalize_docstring(format!(
+        "{}",
+        &register.description
+    ));
 
     let unsafety = unsafety(register.write_constraint.as_ref(), rsize);
 
@@ -374,7 +377,10 @@ pub fn fields(
                     }
                 }
 
-                let description = &f.description;
+                let description = util::normalize_docstring(format!(
+                     "{}",
+                     &f.description
+                ));
                 let sc = &f.sc;
                 r_impl_items.push(quote! {
                     #[doc = #description]
@@ -390,7 +396,7 @@ pub fn fields(
                     let mut vars = variants
                         .iter()
                         .map(|v| {
-                            let desc = v.description;
+                            let desc = util::normalize_docstring(format!("{}", &v.description));
                             let pc = &v.pc;
                             quote! {
                                 #[doc = #desc]
@@ -601,7 +607,7 @@ pub fn fields(
                 }
 
                 let pc_w = &f.pc_w;
-                let pc_w_doc = format!("Values that can be written to the field `{}`", f.name);
+                let pc_w_doc = util::normalize_docstring(format!("Values that can be written to the field `{}`", f.name));
 
                 let base_pc_w = base.as_ref().map(|base| {
                     let pc = base.field.to_sanitized_upper_case();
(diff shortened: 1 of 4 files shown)


Nicoretti:
Hi @all,

I started to write a patch for the [Issue](https://github.com/rust-embedded/svd2rust/issues/232) I have reported. Would be great if somebody could provide feedback in general and also about: 

* Are all spots where this should be applied (fixed) identified? [see ticket](https://github.com/rust-embedded/svd2rust/issues/232)
* Would the approach which I used for the single location valid to use for the other spots
 -> or do you have a better idea on how to fix it in general?
* Is the issue and the fix legit? -> Please let me know then I will proceed fixing the other locations

src/util.rs
```diff
@@ -261,3 +261,36 @@ pub fn only_registers(ercs: &[Either<Register, Cluster>]) -> Vec<&Register> {
...
+        assert_eq!(expected, normalize_docstring(format!("Some docstring{}-{}Line[0:10]", "\r", "\r\r")));
+        assert_eq!(expected, normalize_docstring(format!("Some docstring{}-{}Line[0:10]", "\t", "\t\t")));
+        assert_eq!(expected, normalize_docstring(format!("Some docstring{}-{}Line[0:10]", "\r\n\t", "\n\t\r")));
+        assert_eq!(expected, normalize_docstring(format!("Some docstring{}-{}Line[0:10]", "\r\n", "\n\r")));
+    }
+
+    #[test]
+    fn normalize_dcostring_escapes_brackets_in_docstring()
+    {
+        let expected = "Some docstring - Line\\[0:10\\]";
```
wez: this line appears to be redundant; the next line inlines the expectation

src/generate/register.rs
```diff
@@ -788,7 +794,7 @@ pub fn fields(
                 }
             });
 
-            let description = &f.description;
+            let description = util::normalize_docstring(format!("{}", &f.description));
```
wez: Can these single parameter calls be simplified to `util::normalize_docstring(&f.description)`?

Nicoretti: no, I can't  (not without changing the signature of `util::normalize_docstring()` but it could bechanged to `util::normalize_docstring(f.description)`. I'm unsure if its ok to consume/move that string. `format!` though will yield/create a new string.

Nicoretti: FYI `util::normalize_docstring(f.description)` is compiling without complaint, so it seems nobody is accessing it after it was moved. Still it kinda feels wired to me (but I'm pretty new to rust).

src/util.rs
```diff
@@ -261,3 +261,36 @@ pub fn only_registers(ercs: &[Either<Register, Cluster>]) -> Vec<&Register> {
         .collect();
     registers
 }
+
+/// Return a String which can be used as text in the #[doc] attribute
+pub fn normalize_docstring<S: Into<String>>(doc_string: S) -> String {
+    respace(&doc_string.into())
+        .replace("[", "\\[")
+        .replace("]", "\\]")
```
therealprof: Hm, but this will add more escaping in case the brackets are already escaped.

therealprof:
Meh, I just noticed that I accidentally re-created a solution for this, although in a slightly different way. 😞

Nicoretti:
@therealprof no worries, feel free to discard/reject/close this PR then.

Nicoretti:
this is already fixed by https://github.com/rust-embedded/svd2rust/pull/240

PR #234, opened 2018-09-06
docs: update cortex-m-rt version requirement

japaric:
generated crates are also compatible with v0.6.x of cortex-m-rt

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -53,9 +53,9 @@
 //! ```
 //!
 //! The resulting crate must provide an opt-in "rt" feature and depend on these crates:
-//! `bare-metal` v0.2.x, `cortex-m` v0.5.x, `cortex-m-rt` v0.5.x and `vcell` v0.1.x. Furthermore the
-//! "device" feature of `cortex-m-rt` must be enabled when the "rt" feature is enabled. The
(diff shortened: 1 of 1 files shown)


japaric:
r? @rust-embedded/tools (anyone)
cc @rust-embedded/cortex-m

japaric:
> cc @rust-embedded/cortex-m

to elaborate: there's no need to regenerate the device crate; simply update the version requirement in Cargo.toml.

japaric:
This turned out to be more complicated than I thought it to be.

It is true that device crates are compatible with both v0.5.x and v0.6.x. However, specifying ">=0.5.0, <0.7.0" as the version requirement of a device crate causes Cargo to compile both a v0.5.x version and a v0.6.x version when using that crate with cortex-m-rt v0.5.x. In other words; Cargo won't try to reduce the number of copies of cortex-m-rt in the dependency graph. Having two copies of cortex-m-rt leads to the compiler error we added in rust-embedded/cortex-m-rt#106 so this version requirement actually causes more problems.

therealprof:
Yeah, tell me about it... I've invested many hours already updating my few crates for compatibility with the latest cortex-m-rt version.

japaric:
closing in favor of #235

PR #235, opened 2018-09-17
re-export the `interrupt` attribute

japaric:
this is a breaking change because

- it bumps the cortex-m-rt version requirement to v0.6.4

- it changes the behavior of the `<device>::interrupt` item. This item was a
bang macro and now has become an attribute.

### DIFF
--- a/src/generate/device.rs
+++ b/src/generate/device.rs
@@ -126,7 +126,7 @@ pub fn render(d: &Device, target: &Target, nightly: bool, device_x: &mut String)
             pub use cortex_m::peripheral::Peripherals as CorePeripherals;
             #[cfg(feature = "rt")]
             pub use cortex_m_rt::interrupt;
-            pub use Interrupt as interrupt;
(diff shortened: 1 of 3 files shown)


therealprof:
LGTM

japaric:
Before landing this, the RFC rust-embedded/cortex-m-rt#122 needs to be accepted.

thejpster:
Pinging @whitequark as we'll want to re-spin tm4c12xx when this lands.

adamgreig:
Upstream cortex-m-rt v0.6.5 has landed with `#[interrupt]` now, so this is unblocked (but the dependency will need bumping to 0.6.5).

japaric:
Triage: this has been accepted by the cortex-m team and it's waiting for review.

japaric:
bors r=therealprof

japaric:
fixed the build error. re-r? @therealprof (or anyone else)

PR #236, opened 2018-09-18
Automatically generate feature gates for all peripherals

spacekookie:
This PR adds `cfg(feature = "...")` gates for all the peripherals that are generated. It means that if people only want to use certain peripherals, that they don't need to compile all of them which can take a very long time.

This PR is not yet finished (I want to see CI though :wink: ) – ideally there should also be a `Cargo.toml` skeleton that's generated or even just a `.txt` file which includes all available feature flags for users to expose in the `Cargo.toml`.


japaric:
Thanks for the PR, @spacekookie.

> This PR is not yet finished (I want to see CI though wink ) 

We have disabled running CI on PRs in this repo because the test suite takes very long to run and CI resources are shared with other repos under the rust-embedded org. However, I have you granted you permissions to test this manually by including [`bors try`](https://bors.tech/documentation/) in your comment. You can use that to manually test this PR.

> This PR adds cfg(feature = "...") gates for all the peripherals that are generated. 

In that case you should also update the test script to check the generated crates with `--all-features` to continue checking that code generation for peripherals works as intended.

https://github.com/rust-embedded/svd2rust/blob/77ce4d59fcd49eceedc9b9b26d8c99c59cdd04d3/ci/script.sh#L21

spacekookie:
bors try

jamesmunns:
bors try

spacekookie:
:see_no_evil: 

bors try

please

spacekookie:
cc/ @jamesmunns @japaric I renamed the file it generates again because it seems to be causing issues with CI.

I could always just call it a txt file that people copy stuff out of. However, I'm not 100% why `Cargo.toml` even collides with anything.

Maybe you have some better insights on how it works and/or how to proceed from here on out.

jamesmunns:
bors try

jamesmunns:
bors try

jamesmunns:
bors try

jamesmunns:
bors try

jamesmunns:
bors try

jamesmunns:
bors try

jamesmunns:
bors try

Apologies to anyone with email notifications.

jamesmunns:
Notes to self:

* Clean up the "cat Cargo.toml" line
* Talk to @therealprof about pinning dependencies
* Refactor the `Cargo.toml.bkp` dance?

jamesmunns:
bors try

jamesmunns:
bors try

jamesmunns:
bors try

jamesmunns:
bors try

jamesmunns:
@japaric it works for Cortex m, but causes problems for Msp430. Right now it shouldn't be a breaking change, since I've added a CLI flag to enable the behavior, so by default current behavior should be the same.

I can take some time tomorrow to try and close this out, or at least make it specifically only work for Cortex m for now.

japaric:
@jamesmunns Thanks for the update / clarification. My preference would be to *not* add features that affect codegen. The rationale is that most people use pre-generated crates (so they don't have to compile svd2rust when compiling a device crate) and there would be no way for them to pick between "all peripherals" and "peripherals behind Cargo features". This could to alternate device crates popping up on crates.io (e.g. nrf51 and nrf51-peripherals).

jamesmunns:
Okay, I really think I got it this time,

bors try

jamesmunns:
bors try

jamesmunns:
bors try

jamesmunns:
bors try p=10

jamesmunns:
bors retry

jamesmunns:
bors try

japaric:
bors try
(please?)

jamesmunns:
I think its because it worked one time? I can close this PR, squash my changes, and open a new one anyway, if thats easier.

jamesmunns:
Closing PR because of how gross this has gotten.

PR #243, opened 2018-10-18
Attempt to fix CI by bumping versions

therealprof:


therealprof:
bors try

therealprof:
Doesn't do the trick unfortunately.

PR #252, opened 2018-11-11
make non cortex-m targets compile on stable

japaric:
see individual commits for details

this is a breaking change


japaric:
bors r=therealprof

PR #221, opened 2018-05-17
Improve ci

Emilgardis:
This pr includes

- Support for the changes done in v0.13
- `--toolchain foo` to specify what toolchain to build the chips with (makes building locally easier, no need to rebuild `svd2rust-regress`)
- Add support for `--chip XXX YYY`
- Added validators to chip, manufacturer and architecture
- `-vv` will now output all stderr logs
- Added `--list` to quickly troubleshoot what test cases will be run
- A patched ATSAMD21G18A svd, thanks to @wez 

## Open "issues"
- Should we ignore case when filtering?
- Should a `--chip XXX` always run even if it is set to only run on `--long-test` (and possibly `--bad-test`)
- Add a `--` command to pass arguments to `svd2rust` and possibly also a way to pass arguments to `cargo check`
- Colorize output
- Add regress:beta test to gitlab-runner
- Move test cases to a or multiple config file(s) (e.g json/toml/yaml)

cc @jamesmunns

PR #254, opened 2018-11-11
Squashed commit implementing Feature Flags

jamesmunns:
The following changes, originally developed by @spacekookie, will generate feature flags for each peripheral. This allows users of a given `-pac` crate to "opt-in" to each peripheral they would like to use.

The primary motivator for this is to speed compile time, avoiding compilation of peripherals that may not be used at all. This can be especially painful when tweaking optimization profiles, cargo flags, etc. where multiple rebuilds may be necessary.

Additionally, as this change leans on feature flags so heavily, we have implemented generation of a reasonable `Cargo.toml`, with all of the feature flags generated as described above, listed appropriately in the Cargo.toml.

Additionally, this change makes the output of generated code more consistent. Previously, `cortex-m` targets would create a file called `lib.rs`, while `msp430` and `risc-v` targets would output the generated code to stdout. Now, all CPUs will create a `lib.rs` and `Cargo.toml`, and `cortex-m` targets will additionally create a `device.x` and `build.rs`.


jamesmunns:
bors try

jamesmunns:
bors try

jamesmunns:
Okay @japaric, if this tests passes, it should be ready to go, feature wise. The following are definitely left to do:

- [ ] Document in the changelog, as this is a breaking change
- [ ] Update any other docs, probably add a snippet covering the whole features thing
- [ ] Update `svd2rust-regress`, as some tweaks are probably now necessary (Cargo.toml generation, etc)

The following need review/decision before we should merge:

- [ ] Are the crate versions correct in [`cargo.rs`]? See `generate_dependencies()`
    - I know some of these are out of date, but they match what was used in the current CI
- [ ] Do we want to add any tags in every generated crate in [`cargo.rs`]? See `generate_header()`
- [ ] Should we include `edition = "2018` in [`cargo.rs`]? See `generate_header()`

[`cargo.rs`]: https://github.com/rust-embedded/svd2rust/blob/periph-2/src/generate/cargo.rs

jamesmunns:
bors try

therealprof:
As raised on IRC I am not happy with the approach due to the usability complications this causes for the mere benefit of some compile time gain.

In my opinion we should try to simplify the use of embedded rust crates and increase the portability instead of pinning everything down to the last bit requiring changes all over the map even if just a small detail in the hardware changes.

The way to achieve this portability is by unifying peripheral access creates (`PAC`) and re-exporting access to **all** registers under a common module name allowing simple access from hardware abstraction layer crates (`HAL`), board support package crates (`BSP`) as well as specialised drivers (e.g. the stm32f4-smoltcp ethernet TCP/IP stack) and applications without requiring knowledge about the specifically used PAC and the any features required or enabled previously and also providing that knowledge that in all drivers and applications requiring low-level access to the chip. It is also important to keep in mind that crates are considered incompatible if they're coming from different sources (e.g. crate and github repository) so even if the compile options and the correct source crate are known, the user still has to fight with those clashes!

spacekookie:
I'm not entirely sure why this is a new PR now? You could have just squashed the commits in the end. Because this way it won't also actually not list me as a contributer to svd2rust when this gets merged :/

jamesmunns:
Hey @spacekookie, there were two major reasons why this is a new PR:

1. bors stopped cooperating on the other PR, so it was easier to make a new PR than troubleshoot bors
2. There were a lot of formatting changes upstream, so rebasing was fairly painful, leading me to choose a flat squash-and-merge

That being said, Number 2 was pretty much me being lazy, because I was working on this on a Sunday. The intent wasn't to remove your name from your contributions. I will take some time today and re-engineer the git history and handle the rebase so that your name stays in the commits. When that is ready, I'll force-push that back to this PR.

Thanks for commenting to let me know :)

Emilgardis:
@jamesmunns 
> Update svd2rust-regress, as some tweaks are probably now necessary (Cargo.toml generation, etc)

#221 includes the changes done in v0.13, namely the lib.rs output. It should be easy to build on that pr for svd2rust-regress to work properly

@therealprof 
> It is also important to keep in mind that crates are considered incompatible if they're coming from different sources (e.g. crate and github repository) so even if the compile options and the correct source crate are known, the user still has to fight with those clashes!

Could you clarify what this has to do with usability/portability and this pr?

jamesmunns:
@Emilgardis thanks! I'll take a look at that when I pick this back up.

This was decided to be pushed until after 0.14, due to the issues raised by @therealprof.

The gist of the issue is that if someone is a consumer of a higher level crate, such as a BSP or HAL, and needs to have access to a peripheral that isn't implemented at that layer, they have to "break the abstraction", and go find which -pac crate they are actually using, and activate the features there.

Disasm:
Ping from triage: any progress on this?

jamesmunns:
This is unlikely to be merged, for reasons discussed at the All Hands. I believe @spacekookie will submit an RFC covering the planned alternative.

spacekookie:
@jamesmunns the RFC is pretty much ready, where do I best post it?

PR #261, opened 2018-11-27
Only export Interrupt as interrupt when 'rt' is enabled

### LINKED ISSUE #260: Renaming Interrupt to interrupt should be conditional
@japaric 

The generated code contains a pub use:
```
pub use self::Interrupt as interrupt;
```
This only makes sense when using the "rt" so it should be conditional:
```
#[cfg(feature = "rt")]
```

rnestler:
It only makes sense to export interrupt if the interrupt! macro is
exported as well, since it's the only thing using it.
Fixes #260


rnestler:
@japaric

PR #262, opened 2018-11-28
programmatic library support

jacobrosenthal:
This a probably the laziest take with the bare minimum of exposing the target and generate functions.

Maybe https://github.com/rust-embedded/svd2rust/pull/214/files is better? I dont have idiomatic api rust opinions yet..

Also, I dont have opinions on the rust fmt commit but my client is set up for it so I just put it in a separate commit. Just look at the split commit.

cc @jamesmunns (ignore irc links this is my cleanest copy)

My usage looks like

```
use svd2rust::{generate, target::Target, util::build_rs};

...

    let mut device_x = String::new();
    let items = generate::device::render(&device, &Target::CortexM, true, &mut device_x).unwrap();

    //save other files
    writeln!(File::create("device.x").unwrap(), "{}", device_x).unwrap();
    writeln!(File::create("build.rs").unwrap(), "{}", build_rs()).unwrap();

    //form lib.rs and save
    let lib = format!("{}", quote!(#(#items)*));

...	

```

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -34,20 +34,18 @@ fn commit_info() -> String {
 
 fn commit_hash() -> Result<String, IgnoredError> {
     Ok(try!(String::from_utf8(
-        try!(
-            Command::new("git")
-                .args(&["rev-parse", "--short", "HEAD"])
-                .output()
-        ).stdout,
+        try!(Command::new("git")
+            .args(&["rev-parse", "--short", "HEAD"])
+            .output())
+        .stdout,
     )))
 }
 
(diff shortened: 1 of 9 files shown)


jacobrosenthal:
Oops. I just found https://github.com/rust-embedded/svd2rust/pull/207 where I see exposing internals was already done and backed out again, which led to https://github.com/rust-embedded/svd2rust/pull/214


Sorry for the noise, ill close this and keep pushing on that pr

PR #263, opened 2018-11-30
Escape brackets on all description doc strings

craigjb:
Some SVD files have square brackets '[]' in all different description fields for registers, register values, etc.  These files were causing errors when running `cargo doc` due to the un-escaped brackets in doc strings.

It looks like svd2rust already has a util function to deal with this, but it wasn't applied to all description fields.


jacobrosenthal:
Thanks! I need this for an efm32 svd https://github.com/jacobrosenthal/efm32-rs/blob/master/svd/EFM32HG309F64.svd

therealprof:
bors try

PR #264, opened 2018-12-03
Add support for `Access::WriteOnce` registers

hugwijst:
Currently doesn't enforce the write-once nature of the registers, but
treats them as write-only.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -183,7 +183,9 @@ pub fn access_of(register: &Register) -> Access {
         if let Some(ref fields) = register.fields {
             if fields.iter().all(|f| f.access == Some(Access::ReadOnly)) {
                 Access::ReadOnly
-            } else if fields.iter().all(|f| f.access == Some(Access::WriteOnly)) {
+            } else if fields.iter().all(|f| f.access == Some(Access::WriteOnce)) {
+                Access::WriteOnce
+            } else if fields.iter().all(|f| f.access == Some(Access::WriteOnly) || f.access == Some(Access::WriteOnce)) {
                 Access::WriteOnly
             } else {
                 Access::ReadWrite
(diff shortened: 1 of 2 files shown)


Disasm:
How is this WriteOnly property enforced?

src/util.rs
```diff
@@ -183,7 +183,9 @@ pub fn access_of(register: &Register) -> Access {
         if let Some(ref fields) = register.fields {
             if fields.iter().all(|f| f.access == Some(Access::ReadOnly)) {
                 Access::ReadOnly
-            } else if fields.iter().all(|f| f.access == Some(Access::WriteOnly)) {
+            } else if fields.iter().all(|f| f.access == Some(Access::WriteOnce)) {
+                Access::WriteOnce
+            } else if fields.iter().all(|f| f.access == Some(Access::WriteOnly) || f.access == Some(Access::WriteOnce)) {
```
Disasm: I think this second `f.access == Some(Access::WriteOnce)` comparison is redundant

hugwijst: This checks that all the fields are either `WriteOnly` or `WriteOnce`, instead of only `WriteOnly`. A register that has both `WriteOnly` and `WriteOnce` fields is still `WriteOnly` and not `ReadWrite`.

hugwijst:
> How is this WriteOnce property enforced?

It isn't, and I couldn't come up with a good way to enforce it on a type level :(

src/util.rs
Disasm: > A register that has both `WriteOnly` and `WriteOnce` fields is still `WriteOnly` and not `ReadWrite`.

Yes, but any register with both `WriteOnly` and `WriteOnce` will be processed by first `WriteOnce` `if` statement. The next `if` statement will process just `WriteOnly` registers. Or not?

Disasm:
> It isn't, and I couldn't come up with a good way to enforce it on a type level :(

Maybe `write(self, ...)` instead of `write(&mut self, ...)` will do what you need.

mathk:
bors try

PR #265, opened 2018-12-06
Consistent svd2rust output

jamesmunns:
NOTE: I believe there is still documentation that will need to be updated to represent this. I wanted to make sure my CI changes still work without the other changes (like Cargo.toml) after pulling these changes


jamesmunns:
bors try

jamesmunns:
bors try

PR #266, opened 2018-12-06
v0.14.0

japaric:
this commit:

- fixes cortex-m version in documentation (v0.5.x is the latest available
version)

- updates docs to note that svd2rust writes generated code to lib.rs for all
targets

- updates the CHANGELOG

after this lands we can make a new release of svd2rust

r? @rust-embedded/tools
cc @rust-embedded/cortex-m


japaric:
Published: https://crates.io/crates/svd2rust/0.14.0

PR #269, opened 2018-12-17
ci: remove rustfmt for speed

jacobrosenthal:
Im trying to find a way to speed up ci as were currently failing due to 50min timeout

It appears its the freescale entry in the matrix, most likely because it has 133 svds that it tests, thats 1/5 of the total svds upstreamed there.

It doesnt actually look like our code is the issue. We seem to run in a second or so. It seems like 
 maybe cargo check and rustfmt split the time. Why rustfmt at all? its probably on average 15 seconds * 133 svds, that would be a dent 

Can someone bors this and see what happens?

cc @therealprof

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -14,8 +14,6 @@ test_svd() {
 
     mv lib.rs src/lib.rs
 
-    # ignore rustfmt errors
-    rustfmt src/lib.rs || true
     popd
 
     cargo check --manifest-path $td/Cargo.toml


therealprof:
bors try

jamesmunns:
Doesn't look like a significant change over previous successful builds: https://travis-ci.org/rust-embedded/svd2rust/builds

jamesmunns:
There are a couple groups, such as Atmel, which are completely empty, and wasting CI time. Perhaps removing these would help overall build time?

Or we could pare down the Freescale test set, which takes > 40 minutes by itself.

jacobrosenthal:
Hrm, Disagree. Definitely disappointed, and well have to see on average how much less, but it could approach 5 minutes less which is enough to start passing the occasional over 50min fail.

therealprof:
@rust-embedded/tools Any objections to disabling rustfmt? Does anyone know why it was added in the first place (errors are ignored anyway and we don't keep the formatted files anyway)?

Emilgardis:
rustfmt was introduced in https://github.com/rust-embedded/svd2rust/pull/104/commits/027e79dcc19e544faa547c36e9a6a2ec4179d083 to make errors format nicely. I can see how it can be useful for determining what went wrong, but I personally think we should switch over to svd2rust-regress for ci which could handle this more easily than the sh script. My suggestion would be to only format crates that fail.

PR #270, opened 2019-01-23
Add an "ownership" feature to the generated crate

oli-obk:
Activating the feature will make the generated crate require `&mut` references to the registers in order to modify them. This allows us to manage hardware in an owned manner without breaking anyone's code

### DIFF
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -54,6 +54,14 @@ pub fn render(
                 unsafe { &*#name_pc::ptr() }
             }
         }
+
+        impl DerefMut for #name_pc {
+            type Target = #base::RegisterBlock;
+
+            fn deref_mut(&mut self) -> &mut #base::RegisterBlock {
+                unsafe { &*#name_pc::ptr() }
+            }
+        }
     });
 
     // Derived peripherals do not require re-implementation, and will instead
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -51,6 +51,11 @@ pub fn render(
             pub fn ptr_mut() -> *mut #base::RegisterBlock {
                 #address as *mut _
             }
+            /// Destroys the zero sized handle in favor of a `&'static mut` reference to the
+            /// register block
+            pub fn into_static_mut(self) -> &'static mut #base::RegisterBlock {
+                unsafe { &mut*#name_pc::ptr_mut() }
+            }
         }
 
         impl Deref for #name_pc {
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -54,7 +54,7 @@ pub fn render(
             /// Destroys the zero sized handle in favor of a `&'static mut` reference to the
             /// register block
             pub fn into_static_mut(self) -> &'static mut #base::RegisterBlock {
-                unsafe { &mut*#name_pc::ptr_mut() }
+                unsafe { &mut *#name_pc::ptr_mut() }
             }
         }
 
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -39,6 +39,8 @@ pub fn render(
         pub struct #name_pc { _marker: PhantomData<*const ()> }
 
         unsafe impl Send for #name_pc {}
+        #[cfg(feature = "ownership")]
+        unsafe impl Sync for #name_pc {}
 
         impl #name_pc {
             /// Returns a pointer to the register block
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -47,6 +47,10 @@ pub fn render(
             pub fn ptr() -> *const #base::RegisterBlock {
                 #address as *const _
             }
+            /// Returns a mutable pointer to the register block
+            pub fn ptr_mut() -> *mut #base::RegisterBlock {
+                #address as *mut _
+            }
         }
 
         impl Deref for #name_pc {
@@ -61,7 +65,7 @@ pub fn render(
             type Target = #base::RegisterBlock;
 
             fn deref_mut(&mut self) -> &mut #base::RegisterBlock {
-                unsafe { &*#name_pc::ptr() }
+                unsafe { &mut *#name_pc::ptr_mut() }
             }
         }
     });
--- a/src/generate/device.rs
+++ b/src/generate/device.rs
@@ -80,7 +80,7 @@ pub fn render(d: &Device, target: &Target, nightly: bool, device_x: &mut String)
         extern crate bare_metal;
         extern crate vcell;
 
-        use core::ops::Deref;
+        use core::ops::{Deref, DerefMut};
         use core::marker::PhantomData;
     });
 
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -44,6 +44,20 @@ pub fn render(
         reg_impl_items.push(quote! {
             /// Modifies the contents of the register
             #[inline]
+            #[cfg(feature = "ownership")]
+            pub fn modify<F>(&mut self, f: F)
+            where
+                for<'w> F: FnOnce(&R, &'w mut W) -> &'w mut W
+            {
+                let bits = self.register.get();
+                let r = R { bits: bits };
+                let mut w = W { bits: bits };
+                f(&r, &mut w);
+                self.register.set(w.bits);
+            }
+            /// Modifies the contents of the register
+            #[inline]
+            #[cfg(not(feature = "ownership"))]
             pub fn modify<F>(&self, f: F)
             where
                 for<'w> F: FnOnce(&R, &'w mut W) -> &'w mut W
@@ -86,6 +100,18 @@ pub fn render(
         reg_impl_items.push(quote! {
             /// Writes to the register
             #[inline]
+            #[cfg(feature = "ownership")]
+            pub fn write<F>(&mut self, f: F)
+            where
+                F: FnOnce(&mut W) -> &mut W
+            {
+                let mut w = W::reset_value();
+                f(&mut w);
+                self.register.set(w.bits);
+            }
+            /// Writes to the register
+            #[inline]
+            #[cfg(not(feature = "ownership"))]
             pub fn write<F>(&self, f: F)
             where
                 F: FnOnce(&mut W) -> &mut W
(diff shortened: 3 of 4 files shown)


src/generate/peripheral.rs
```diff
@@ -39,12 +39,23 @@ pub fn render(
...
                 #address as *const _
             }
+            /// Returns a mutable pointer to the register block
+            pub fn ptr_mut() -> *mut #base::RegisterBlock {
+                #address as *mut _
+            }
+            /// Destroys the zero sized handle in favor of a `&'static mut` reference to the
+            /// register block
+            pub fn into_static_mut(self) -> &'static mut #base::RegisterBlock {
+                unsafe { &mut*#name_pc::ptr_mut() }
```
therealprof: Is that going to compile (&mut* back to back with the name)? Anyway, I'd prefer consistency.

```suggestion
                unsafe { &mut *#name_pc::ptr_mut() }
```

jamesmunns:
We should probably test at least one SVD builds with the feature enabled in CI (or test all with the flag, and test at least one without).

japaric:
Conceptually, I'm not in favor of this change. Registers are *not* normal
memory; they are a shared-memory, "IPC" mechanism for communication between
the core processor and a peripheral. A register (`&Register`) is semantically
close to `&AtomicUsize` (minus atomic CAS/RMW operations) where the peripheral
*always* owns one such shared reference.

IIUC, what's being proposed here is to convey the idea that having `&mut
Register` gives you *exclusive* access to the register for writes and RMW ops. I
find this wrong for a few reasons:

- Having `&mut Register` does *not* mean that the value won't change while you
  have this exclusive reference. The hardware may change the value at any time.

- Doing a read (`fn read(&self)`) *can* change the value behind `&Register`
  (see clear on read bits). This defeats the idea that a `&mut Register` is
  required to modify a register.

- Multiple registers may provide access to the same peripheral state. Meaning
  that code like the one shown below would compile *and panic*, even though you
  would not expect it from just looking at the signature.

``` rust
// e.g. SomeRegister = GPIOA_ODR, OtherRegister = GPIOA_BSRR (STM32F1)
fn foo(reg: &SomeRegister, not_an_alias_of_reg: &mut OtherRegister) {
    let old = reg.read().bits();

    not_an_alias_of_reg.write(|w| w.bits(!0));

    assert_eq!(old, reg.read.bits()); // may panic
}
```

This PR also implements `Sync` for the ZST singleton. This would let people
store the ZST in a `static`. Paired with a `fn read(&self)` that may clear bits
this can easily result in hard to debug race conditions.

I also see an implementation problem: enabling the `ownership` feature changes
the API and can break existing code. Cargo features should never do that as
they are incremental and have "union" semantics across the dependency graph.
This is a problematic example:

``` rust
fn foo(reg: &SomeRegister) {
    reg.write(|w| w.bits(0));
}
```

This function no longer compiles if the `ownership` feature is enabled, which
*any* crate is the dependency graph is free to do.

The only change in this PR I would be OK accepting is adding the
`into_static_mut` method.

@oli-obk could you say more about the use cases you have for this?

therealprof:
@japaric Good points.

If such a change was made I would actually recommend that read also takes a `&mut` instead of just a reference for the reasons you mentioned. That would not get rid of the mentioned aliasing problem but actually I haven't seen any approach that would and there're plenty of ways running into aliasing troubles (bitbanding, mirrored/shadowed/coupled registers, DMA, Nordic "shortcuts", ...).

In my eyes this change would still be an improvement because it restricts what one can do with a reference.

oli-obk:
> Having &mut Register does not mean that the value won't change while you
have this exclusive reference. The hardware may change the value at any time.

That's not much of an issue. I think we can accept this. This is no different than accessing the next element of an `std::sync::channel::Receiver`

> Doing a read (fn read(&self)) can change the value behind &Register
(see clear on read bits). This defeats the idea that a &mut Register is
required to modify a register.

Is that information exposed by the svd files? If so, we can just mark the relevant read functions as `&mut self`, too

> Multiple registers may provide access to the same peripheral state. Meaning
that code like the one shown below would compile and panic, even though you
would not expect it from just looking at the signature.

Hardware is funky enough, I think it's fine to expect this. I mean, GPIO ports can easily cause other GPIO ports to change their value.

> This PR also implements Sync for the ZST singleton. This would let people
store the ZST in a static. Paired with a fn read(&self) that may clear bits
this can easily result in hard to debug race conditions.

See above for `read(&mut self)` in that case

> I also see an implementation problem: enabling the ownership feature changes
the API and can break existing code. Cargo features should never do that as
they are incremental and have "union" semantics across the dependency graph.
This is a problematic example:

I am fully aware of this. Crates using the ownership method are inherently incompatible with crates that do. While the feature is not additive, I see no reason this will ever be a problem in practice. If you change your crate's HAL crate, downstream crates will most definitely break. The same would go for the "ownership" feature.

> The only change in this PR I would be OK accepting is adding the
into_static_mut method.

I can start with that and see how far I get.

> @oli-obk could you say more about the use cases you have for this?

@phil-opp can probably explain our use case better

pftbest:
Suppose I have two output pins, PD1 and PD2. To change their value they both need to write to GPIOD.bsrr register. They both can't own this register and can't share &mut reference to it, so they will have to construct it out of thin air using `ptr_mut` function, then dereference it and pass it as a &mut reference to `write` function. In this scenario, we can have two mutable references pointing to the same memory address. Will this be considered undefined behavior in Rust?

My only concern is that the ownership model will forbid valid designs like the one above (which would be perfectly safe when using shared references) by introducing undefined behavior.

therealprof:
@pftbest Isn't the pointer creation out of thin air already happening in a variety of places despite the possibility of using regular references?  Especially in your example because these don't use RMW and can update individual bits atomically. 

Also in the `stm32f0-hal` we recently changed everything to use `&mut` specifically to prevent sharing of addresses because that would potentially allow clashing accesses. Also places which require to do RMW at shared resource now need to pass a `cs` to make it perfectly safe.

oli-obk:
> My only concern is that the ownership model will forbid valid designs like the one above (which would be perfectly safe when using shared references) by introducing undefined behavior.

What we're doing there is wrapping the register in a `Mutex` to get synchronized access in a Rusty way. Our `Mutex` type is just a glorified interrupt disabling critical section.

pftbest:
@oli-obk, You would also need `RefCell` to go along with the `Mutex`, so you already have both time and space overhead for an operation that is already atomic and doesn't require any of this. My fear is that ownership model will not allow to opt-out of this overhead. Because even unsafe code is not allowed to break aliasing rules and cause UB, so no amount of unsafe code will help me to remove this `Mutex` and `RefCell`.

oli-obk:
Why would I need a `RefCell`?

Also: unsafe code may break aliasing if there's an `UnsafeCell` involved.

oli-obk:
> When they will switch to mutable references there will be no api left for me to use. I think it will be good if the old functions taking shared references remain even if under unsafe and/or different names.

This is all gated behind a cargo feature. Your code won't be affected at all

> You need RefCell because the Mutex you described is a recursive mutex. And in Rust recursive mutex cannot give you the mutable reference, so you ether have to use RefCell to get it, or put a boolean flag inside your Mutex to make it non-recursive. Both have the same overhead.

Right. Our Mutex does have an `AtomicBool` flag with additional memory overhead.

Still, this is only needed for the very few registers that are shared permanently between separate code parts. Most of our code is `Mutex`-free.

japaric:
@therealprof

> That would not get rid of the mentioned aliasing problem but actually I haven't seen any approach that would and there're plenty of ways running into aliasing troubles (bitbanding, mirrored/shadowed/coupled registers, ~~DMA~~, Nordic "shortcuts", ...).

This "aliasing problem" is not a problem that needs to be fixed, IMO; it's a fact of reality that we have to accept and convey in the API. That's why I think that, at this layer, `Atomic*`-like shared (interior) mutability is the right call. Aliasing due to bitbanding and coupled registers abide to Rust aliasing rules under the current shared mutability API; they are OK. (I don't know what Nordic "shortcuts" are)

> DMA

(DMA deals with normal memory; you should not break the aliasing rules on normal memory or you'll run into UB)

@oli-obk

> Is that information exposed by the svd files?

The SVD spec does provide a way to encode this information (see [`<readAction>`](http://www.keil.com/pack/doc/CMSIS/SVD/html/elem_registers.html#elem_fields)) but most vendors don't encode this information in their SVDs. The way the SVD spec is written omitting `<readAction>` means that the read operation has no side effects so, unless `<readAction>` is correctly and consistently used in the *whole* file, tweaking `svd2rust` codegen based on their presence will yield incorrect results.

> Hardware is funky enough, I think it's fine to expect this.

I'm more concerned about this code exploding if rustc ever decides to do more pre-LLVM optimizations.

>  While the feature is not additive

But Cargo features *must* be additive.

> If you change your crate's HAL crate, downstream crates will most definitely break

This would require a semver bump which downstream users have to explicitly opt into and they expect breaking changes in that scenario. OTOH, nobody expects that enabling a feature, either directly or through a dependency, will break their code.

Let's not use Cargo features for this; it's not the right mechanism.

@pftbest

> In this scenario, we can have two mutable references pointing to the same memory address. Will this be considered undefined behavior in Rust?

This is already UB today. Ask anyone in the Unsafe Code Guidelines team and they'll answer yes.

If aliased `&mut-` references paired with volatile operations doesn't produce UB (read: miscompile) today is because rustc+LLVM is doing something sensible, but there's no guarantee that this behavior will remain if (a) rustc backend changes (e.g. Cranelift) or (b) rustc starts doing (more) optimizations before handing things to LLVM.

therealprof:
@japaric 

> This "aliasing problem" is not a problem that needs to be fixed, IMO; it's a fact of reality that we have to accept and convey in the API.

That sounds totally weird coming from you. :-D

> That's why I think that, at this layer, Atomic*-like shared (interior) mutability is the right call.

Atomics only protect the exact address you're writing to, so they're useful for interrupt handlers and threading. If you have the ability to alias in hardware, then you'll need hardware support to provide coherency. That is hard enough to do when you _just_ have caches interfering (which is one reason why newer ARM Cortex-Ms don't do bitbanding anymore) but I highly doubt that this is implemented for anything else (maybe with the exception of DMA).

> Aliasing due to bitbanding and coupled registers abide to Rust aliasing rules under the current shared mutability API; they are OK.

Is that so? To be perfectly correct IMHO an API would need to capture all pointers providing access to the same resource. Otherwise you could never be sure that someone is not changing the facts behind your back (with safe code) while Rust thinks you're the sole owner and only guarantees that no one is accessing it by the handle you're holding.

> (I don't know what Nordic "shortcuts" are)

Magic registers which trigger modifications to multiple other registers when written to.

mathk:
Ping from triage: a lots of discussion and no activity for more than 18 days. Do we need to bring that in the weekly meeting ? @japaric

oli-obk:
I think we'll try to work with the existing API for now instead of trying to shoehorn svd2rust into ours. I'll see how far I get with this and come back with my findings.

PR #278, opened 2019-02-03
This fixes issue-174 and issue-175 by outputting the proper write fun…

ammoniak:
…ctions even if the reset-value is missing.

I'm currently working with my Arduino Due and try to get things working. I wrote a quick hello world-application that blinks. All the registers that were needed work with this patch. I did not do much further testing and I'm relatively new to Rust.

I tested with the sam3x8e-SVD-file from the Atmel homepage.

The changes are based on another older pull request, which cannot be applied to the current repository and seem to work fine.

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -96,7 +96,6 @@ pub fn render(
             .reset_value
             .or(defs.reset_value)
             .map(util::hex);
-            //.ok_or_else(|| format!("Register {} has no reset value", register.name))?;
         if let Some(rv) = rv {
             reg_impl_items.push(quote! {
                 /// Writes to the register
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -83,18 +83,7 @@ pub fn render(
     }
 
     if access == Access::WriteOnly || access == Access::ReadWrite {
-        reg_impl_items.push(quote! {
-            /// Writes to the register
-            #[inline]
-            pub fn write<F>(&self, f: F)
-            where
-                F: FnOnce(&mut W) -> &mut W
-            {
-                let mut w = W::reset_value();
-                f(&mut w);
-                self.register.set(w.bits);
-            }
-        });
+
 
         mod_items.push(quote! {
             /// Value to write to the register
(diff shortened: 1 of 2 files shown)


src/generate/register.rs
```diff
@@ -106,23 +95,52 @@ pub fn render(
         let rv = register
             .reset_value
             .or(defs.reset_value)
-            .map(util::hex)
-            .ok_or_else(|| format!("Register {} has no reset value", register.name))?;
+            .map(util::hex);
+            //.ok_or_else(|| format!("Register {} has no reset value", register.name))?;
```
therealprof:
```suggestion
```

ammoniak:
Is there anything I can do to help with the review?

kjetilkjeka:
This is great! Unspecified reset values are valid in SVD but unforuntantely have made the whole device fail to compile when used with svd2rust.

Any reason this is not a "no-brainer"?

therealprof:
> Any reason this is not a "no-brainer"?

Well, a test case exploiting/using that would be great.

Other than said, it looks good to me, but in order to approve it I'd need to somehow verify the effect and correctness of this change myself for which I haven't found the time yet.

kjetilkjeka:
I made a PR on posborne/svd-cmsis to add some SVD files where this is the case to enable basic automated testing against them. https://github.com/posborne/cmsis-svd/pull/67

I've added a file that contains definitions that doesn't compile unless using the code from this PR. Hope that can be of some convinience.

[ATSAMV71Q21B.svd.txt](https://github.com/rust-embedded/svd2rust/files/2872405/ATSAMV71Q21B.svd.txt)

jamesmunns:
I took a quick look at the code, it seems like you are silently degrading `write` to be the same as `modify` when a reset value is not specified?

in general:

* Write starts with the reset value, and applies any values you change to the reset value
* Modify starts with the current value, and applies any values you change to the current value

If this is correct, is there a reason you want to use `write` instead of just `modify`? Personally, I am not a fan of silently changing the behavior, as this would be hard to catch for most users of a `-pac` crate, and would make the behavior of `-pac` crates inconsistent (`write` does one thing on some platforms, and another on other platforms). They would have to look at the code to see what has been generated.

Please let me know if I misunderstood your intent, or what the code actually does.

src/generate/register.rs
```diff
@@ -106,23 +95,51 @@ pub fn render(
...
+        else{
+            reg_impl_items.push(quote! {
+                /// Writes to the register
+                #[inline]
+                pub fn write<F>(&self, f: F)
+                where
+                    F: FnOnce(&mut W) -> &mut W
+                {
+                let bits = self.register.get();
+                let mut w = W { bits: bits };
```
jamesmunns: How does this work if `access == Access::WriteOnly`? This is likely to load invalid data as `bits`, and may cause problems on some targets.

ammoniak:
I'm trying to get the CI working on my system. Therefore sorry for the "force-push". I'll create another branch and do my testing there.

japaric:
Semantically, this PR is a breaking change. The svd2rust documentation specifies: "the write method writes some value to the register using a **single**, volatile STR instruction.". This PR changes `write` into a RMW operation in some cases; there's no precedent for the svd2rust API having different semantics based on the content of the SVD file (we only ever add / remove API based on the information available) and I'd rather keep it that way ("no suprises and backward compatible when improving the SVD file").

I have mentioned before (either on this issue tracker or IRC) that when resetValue info is missing there should be some "write" method (not exactly with that name)  that does a single volatile write and whose argument forces you to specify all the (non-reserved) bit fields of the register (there are a few ways to do that, but a struct would be the easiest).

ammoniak:
That makes sense. I try to come up with a new function (maybe something like `reset`?). It will need some time, I think I have to become more familiar with the whole tool.

ammoniak:
I think we can close this PR and if I (or someone else) comes up with a better solution, we open a new one.

kjetilkjeka:
In the short term, would it be acceptable to remove everything inside the "else" block? This would mean that there would not be a `write` method for registers with unknown reset value. 

Alternatively, what about not compiling anything for the register, issue a warning and moving along with the file?

Anything would be better than the current solution where the whole file fails to compile if the reset value is undefined.

ammoniak:
Most of the registers of my SVD don't have a reset value, which means I can't write any registers. I'm not sure how that's better.

mathk:
Ping from triage: @ammoniak Do you still want to close this PR or consider the change from @kjetilkjeka.

@therealprof @jamesmunns is the @kjetilkjeka  proposition valuable for sdv2rust?

ammoniak:
Most of the registers in my SVD file miss the reset value, so "ignoring" those, won't help at all.

mathk:
So I suggest that you close this PR.

PR #282, opened 2019-02-08
Fixed all remaining clippy lints (except the complexity ones)

therealprof:

### DIFF
--- a/src/generate/device.rs
+++ b/src/generate/device.rs
@@ -9,7 +9,12 @@ use Target;
 use generate::{interrupt, peripheral};
 
 /// Whole device generation
-pub fn render(d: &Device, target: &Target, nightly: bool, device_x: &mut String) -> Result<Vec<Tokens>> {
+pub fn render(
+    d: &Device,
+    target: Target,
+    nightly: bool,
+    device_x: &mut String,
+) -> Result<Vec<Tokens>> {
     let mut out = vec![];
 
     let doc = format!(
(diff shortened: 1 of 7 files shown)


jamesmunns:
Hm. Tests passed but Bors died :(

bors retry

jamesmunns:
Hey @japaric, it looks like bors is having some issues here. Builds are passing, but bors is cancelling the merge.

jamesmunns:
Update: Somehow I have bors control here, but I am not a codeowner, so I can't approve the review. Sorry for all the noise. We'll have to wait for @rust-embedded/tools.

src/generate/device.rs
```diff
@@ -100,21 +105,19 @@ pub fn render(d: &Device, target: &Target, nightly: bool, device_x: &mut String)
...
     out.extend(interrupt::render(target, &d.peripherals, device_x)?);
 
-    let core_peripherals: &[&str];
-
-    if fpu_present {
-        core_peripherals = &[
-            "CBP", "CPUID", "DCB", "DWT", "FPB", "FPU", "ITM", "MPU", "NVIC", "SCB", "SYST", "TPIU"
-        ];
+    let core_peripherals = if fpu_present {
+        vec![
```
japaric: why did this become a vec? You can write, for example:

``` rust
let x: &[_] = if true {
    &[0, 1]
} else {
    &[2, 3, 4]
};
```

src/generate/device.rs
```diff
@@ -49,7 +54,7 @@ pub fn render(d: &Device, target: &Target, nightly: bool, device_x: &mut String)
         });
     }
 
-    match *target {
+    match target {
```
japaric: Just a comment: this opts into the "match ergonomics" feature which may raise the MSRV -- not that it was ever documented or guaranteed.

src/generate/device.rs
therealprof: Was recommended by clippy. Dunno. I can undo it.

therealprof: Oh, I remember. rustc complained about not being able to assign arrays with different lengths.

therealprof: TIL I can assign arrays of various lengths by declaring the type as `&[_]`. How odd. :-D

PR #288, opened 2019-05-02
Make lib.rs accessible to dependers

vojtechkral:
Hi,
  we are trying to use `svd2rust` from a `build.rs` file. As of now, cargo build-dependencies can't be binaries and using `svd2rust` as a dependency yields an empty crate. `svd2rust` has to be installed manually for our `build.rs` to work.

This is a proposal to fix that. I'm not sure this patch is the right way, but it seems all we need is access to the `generate()` function and we're good to go.

Thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,10 @@ name = "svd2rust"
 repository = "https://github.com/japaric/svd2rust"
 version = "0.14.0"
 
+[lib]
+name = "svd2rust"
+path = "src/lib.rs"
+
 [[bin]]
 doc = false
 name = "svd2rust"


therealprof:
I though there was a similar PR to this but I can't find it. Sounds good to me in principle but I'm not a very keen user of build.rs so I'd like to hear other opinions.

CC @rust-embedded/tools

Emilgardis:
#214 did indeed do this. I can't reproduce the mentioned issue when using latest changes (currently unreleased, candidate for 0.15) in `cargo 1.33.0 (2019-02-12)`. Unless there's a good reason to add this I think this should be closed.

vojtechkral:
Yes, using the current `master`, it seems to work as expected. I must've gotten 0.14 and master mixed up, sorry.

Is there an ETA for 0.15?

PR #290, opened 2019-05-07
Try updating base image to bionic to fix build woes

therealprof:


therealprof:
bors try

Emilgardis:
The error (AFAICS) is from cross, furthermore, the image cross uses.

This is rather frustrating but it is how things are right now. We'll need to provide (or find) images for cross that has our needs. See https://github.com/rust-embedded/cross/pull/164#issuecomment-466658347 for why we're not updating glibc in "official" cross images. 

As for where this error comes from, I'm not sure. Could be `git`

I think this is something we need to prioritize for svd2rust, as it's currently the biggest blocker for PRs. May be something for 2019 tools team objectives.

therealprof:
TBH I've no idea why we need this amount of magic and fluff anyway instead of simply using the standard rust toolchain to test compile the generated PACs.

Disasm:
I see two solutions:
* Fixing all the targets in `cross` and rolling out `cross-2.0` release. I created [a PR for relevant targets](https://github.com/rust-embedded/cross/pull/270).
* Temporarily disabling cross-builds (replacing with cargo-builds) in `svd2rust` and optionally goto step1. This involves disabling some non-Linux CI targets.

therealprof:
While cross-compilation is nice and very useful, I don't really get why we to support cross-compilation for ancient or nostalgic targets. If someone wants to build for CentOS 7 why not use a CentOS 7 docker container and compile natively?

Disasm:
@therealprof No, Centos 7 is too modern, we need Centos 5 :smile: 
https://github.com/rust-lang/rust/blob/548add7f61bfcbe3bea3f5ccefb53c84da8fefe4/src/ci/docker/dist-x86_64-linux/Dockerfile

therealprof:
Withdrawing due to #297 fixing the issue.

PR #294, opened 2019-05-10
Add constants for minimum and maximum interrupt values.

bjc:
This adds two new constants in the generated crate: `INTERRUPT_MIN` and `INTERRUPT_MAX` which can be used in higher level crates for generating statically allocated arrays and other structures that can store per-interrupt data.

### DIFF
--- a/src/generate/interrupt.rs
+++ b/src/generate/interrupt.rs
@@ -24,6 +24,16 @@ pub fn render(
 
     let mut interrupts = interrupts.into_iter().map(|(_, v)| v).collect::<Vec<_>>();
     interrupts.sort_by_key(|i| i.value);
+    let min_intnr = util::unsuffixed(u64(interrupts
+        .iter()
+        .min_by(|a, b| a.value.cmp(&b.value))
+        .unwrap()
+        .value));
+    let max_intnr = util::unsuffixed(u64(interrupts
+        .iter()
+        .max_by(|a, b| a.value.cmp(&b.value))
+        .unwrap()
+        .value));
 
     let mut root = vec![];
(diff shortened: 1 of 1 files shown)


ryankurte:
thanks for the pr, seems good to me! @rust-embedded/tools any thoughts?

src/generate/interrupt.rs
```diff
@@ -24,6 +24,16 @@ pub fn render(
 
     let mut interrupts = interrupts.into_iter().map(|(_, v)| v).collect::<Vec<_>>();
     interrupts.sort_by_key(|i| i.value);
+    let min_intnr = util::unsuffixed(u64(interrupts
+        .iter()
+        .min_by(|a, b| a.value.cmp(&b.value))
+        .unwrap()
```
therealprof: Hah! https://travis-ci.org/rust-embedded/svd2rust/jobs/535678429

It seems that interrupt definitions are not mandatory so this `unwrap()` will choke on the non-existing interrupts.

Not sure whether the better assumption is to assume a value of `0` for the constants or not define the constants at all.

bjc: I don't have a good solution for this. I don't like either of the options above. Assuming a value of `0` for both minimum and maximum is potentially a valid scenario (although obviously probably not). It's at best ambiguous.

And not having the constants prevents the kinds of consumers I was hoping to write. We could wrap them in an `Option`, but that has similar problems for me, although it's the most obviously correct option.

I'll close this for now, and will re-open it if I have a better solution. In the mean time, the code I had that was using this no longer requires it, so it's less important for me.

Issue #289, opened 2019-05-04
Support for 64-bit registers


laanwj:
Some Kendryte K210 (RISC-V RV64GC) peripherals use 64-bit registers (see https://github.com/riscv-rust/k210-pac/issues/1), and, *I think*, also requires 64-bit accesses to these. However, the current version of `svd2rust` seems to ignore `<size>` specification for registers completely.

PR #295, opened 2019-05-14
Add support for 64-bit registers

Disasm:
Some registers on K210 chip are 64-bit, so it's better to declare them as u64 in SVD for the reasons mentioned here: https://github.com/riscv-rust/k210-pac/issues/1#issuecomment-489311245
At the moment, svd2rust forbids 64-bit register declarations. This PR fixes this.

This change can cause silent bugs on platforms without 64-bit memory access operations due to the need for proper access sequence to 64-bit registers with two 32-bit accesses.

Closes https://github.com/rust-embedded/svd2rust/issues/289

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -462,7 +462,7 @@ main() {
             # Test RISC-V K210
             pushd $td
 
-            $cwd/target/$TARGET/release/svd2rust --target riscv -i $td/k210.svd
+            RUST_BACKTRACE=1 $cwd/target/$TARGET/release/svd2rust --target riscv -i $td/k210.svd
             mv $td/lib.rs $td/src/lib.rs
             rustfmt $td/src/lib.rs || true
 
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -427,9 +427,12 @@ main() {
                 cd $td &&
                     curl -LO \
                          https://github.com/pftbest/msp430g2553/raw/v0.1.0/msp430g2553.svd
-                 cd $td &&
-                     curl -LO \
-                          https://raw.githubusercontent.com/riscv-rust/e310x/master/e310x.svd
+                cd $td &&
+                    curl -LO \
+                         https://raw.githubusercontent.com/riscv-rust/e310x/master/e310x.svd
+                cd $td &&
+                    curl -LO \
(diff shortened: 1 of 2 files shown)


laanwj:
I tried this on https://github.com/riscv-rust/k210-pac/pull/14 and there was no change at all :/ How to make it pick up that registers for a peripheral are 64 bit?

Disasm:
Seems like `<size>64</size>` is just ignored at that point. I tested on CLINT and it works fine there.

laanwj:
The difference is that CLINT has both 64-bit and 32-bit registers, DMAC and KPU have *only* 64-bit registers, so it'd make sense to specify it at the peripheral level. But yes I guess specifying it for every register individually would work.

Disasm:
This `<size>` specification at the peripheral level is valid, so it should be supported as well. However, I'm not sure at which point this problem happens.

therealprof:
@Disasm Is there anything we should/can test here wrt to 64bit support?

therealprof:
bors try

Disasm:
@therealprof I hope that CI is enough. Before this PR 64-bit register declarations led to `svd2rust` errors, but now they shouldn't. So I assume that this PR shouldn't break anything that wasn't broken before.

therealprof:
Shouldn't we at least have some files enabled with this PR which exercise this new part of the code?

Disasm:
Ah, OK, I'll add them.

Disasm:
Done!

ci/script.sh
```diff
@@ -455,6 +458,17 @@ main() {
             popd
 
             cargo check --manifest-path $td/Cargo.toml
+
+            # Test RISC-V K210
+            pushd $td
+
+            $cwd/target/$TARGET/release/svd2rust --target riscv -i $td/k210.svd
```
therealprof:
```suggestion
            RUST_BACKTRACE=1 $cwd/target/$TARGET/release/svd2rust --target riscv -i $td/k210.svd
```

I've just issued #298 to get some feedback in case `svd2rust` crashes during CI. Mind adding that here as well?

Disasm: Sure!

PR #297, opened 2019-05-21
CI: use Cargo instead of Cross

japaric:
this removes the Linux GNU targets used to produce binaries but we still have
the MUSL builds which will work everywhere because they statically linked
binaries

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,12 +36,12 @@ matrix:
     - env: TARGET=x86_64-unknown-linux-gnu VENDOR=OTHER
       rust: nightly
 
-    # Stable, used to produce binary relreases
-    # Linux
+    # Stable, used to produce binary releases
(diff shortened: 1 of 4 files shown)


jamesmunns:
bors try

japaric:
Forgot the `rust target add` in ci/install.sh.

therealprof:
bors try

japaric:
> error occurred: Failed to find tool. Is `musl-gcc` installed?

This package is not the [package whitelist](https://github.com/travis-ci/apt-package-safelist/blob/master/ubuntu-trusty). Let's skip the musl builds on CI; we can build the binary releases locally and upload the binaries.

> https://travis-ci.org/rust-embedded/svd2rust/jobs/535456598

A network error ...

therealprof:
bors try

PR #298, opened 2019-05-22
Added RUST_BACKTRACE=1 to enable backtraces

therealprof:
If svd2rust fails its always useful to get some information about the where and why


therealprof:
bors try

therealprof:
bors r=ryankurte

PR #256, opened 2018-11-17
WIP: Register derivedFrom and svd updates

jamesmunns:
This PR extends https://github.com/rust-embedded/svd2rust/pull/193 (edit: and https://github.com/rust-embedded/svd2rust/pull/190), and also brings support for the changes made in `svd-parser`, primarily understanding the `RegisterCluster` type.

PR #300, opened 2019-06-01
bump svd v0.7, rm ref

burrbull:
bump svd-parser to v0.7, rm `ref` in match patterns without changing functionality


src/main.rs
```diff
@@ -76,7 +71,7 @@ fn run() -> Result<()> {
         }
     }
 
-    let device = svd::parse(xml);
+    let device = svd::parse(xml).unwrap();
```
burrbull: Not sure how to convert Error here.

Emilgardis:
Thanks for the pr!

This is a duplicate of #256, if you did a pr to [register-derived-from](/rust-embedded/svd2rust/tree/register-derived-from) fixing merge conflicts I would be happy to approve that :)

Issue #285, opened 2019-02-23
use #[derive()] attributes for read/write/etc


burrbull:
External traits crate:
```
pub trait ResetValue {
    #[doc = r" Reset value of the register"]
    const RESET_VALUE: Self;
}

pub trait BitOffset {
    #[doc = r" Bit offset of the field"]
    const OFFSET: u8;
}

pub trait WriteRegister<W>
where
    W: ResetValue,
{
    #[doc = r" Writes to the register"]
    #[inline]
    fn write<F>(&self, f: F)
    // &self
    where
        F: FnOnce(&W) -> &mut W;
}

pub trait ReadRegister<R> {
    #[doc = r" Reads the contents of the register"]
    fn read(&self) -> R;
}

pub trait ResetRegister<W>: WriteRegister<W>
where
    W: ResetValue,
{
    #[doc = r" Writes the reset value to the register"]
    fn reset(&self) {
        // &self
        self.write(|w| w)
    }
}

pub trait ModifyRegister<R, W> {
    #[doc = r" Modifies the contents of the register"]
    #[inline]
    fn modify<F>(&self, f: F)
    // &self
    where
        for<'w> F: FnOnce(&R, &'w mut W) -> &'w mut W;
}

pub trait SingleBit<'a, W>
where
    Self: core::marker::Sized + BitOffset,
{
    #[doc = r" Sets the field bit"]
    fn set_bit(self) -> &'a mut W {
        self.bit(true)
    }
    #[doc = r" Clears the field bit"]
    fn clear_bit(self) -> &'a mut W {
        self.bit(false)
    }
    #[doc = r" Writes raw bits to the field"]
    #[inline]
    fn bit(self, value: bool) -> &'a mut W;
}
```

External macro crate:
```
extern crate proc_macro;

use crate::proc_macro::TokenStream;

use quote::quote;

use syn::{parse_macro_input, DeriveInput};

#[proc_macro_derive(WriteRegister)]
pub fn write_register_derive(input: TokenStream) -> TokenStream {
    let input = parse_macro_input!(input as DeriveInput);

    let regname = &input.ident;

    let expanded = quote! {
        impl crate::WriteRegister<W> for #regname {
            #[inline]
            fn write<F>(&self, f: F) // &self
            where
                F: FnOnce(&mut W) -> &mut W,
            {
                let mut w = W::RESET_VALUE;
                f(&mut w);
                self.register.set(w.bits);
            }
        }
    };
[...]

PR #303, opened 2019-07-03
[WIP] DO NOT MERGE: Code reduction

therealprof:
Reduce code by eliminating duplication and boilerplating

### DIFF
--- a/src/generate/device.rs
+++ b/src/generate/device.rs
@@ -6,7 +6,7 @@ use crate::errors::*;
 use crate::util::{self, ToSanitizedUpperCase};
 use crate::Target;
 
-use crate::generate::{interrupt, peripheral};
+use crate::generate::{interrupt, peripheral, generic};
 
 /// Whole device generation
 pub fn render(
@@ -141,6 +141,8 @@ pub fn render(
             });
         }
     }
+
+    out.extend(generic::render()?);
 
     for p in &d.peripherals {
         if target == Target::CortexM && core_peripherals.contains(&&*p.name.to_uppercase()) {
--- a/src/generate/generic.rs
+++ b/src/generate/generic.rs
@@ -9,12 +9,12 @@ pub fn render() -> Result<Vec<Tokens>> {
     code.push(quote! {
         /// Single bit read access proxy
         trait BitR {
-            #[doc = r" Returns `true` if the bit is clear (0)"]
+            #[doc = r"Returns `true` if the bit is clear (0)"]
             #[inline]
             fn bit_is_clear(&self) -> bool {
                 !self.bit()
             }
-            #[doc = r" Returns `true` if the bit is set (1)"]
+            #[doc = r"Returns `true` if the bit is set (1)"]
             #[inline]
             fn bit_is_set(&self) -> bool {
                 self.bit()
@@ -25,14 +25,14 @@ pub fn render() -> Result<Vec<Tokens>> {
 
         /// Single bit write access proxy
         trait BitW<'a, W> {
-            #[doc = r" Sets the field bit"]
+            #[doc = r"Sets the field bit"]
             fn set_bit(self) -> &'a mut W
             where
                 Self: core::marker::Sized,
             {
                 self.bit(true)
             }
-            #[doc = r" Clears the field bit"]
+            #[doc = r"Clears the field bit"]
             fn clear_bit(self) -> &'a mut W
             where
                 Self: core::marker::Sized,
@@ -43,6 +43,16 @@ pub fn render() -> Result<Vec<Tokens>> {
             fn bit(self, value: bool) -> &'a mut W
             where
                 Self: core::marker::Sized;
+        }
+
+        /// Wiggle in the specified value into the given bits with mask and the offset and return
+        /// the new value
+        #[inline]
+        const fn set_bits(bits: u32, mask: u32, offset: u8, value: u32) -> u32 {
+            let mut bits = bits;
+            bits &= !((mask as u32) << offset);
+            bits |= ((value & mask) as u32) << offset;
+            bits
         }
     });
 
--- a/src/generate/generic.rs
+++ b/src/generate/generic.rs
@@ -1,0 +1,28 @@
+use quote::Tokens;
+
+use crate::errors::*;
+
+/// Generates generic bit munging code
+pub fn render() -> Result<Vec<Tokens>> {
+    let mut code = vec![];
+
+    code.push(quote! {
+        /// Single bit read access proxy
+        trait BitR {
+            #[doc = r" Returns `true` if the bit is clear (0)"]
+            #[inline]
+            fn bit_is_clear(&self) -> bool {
+                !self.bit()
+            }
+            #[doc = r" Returns `true` if the bit is set (1)"]
+            #[inline]
+            fn bit_is_set(&self) -> bool {
+                self.bit()
+            }
+
+            fn bit(&self) -> bool;
+        }
+    });
+
+    Ok(code)
+}
--- a/src/generate/generic.rs
+++ b/src/generate/generic.rs
@@ -9,30 +9,33 @@ pub fn render() -> Result<Vec<Tokens>> {
     code.push(quote! {
         /// Single bit read access proxy
         trait BitR {
-            #[doc = r"Returns `true` if the bit is clear (0)"]
+            /// Returns `true` if the bit is clear (0)
             #[inline]
             fn bit_is_clear(&self) -> bool {
                 !self.bit()
             }
-            #[doc = r"Returns `true` if the bit is set (1)"]
+            /// Returns `true` if the bit is set (1)
             #[inline]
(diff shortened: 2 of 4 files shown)


therealprof:
Those two changes reduce the file size of a `cargo fmt`ed STM32F042 SVD `lib.rs` from `7893118` to `5911219` and compile time from `26.57s`/`24.28s` to `17.00s`/`16.95s` (unoptimized/optimized).

therealprof:
Now down to `5097551` bytes and `14.43s`.

therealprof:
CC @adamgreig @jamesmunns @japaric 

Interested to learn what others think. Changes should be API compatible with what we have now but `stm32-rs` would need some adjustments.

It would also be possible to move all of that code (and more) into a generic `memory-mapped` crate allowing for far more features and custom implementations of a memory mapped register interface but for now I wanted to keep things simple to play around with some changes.

burrbull:
`reset_value()` should be marked as `const fn` or replaced with
`pub const RESET_VALUE: W = W { bits: #rv };`

therealprof:
@burrbull I guess there're a few non-breaking uncontroversial changes which could still be applied. But at the moment this is not going to fly; it's not particularly pretty but also breaks the HAL impls due to the requirement to import the traits.

I'm still looking into ways to generically express the individual registers in a way that would allow us have a default trait impl for `read`, `write` and `modify` which would yield the largest reduction in code since those are completely identical. But I haven't found the key to do that yet.

burrbull:
I also made many experiments with traits to make default implementation and could not find good solution of the issue. The main problem is access to bits `self.register.set(w.bits);`

The only way I could reduce code size is using [macros](https://github.com/burrbull/svd2rust/tree/macros).

burrbull:
This is the best I could think of.
```
pub trait FromBits<T> {
    fn from_bits(bits: T) -> Self;
}

pub trait ResetValue {
    #[doc = r" Reset value of the register"]
    const RESET_VALUE: Self;
}

/// Reads the contents of the register
pub trait ReadRegister<R, T>: core::ops::Deref<Target=vcell::VolatileCell<T>>
where
    R: FromBits<T>,
    T: Copy,
{
    fn read(&self) -> R {
        R::from_bits((*self).get())
    }
}

/// Writes to the register
pub trait WriteRegister<W, T>: core::ops::Deref<Target=vcell::VolatileCell<T>>
where
    W: ResetValue + core::ops::Deref<Target=T>,
    T: Copy,
{
    #[inline]
    fn write<F>(&self, f: F)
    where
        F: FnOnce(&mut W) -> &mut W,
    {
        let mut w = W::RESET_VALUE;
        f(&mut w);
        (*self).set(*w);
    }
}

/// Writes the reset value to the register
pub trait ResetRegister<W, T>: WriteRegister<W, T>
where
    W: ResetValue + core::ops::Deref<Target=T>,
    T: Copy,
{
    fn reset(&self) {
        self.write(|w| w)
    }
}

/// Modifies the contents of the register
pub trait ModifyRegister<R, W, T>: core::ops::Deref<Target=vcell::VolatileCell<T>>
where
    R: FromBits<T>,
    W: FromBits<T> + core::ops::Deref<Target=T>,
    T: Copy,
{
    #[inline]
    fn modify<F>(&self, f: F)
    where
        for<'w> F: FnOnce(&R, &'w mut W) -> &'w mut W
    {
        let bits = (*self).get();
        let r = R::from_bits(bits);
        let mut w = W::from_bits(bits);
        f(&r, &mut w);
        (*self).set(*w);
    }
}
#[macro_export]
macro_rules! deref_cell {
    ($REG:ty, $T:ty) => {
        impl core::ops::Deref for CR1 {
            type Target = vcell::VolatileCell<$T>;
            fn deref(&self) -> &Self::Target {
                &self.register
            }
        }
    }
}

#[macro_export]
macro_rules! w {
    ($T:ty) => {
        pub struct W {
            bits: $T,
        }

        impl core::ops::Deref for W {
            type Target = $T;
            fn deref(&self) -> &Self::Target {
                &self.bits
            }
        }

        impl crate::FromBits<$T> for W {
            fn from_bits(bits: $T) -> Self {
                Self { bits }
            }
        }
    }
}

#[macro_export]
macro_rules! r {
    ($T:ty) => {
        pub struct R {
            bits: $T,
        }

        impl crate::FromBits<$T> for R {
            fn from_bits(bits: $T) -> Self {
                Self { bits }
            }
        }
    }
}
```

Usage:
```
pub struct CR1 {
    register: vcell::VolatileCell<u32>,
}
crate::deref_cell!(CR1, u32);

pub mod cr1 {
    crate::w!(u32);
    crate::r!(u32);

    impl crate::ResetValue for W {
        const RESET_VALUE: Self = Self {bits: 0};
    }
    impl crate::WriteRegister<W, u32> for super::CR1 {}
    impl crate::ResetRegister<W, u32> for super::CR1 {}
    impl crate::ReadRegister<R, u32> for super::CR1 {}
    impl crate::ModifyRegister<R, W, u32> for super::CR1 {}
}
```

burrbull:
#258

burrbull:
You forgot about 1bit-wise `variants`.

burrbull:
New implementation: #306

therealprof:
@burrbull That is very interesting! Lets see whether @japaric is happy to take the vcell change or whether we have to move it someplace else but I like your use of traits; feel free to move more of my code around if you see the chance; I'm not particularly fond of the set...bits functions, they're rather hacky.

burrbull:
I would bring `set_bits_` back to `svd2rust`, but I don't know how to access to them:
`crate::` works if functions on top level.

But in this case they must be `use`d from subpackage: `https://github.com/stm32-rs/stm32-rs/blob/master/stm32f0/src/lib.rs`

therealprof:
@burrbull Maybe it'd be possible to use traits and default impls by using the trick here: https://github.com/dtolnay/inherent

burrbull:
It depends on proc_macro2. It will take time to compile dependencies.
If use procedural macros, we can use traits+derive as I proposed in #285 (of course with some changes).

therealprof:
@burrbull I did not necessarily mean the crate but the idea to avoid having to `use` the crates implementating the traits everywhere. We're already already generating code in svd2rust so it'd be simple to incorporate the `inherent` behaviour, though I'm not sure it is worth it.

burrbull:
Incorporate where?
Into svd2rust (I don't see how it can help)
or in generated code?

therealprof:
In the generated code.

burrbull:
I've applied `cargo expand` on example in `inherent` repo and what I get:

```rust
#![feature(prelude_import)]
#![no_std]
#[prelude_import]
use ::std::prelude::v1::*;
#[macro_use]
extern crate std as std;
mod types {
    use inherent::inherent;
    trait Trait {
        fn f(self);
    }
    pub struct Struct;
    impl Struct {
        pub fn f(self) {
            <Self as Trait>::f(self)
        }
    }
    impl Trait for Struct {
        fn f(self) {}
    }
}
fn main() {
    types::Struct.f();
}
```

it added
```
impl Struct {
   ...
}
```

therealprof:
Yep, I was think along the lines of adding traits for register Read, Write, ... whatever and impls for that traits and those `inherent` style proxy functions hiding away the trait impls for the outside. Not sure it's worth it in terms of code size and/or compile time though.

burrbull:
Looks like `inherent` doesn't support default impls in traits. So it senseless for us.

therealprof:
I'm still looking at the mechanism, not this particular crate.

burrbull:
I have idea:
https://gist.github.com/6bd0d6a20a52c0704c767c8bb3813c6a

The main change is to use generic writer/reader structures (AKA `W`, `R`) instead of
generating them in each module.

therealprof:
Nice!

burrbull:
In any case, we must release `svd2rust` as is (or with minimal changes) and think how to improve it in future.

therealprof:
Yes, we should do a release soon...

burrbull:
https://gist.github.com/rust-play/3a76e3d892ee06fff295ed52d57ab3a8

burrbull:
I made some changes:

https://gist.github.com/24a0d1a2e13bb464a7c579a07902a3c6

First part: #331

burrbull:
I made draft of WProxy. See from 200 line:
https://gist.github.com/fb6ecf0f0b9c23e9d1d1ea42637cc7f9
This variant with inline offset. As we saw it require some additional memory in debug.

With const generics it will be:
```rust
    ///Writer Proxy
    pub struct WProxy<'a, U, REG, N, FI, const: O: u8> {
        w: &'a mut W<U, REG>,
        _field: marker::PhantomData<(FI, N)>,
    }
```

burrbull:
I tried to use associated size type where it's possible and got very nice code:
https://github.com/burrbull/svd2rust/tree/associated
but firmware size increased ~5% for both release and debug build. :cry:

therealprof:
Let's close this superseded PR.

PR #305, opened 2019-07-11
register access traits

burrbull:
I tried to implement my thoughts from #303 .

All common traits I put to separate crate (including BitR/W): https://github.com/burrbull/ra

@therealprof  Can you test this variant?

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -51,6 +51,7 @@ main() {
     echo 'cortex-m = "0.5.0"' >> $td/Cargo.toml
     echo 'cortex-m-rt = "0.5.0"' >> $td/Cargo.toml
     echo 'vcell = "0.1.0"' >> $td/Cargo.toml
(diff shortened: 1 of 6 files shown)


therealprof:
@burrbull I must say I'm not too hot about the use of macros here since it makes everything even more intransparent than it already is. Also it would be great if you could split up the code cleanup into a separate PR which we could apply right away.

burrbull:
Ok. The main question is what do you think about separation of traits in a other crate? And if you agree, how should it be called?

therealprof:
@burrbull I thought about this. Doing so has both advantages and disadvantages so I'm not quite clear whether this is desirable. My plan was to first make it work and then look into whether it makes sense splitting it out. I would also think that maybe we should get rid of VCell while we are at it plus I'm not too happy about the implicit register offset magic, so I was thinking about putting the absolute addresses into associated constants...

I haven't had much time the last week to look into this unfortunately.

PR #306, opened 2019-07-12
Register access traits

burrbull:
Move read/write/modify and set/clear bits into traits:

https://github.com/japaric/vcell/pull/8


src/util.rs
```diff
@@ -264,10 +264,10 @@ impl U32Ext for u32 {
     fn to_ty(&self) -> Result<Ident> {
         Ok(match *self {
             1 => Ident::new("bool"),
-            2...8 => Ident::new("u8"),
-            9...16 => Ident::new("u16"),
-            17...32 => Ident::new("u32"),
-            33...64 => Ident::new("u64"),
+            2..=8 => Ident::new("u8"),
```
therealprof: Would you mind moving out all those unrelated cleanups into a separate PR?

burrbull: Done #307

src/generate/register.rs
```diff
@@ -166,16 +143,25 @@ pub fn render(
...
             }
         });
     }
 
-    if access == Access::WriteOnly || access == Access::ReadWrite {
+    if write_access {
         mod_items.push(quote! {
+            /// Wiggle in the specified value into the given bits with mask and the offset and return the new value
+            #[inline]
+            pub const fn set_bits(bits: #rty, mask: #rty, offset: u8, value: #rty) -> #rty {
```
burrbull: @therealprof  I moved `set_bits` here. I think it is the best trade-off: one `const set_bits` per writable register as they can have different bit-size.

burrbull:
I've done everything I could.

therealprof:
I think it should be possible to have the traits defined in vcell directly in the generated code if you put them into a module.

burrbull:
Where is the use of traits if each generated crate have own incompatible traits?
We will need to re-export them several times:
1) In header of common (family) pac crate
2) In header of hal trait
3) In the place where they are used

With `vcell` (or other common crate) only 3 item is actual. And potentially they can be used for inheritance.

therealprof:
> Where is the use of traits if each generated crate have own incompatible traits?

Each `RegisterBlock` (and everything underneath) is completely freestanding and incompatible anyway.

> With vcell (or other common crate) only 3 item is actual. And potentially they can be used for inheritance.

Sure but first you'll need to get your PR included there and I'm somewhat suspicious this is going to happen. 😏

For development it's a lot easier if everything is self-contained...

I'm not too hot on having traits at all; I only care about real source code and compile time reduction, so macro tricks do not apply.

burrbull:
Speaking of speed. How you tested it? Clean after each attempt? Can you repeat your tests with this variant?

burrbull:
> Sure but first you'll need to get your PR included there
It does not have to be  exactly `vcell`. It can be crate in workspace of svd2rust.

therealprof:
No, I can't test this "as-is" due to the custom `vcell` repository. That's exactly what I'm talking about, it makes everything so much more complicated... I might be able to patch it but the version resolution plays havoc if you're trying to use github repositories in multiple crates...

burrbull:
https://doc.rust-lang.org/nightly/edition-guide/rust-2018/cargo-and-crates-io/replacing-dependencies-with-patch.html

therealprof:
Thanks, I know. You probably missed the important bit here (emphasis mine):

> I might be able to patch it but the version resolution plays havoc if you're trying to use github repositories in **multiple crates**

burrbull:
Sorry. I'll try do something tomorrow.

burrbull:
Try to test this branch:
https://github.com/burrbull/svd2rust/tree/internal-traits

PR #310, opened 2019-07-14
Coalesce write implementation into a single line down from three

therealprof:


therealprof:
bors r=ryankurte

PR #313, opened 2019-07-15
Register derived from (yet another try resolve)

burrbull:


burrbull:
bors r=ryankurte

burrbull:
cc @therealprof

therealprof:
Last time I manually merged it into the other branch, but the right thing to do would be the original PR owner to do a rebase. Otherwise the history is going to be really messy.

PR #314, opened 2019-07-15
Register derived from3

burrbull:
Merge all merges.

cc @therealprof 

How about such variant?


therealprof:
bors try

therealprof:
Let's give it a spin, I'll still wait for @jamesmunns to say something because IIRC there's also a story behind the story...

PR #315, opened 2019-07-18
Remove Mask offset consts

burrbull:
cc @therealprof 

Example:
```rust
pub unsafe fn bits(self, value: u8) -> &'a mut W {
    self.w.bits &= !(0x1f << 8);
    self.w.bits |= ((value as u32) & 0x1f) << 8;
    self.w
}

pub fn dbl(&self) -> DBLR {
    let bits = ((self.bits >> 8) & 0x1f) as u8;
    DBLR { bits }
}
```

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -516,7 +516,7 @@ pub fn fields(
                     #[doc = #description]
                     #[inline]
                     pub fn #sc(&self) -> #pc_r {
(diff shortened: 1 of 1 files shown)


therealprof:
Hm, that makes me wonder for the reason why it was more complicated before...

therealprof:
bors try

PR #316, opened 2019-07-19
move reset_value

burrbull:
move `reset_value` to be public method of register struct (not `W`).

It's breaking change, but I think this can't break existent code.


therealprof:
Nice, I like it.

PR #141, opened 2017-09-11
Replace #[inline(always)] by #[inline] due to the later being conside…

therealprof:
…red harmful

Signed-off-by: Daniel Egger <daniel@eggers-club.de>

PR #324, opened 2019-07-20
improve dev builds

therealprof:
This PR basically reverts #141 and plays with inline markers for some large dev build binary size gains

### DIFF
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -54,6 +54,7 @@ pub fn render(
 
         impl #name_pc {
             /// Returns a pointer to the register block
+            #[inline(always)]
             pub const fn ptr() -> *const #base::RegisterBlock {
                 #address as *const _
             }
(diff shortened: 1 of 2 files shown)


burrbull:
@ryankurte can you review this PR?

burrbull:
failed with
```
error: literal out of range for i32
```
on several targets.

src/generate/peripheral.rs
```diff
@@ -48,14 +48,15 @@ pub fn render(
...
 
         unsafe impl Send for #name_pc {}
 
         impl #name_pc {
             /// Returns a pointer to the register block
-            pub fn ptr() -> *const #base::RegisterBlock {
-                #address as *const _
+            #[inline(always)]
+            pub const fn ptr() -> *const #base::RegisterBlock {
+                #address as _
```
burrbull: incorrect cast here (`*i32` instead of `*<width>`)

therealprof: That is very weird. Why could a elided type result in an incompatible pointer?

therealprof: Most likely the issue is not this change but the `PhantomMarker` change above. Anyway let's back that out.

therealprof:
@ryankurte I removed the problematic change. Please have another look.

PR #326, opened 2019-07-21
*const usize cast

burrbull:
r? @therealprof 

Test before merge, because I'm not sure is this right.


src/generate/peripheral.rs
```diff
@@ -56,7 +56,7 @@ pub fn render(
             /// Returns a pointer to the register block
             #[inline(always)]
             pub const fn ptr() -> *const #base::RegisterBlock {
-                #address as _
+                #address as *const usize as _
```
therealprof: I just changed that, why are you changing it back?

PR #328, opened 2019-07-21
Prevent quoted doc strings from generating extra blanks at the beginning

therealprof:
Using `/// something` in a quote! block generates
`#[doc = r" something"]`
need to use `///something` to get the more appropriate
`#[doc = r"something"]`
instead


therealprof:
@adamgreig Try again please.

PR #329, opened 2019-07-23
Generic Writer/Reader

burrbull:


therealprof:
It works fine and the file size is vastly reduced though the compile time stays almost the same (cargo check on the pac itself). I haven't quite figured out why but the dev binary size is also larger; maybe a missed inline somewhere. But to be honest, I'm no fan of all that macro_rule macros, like not at all.

If we want to go "all macro" I'd crate a procedural macro crate and have either `derive` or `attribute` macros for everything; basically just rewrite the SVD file into fully annotated structures and a bit of fairy dust.

burrbull:
It works not fine.
I can't to set different offsets for derived fields as they have same type.
Looks like I need const generics to solve the problem.

burrbull:
Close in favour of #331 .

Issue #115, opened 2017-06-11
Very repetitious output when deriving enumeratedValues


roysmeding:
As I've said elsewhere, I'm working on adding `enumeratedValues` to the SVD for the MCU I'm using. Since there are a lot of fields that use the same enumerations, I'm reusing them using e.g. `<enumeratedValues derivedFrom="GPIOA.MODER.MODER15.MODE" />` in each field.

However, when inspecting the output, I noticed this results in a lot of duplicated code. Most of it seems to be due to the separate write proxies for each field, with each proxy defining all the variant methods, despite their implementation being the exact same each time.

This isn't a huge problem, and I understand why the separate write proxy objects are necessary (each field has a different offset), but it seems like there should be a less boilerplate-y way to accomplish the same thing? The first thing that comes to mind is generics with numerical arguments, but those aren't a part of Rust yet, unfortunately. I'm fairly new to Rust, though, so maybe someone has a better idea?

Some of this would be ameliorated by #44 but that wouldn't stop the duplication entirely.

japaric:
> The first thing that comes to mind is generics with numerical arguments

I don't think generics are required. The field offset could be stored as a field in the proxy object:

``` rust
struct CommonFieldW<'a> {
    w: &'a mut W,
    offset: u8,
}

// instead of FieldAW use CommonFieldW { w, offset: 0 }
// instead of FieldBW use CommonFieldW { w, offset: 1 }
// instead of FieldCW use CommonFieldW { w, offset: 2 }
// ...
```

burrbull:
Closing as implemented. Open new issue if you find it.

PR #331, opened 2019-07-24
Generics Part 1

burrbull:
r? @therealprof 

I decided to split my changes on 3 steps:

1 - this one adds `Reg` Wrapper for register struct with read, write and other methods (instead of traits we discussed) and replace register's `R` and `W` on generic versions.

2 - replace field reader with generic `R` 

3 - replace write proxy with generic `WProxy` ~~(require const generics in rust)~~ (see #115 )

Example: https://gist.github.com/24a0d1a2e13bb464a7c579a07902a3c6

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -17,6 +17,7 @@ pub fn render(
     let access = util::access_of(register);
     let name = util::name_of(register);
     let name_pc = Ident::from(&*name.to_sanitized_upper_case());
+    let _name_pc = Ident::from(format!("_{}", &*name.to_sanitized_upper_case()));
     let name_sc = Ident::from(&*name.to_sanitized_snake_case());
     let rsize = register
         .size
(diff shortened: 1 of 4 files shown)


src/generate/register.rs
```diff
@@ -32,56 +33,26 @@ pub fn render(
...
-            pub fn bits(&self) -> #rty {
-                self.bits
-            }
+            #[doc = #desc]
+            pub type R = crate::R<#rty, super::#_name_pc>;
         });
     }
 
     if can_write {
+<<<<<<< HEAD
```
therealprof: Uhm, conflict markers?

burrbull: It happens.
Fixed.

therealprof: Sure, just makes me wonder how you actually tested this... 🤔

burrbull: I made merge after PR and tests.

burrbull:
Rebased on master. Added `Safe`/`Unsafe` marker structs.

This part should not break HALs, but I'm not sure.

burrbull:
Close in favour of #339

PR #333, opened 2019-07-25
Refactor fields

burrbull:
Non-breaking change.

r? @therealprof


therealprof:
@burrbull For the future: Please avoid merging in changes from the master branch (or anywhere else) and rebase your change instead to keep the history clean.

PR #339, opened 2019-07-26
Generics

burrbull:
r? @therealprof 
cc @adamgreig 

I did it. Almost.

But still have problems with naming.

I want to name field enum as `FIELD`. But in some cases this cause troubles.

Now such system:
- `_R` and `_W` for registers readers/writers (better `R` and `W` as it was before, but conflicts with field `W`)
- `FIELD` for field enums/structs.
- `FIELD_R` - field reader (was `FIELDR` before): now conflicts with `CLK_EN_R` field on Fujitsu and Spansion targets.
- `FIELD_W` - field writer (was `_FIELDW` + `FIELDW` if have enumerated values)

### DIFF
--- a/src/generate.rs
+++ b/src/generate.rs
@@ -1,4 +1,5 @@
 pub mod device;
+pub mod generic;
 pub mod interrupt;
 pub mod peripheral;
 pub mod register;
--- a/src/generate/device.rs
+++ b/src/generate/device.rs
@@ -6,7 +6,7 @@ use crate::errors::*;
 use crate::util::{self, ToSanitizedUpperCase};
 use crate::Target;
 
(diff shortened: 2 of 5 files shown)


therealprof:
@burrbull You're too quick for me. Why did you close this? 😅

burrbull:
I found one bug in write proxy (tha. They shall not be derived (`type FIELD_W<'a> = BASEFIELD_W<'a>`) if base enum is out of register.
It should be new `type FIELD_W<'a> = WProxy<'a, u32, REGNAME, bool, BASEFIELD, crate::Safe>`.
Otherwise it will create incorrect register writers.

I thought I fix all main issues and create this common PR.
I can reopen #331 until find solution.

Also, there is a problem with names.

PR #340, opened 2019-07-27
offset-in-proxy

burrbull:
@therealprof 

Can you tests this change. It adds offset into proxy structure according #115 and can reduce code size in some cases.
It's interesting how debug builds react on this.

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -535,6 +535,9 @@ pub fn fields(
             let mut unsafety = unsafety(f.write_constraint, f.width);
             let width = f.width;
 
+            let _pc_w = &f._pc_w;
+            let mut derived = false;
+
             if let Some((evs, base)) = lookup_filter(&lookup_results, Usage::Write) {
                 let variants = Variant::from_enumerated_values(evs)?;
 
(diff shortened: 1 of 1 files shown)


burrbull:
r? @therealprof 

изображение

therealprof:
Sure, I'll give it a shot.

therealprof:
Increases code size on dev builds usually by around 10 percent, no change on release builds.

burrbull:
So we have to wait const generics for third part from https://github.com/rust-embedded/svd2rust/pull/331#issue-300628516

therealprof:
Have you tried factoring the bits function into a generic bits function, à la:
```
fn bits_u32 (self, value: u32, field: u32, mask: u32, offset: usize) -> u32 {
    value &= !(mask << offset);
    value |= ((field as u32) & mask) << offset;
    value
}
```
and calling that instead?

burrbull:
I don't see how this can help.

therealprof:
Well, we're trying to find a solution for two different problems:

1. For dev builds we need to have the least amount of code because it won't be optimised away under any circumstance; huge amounts of could and also repetitive code are very bad here
2. For release builds we're interested in having code which the compiler can properly optimise; repetitive code doesn't matter but if we get rid of it it must be done in a way that the compiler can still see how to get rid of it again

burrbull:
I agree with the goals.  It is easy to implement generic code for readers as they don't have feedback with registers, but it’s harder for writers.
Field writers produce register writers and have different offsets.

Issue #299, opened 2019-05-23
Incorrect handling of cluster `size` field


burrbull:
size | Define the default bit-width of any device register (implicit inheritance).

It's size of each register in cluster, not a size of whole cluster.

https://github.com/riscv-rust/k210-pac/pull/24

Emilgardis:
I think this is where we wrongly use `cluster_size` for determining sequentiality. https://github.com/rust-embedded/svd2rust/blob/20d8bd4c7554c5b9ced6c1c6941a81747274e9e4/src/generate/peripheral.rs#L614-L644

I'm not sure how this has not tripped anything earlier, thanks for reporting it 👍 
The fix should be easy, I'll see what I can do. Interested in how many svds this is effected by.

burrbull:
I think first of all you should ignore `size` field here https://github.com/rust-embedded/svd2rust/blob/20d8bd4c7554c5b9ced6c1c6941a81747274e9e4/src/generate/peripheral.rs#L601 and always use `cluster_size_in_bits`:
`let cluster_size = cluster_size_in_bits(cluster, defs).ok_or_else(...);`
This should fix most of broken SVDs.

In my view `size` should be argument of `parse` and be passed down the chain from device to register.
But it will require changes in `svd-parser`.

PR #341, opened 2019-07-27
sizes, 64bit

burrbull:
r? @therealprof 

Right use of cluster.size + better support of 64bit registers.

See #299 , partially closed in #325 .

Not tested yet.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -216,10 +216,14 @@ pub fn access_of(register: &Register) -> Access {
 }
 
 /// Turns `n` into an unsuffixed separated hex token
-pub fn hex(n: u32) -> Tokens {
+pub fn hex(n: u64) -> Tokens {
     let mut t = Tokens::new();
-    let (h2, h1) = ((n >> 16) & 0xffff, n & 0xffff);
-    t.append(if h2 != 0 {
+    let (h4, h3, h2, h1) = ((n >> 48) & 0xffff, (n >> 32) & 0xffff, (n >> 16) & 0xffff, n & 0xffff);
+    t.append(if h4 != 0 {
+        Ident::from(format!("0x{:04x}_{:04x}_{:04x}_{:04x}", h4, h3, h2, h1))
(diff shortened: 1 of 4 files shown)


burrbull:
изображение

src/util.rs
```diff
@@ -216,10 +216,16 @@ pub fn access_of(register: &Register) -> Access {
...
 
 /// Turns `n` into an unsuffixed separated hex token
-pub fn hex(n: u32) -> Tokens {
+pub fn hex(n: u64) -> Tokens {
     let mut t = Tokens::new();
-    let (h2, h1) = ((n >> 16) & 0xffff, n & 0xffff);
-    t.append(if h2 != 0 {
+    let (h4, h3, h2, h1) = ((n >> 48) & 0xffff, (n >> 32) & 0xffff, (n >> 16) & 0xffff, n & 0xffff);
+    t.append(if n == 1 {
+        Ident::from("1")
```
therealprof: This special case is weird...

therealprof:
I was kind of hoping that at some point we could get rid of the custom formatting hacks but at the moment there still seems to be no way to generate lint-less long literals with the built-in formatting mechanics...

burrbull:
Don't merge. Something wrong.

src/util.rs
burrbull: What are you offering?

therealprof: Getting rid of it? I'd fully expect this function to always return a real hex literal, not a hex literal or "1".

burrbull: Or you about "1"? I can delete this.

therealprof: Yes, I don't like a function called `hex` to treat the number 1 special and not turn it into hex. ;)

burrbull:
Fixed.

therealprof:
Looks great. Can you pleasse add a CHANGELOG entry, too? That way we don't have to hunt down all changes after the fact. ;)

burrbull:
Done.

PR #342, opened 2019-07-27
Don't use quote concats to munge the Vec<Tokens> into one huge string

therealprof:
By manually iterating over the Vec<Token> when writing out to lib.rs we
can insert new lines after each token. In addition we split up some
of the inner quote repetition operator uses into regular iteration, too,
to generate individual Tokens. This changes a typical lib.rs file from 2
to 20k lines (200k after rustfmt) which should help a lot to prevent
tools (like rust when errors happen) from going insane and makes the
files a lot more directly usable without having to rustfmt all the time.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -54,7 +54,7 @@ fn run() -> Result<()> {
                     env_logger::DEFAULT_FILTER_ENV
                 ))
                 .takes_value(true)
-                .possible_values(&["off", "error", "warn", "info", "debug", "trace"])
+                .possible_values(&["off", "error", "warn", "info", "debug", "trace"]),
         )
         .version(concat!(
             env!("CARGO_PKG_VERSION"),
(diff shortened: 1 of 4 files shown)


therealprof:
r? adamgreig

src/main.rs
```diff
@@ -54,7 +54,7 @@ fn run() -> Result<()> {
                     env_logger::DEFAULT_FILTER_ENV
                 ))
                 .takes_value(true)
-                .possible_values(&["off", "error", "warn", "info", "debug", "trace"])
+                .possible_values(&["off", "error", "warn", "info", "debug", "trace"]),
```
adamgreig: What's this comma doing?

adamgreig:
I think this looks good as it is and it's already a big improvement. Do you think it's worth looking for other similar improvements to the output format now, or would you rather merge this PR and go from there?

src/main.rs
therealprof: Must me coming from `rustfmt`; I thought I stripped all of the formatting changes out but I must have missed that. 🤷🏻‍♂️

therealprof:
I'd keep it like that for now. We're going to need to rewrite the code generation somewhat anyway. The current approach is not really compatible with future versions of `quote`...

PR #344, opened 2019-07-28
Join the bits muching into a single line to improve dev builds

therealprof:
Also this rectifies many of the tons of
https://rust-lang.github.io/rust-clippy/master/index.html#identity_op
lints we're getting.

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -323,7 +323,7 @@ pub fn fields(
 
         let bits = &f.bits;
         let mask = &f.mask;
-        let offset = &f.offset;
+        let offset: usize = f.offset.parse().unwrap();
         let fty = &f.ty;
 
         let lookup_results = lookup(
@@ -341,8 +341,15 @@ pub fn fields(
             } else {
                 quote! { as #fty }
             };
-            let value = quote! {
-                ((self.bits >> #offset) & #mask) #cast
+            let value = if offset != 0 {
+                let offset = &f.offset;
+                quote! {
+                    ((self.bits >> #offset) & #mask) #cast
+                }
+            } else {
+                quote! {
+                    (self.bits & #mask) #cast
+                }
             };
 
             if let Some((evs, base)) = lookup_filter(&lookup_results, Usage::Read) {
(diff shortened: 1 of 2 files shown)


burrbull:
Interesting solution.

therealprof:
I'm happy to report that this change makes it possible for the first time to build one of my examples (for a STM32F042) in dev mode:
```
  32646	      0	      4	  32650	   7f8a	target/thumbv6m-none-eabi/debug/examples/i2c_hal_ina260reader
```
😅

src/generate/register.rs
```diff
@@ -720,13 +727,24 @@ pub fn fields(
...
+                let offset = &f.offset;
+                quote! {
+                    ///Writes raw bits to the field
+                    #[inline(always)]
+                    pub #unsafety fn #bits(self, value: #fty) -> &'a mut W {
+                        self.w.bits = self.w.bits & !(#mask << #offset) | ((value as #rty) & #mask) << #offset;
+                        self.w
+                    }
+                }
+            } else {
```
burrbull:
```rust
} else if f.width == rsize {
 ...
     self.w.bits = value
     self.w
} else {
```
for `0xffff_ffff` etc.
`rsize` you can pass with `fields()` instead of `rty`

therealprof: I know. 😅Small steps please so our fellow reviewers/approvers can follow what's going on.

src/generate/register.rs
```diff
@@ -720,13 +727,24 @@ pub fn fields(
...
-                    self.w.bits &= !(#mask << #offset);
-                    self.w.bits |= ((value as #rty) & #mask) << #offset;
-                    self.w
+            proxy_items.push(if offset != 0 {
+                let offset = &f.offset;
+                quote! {
+                    ///Writes raw bits to the field
+                    #[inline(always)]
+                    pub #unsafety fn #bits(self, value: #fty) -> &'a mut W {
+                        self.w.bits = self.w.bits & !(#mask << #offset) | ((value as #rty) & #mask) << #offset;
```
adamgreig: What do you think about adding some extra parenthesis to make it clearer what the operator precedence is? This is basically `x = x & !y | z << w` which is something of an exercise to parse in your head...

therealprof: Uhm, I could do but would it actually make a difference? And where would you like them? ;)

adamgreig: The current logic is correct so I suggest adding them to maintain correctness :p 

```rust
self.w.bits = (self.w.bits & !(#mask << #offset)) | (((value as #rty) & #mask) << #offset);
```

therealprof: Done.

PR #345, opened 2019-07-28
Cut down excessive test list of "Freescale" targets and add new ones

therealprof:
Freescale is constantly holding up CI (which is bad) and every now and
then even bumping into time limits. Worst of all, pretty much all of the
SVDs are outdated since that devision of Freescale was acquired by NXP a
long time ago. So let's cut down the outdated list of test targets by a
few members of each family and activate some NXP test targets instead
which should work now, but didn't before.

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -4,7 +4,7 @@ test_svd() {
     (
         cd $td &&
             curl -LO \
-                 https://raw.githubusercontent.com/posborne/cmsis-svd/python-0.4/data/$VENDOR/${1}.svd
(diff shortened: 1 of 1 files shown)


therealprof:
Let's see how this works...

bors try

therealprof:
bors try

therealprof:
bors try

therealprof:
Oh god, the pulled SVD files are from a ghost branch "python-0.4"...

therealprof:
bors try

therealprof:
So I also changed from the ghost branch to master and adjusted the STMicro tests accordingly. Can I get a yay?

PR #347, opened 2019-07-29
Generic Part1

burrbull:
Rebased and reopened.

### DIFF
--- a/src/generate/generic.rs
+++ b/src/generate/generic.rs
@@ -10,6 +10,27 @@ pub fn render() -> Result<Vec<Tokens>> {
     generic_items.push(quote! {
         use core::marker;
 
+        ///This trait shows that register has `read` method
+        ///
+        ///Registers marked with `Writable` can be also `modify`'ed
+        pub trait Readable {}
+
+        ///This trait shows that register has `write`, `write_with_zero` and `reset` method
+        ///
+        ///Registers marked with `Readable` can be also `modify`'ed
+        pub trait Writable {}
+
+        ///Reset value of the register
+        ///
+        ///This value is initial value for `write` method.
+        ///It can be also directly writed to register by `reset` method.
+        pub trait ResetValue {
+            ///Register size
+            type Type;
+            ///Reset value of the register
+            fn reset_value() -> Self::Type;
+        }
+
         ///Converting enumerated values to bits
         pub trait ToBits<N> {
             ///Conversion method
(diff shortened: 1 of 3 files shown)


burrbull:
cc @therealprof 

With this PR debug builds grow ~1.5% (I think it is because `Deref`), but source now much cleaner.

It also depends on #358.

burrbull:
cc @Disasm

burrbull:
@therealprof  I've rebased this on master. Test, please.

src/generate/generic.rs
```diff
@@ -18,23 +30,97 @@ pub fn render() -> Result<Vec<Tokens>> {
...
+                self.register.set(f(&mut W {bits: Self::reset_value(), _reg: marker::PhantomData}).bits);
+            }
+        }
+    });
+
+    generic_items.push(quote! {
+        impl<U, REG> Reg<U, REG>
+        where
+            Self: Writable,
+            U: Copy + Default
```
Disasm: Is this `Default` equals to "returns zero"?

Disasm:
`Send` and `!Sync` were implemented on registers before, with this change they are not implemented.

src/generate/generic.rs
burrbull: Yes, it is for `bool`, `u8`, `u16` `u32` and `u64`.

burrbull:
> `Send` and `!Sync` were implemented on registers before, with this change they are not implemented.

It was auto-implementation?

src/generate/generic.rs
Disasm: Yes, but these types are not enforced I guess

Disasm:
Yes, I think that was due to the inner `VolatileCell`: it's also `Send` and `!Sync`.

Disasm:
As discussed on Matrix, documentation becomes even more difficult to understand with this change: it's unclear how to read and write registers, also there is no information about field variants.

therealprof:
TBH I'm not too sure whether we really need that `VolatileCell` anyway once we've remodelled everything. 😅

Disasm:
@therealprof Yes, but it's better to have `Send` and `!Sync` anyway, because they add important properties.

burrbull:
error[E0658]: negative trait bounds are not yet fully implemented; use marker types for now

What to do with this?

Disasm:
I think that PhantomData with some weird type can help

burrbull:
Can you write how you see implementation for `!Sync`?

Disasm:
Seems like `!Sync` is already implemented for `Reg`, but as for `Send` it's difficult to say whether it's implemented or not for a specific register.

burrbull:
I've added `Send` and slightly improved docs.

src/generate/generic.rs
burrbull: They are enforced by size type of register or field.

burrbull:
What you mean `there is no information about field variants`?

Disasm:
@burrbull It's no longer accessible from the register documentation pages. Only on upper levels.

src/generate/generic.rs
Disasm: But it's possible to create a wrapper with the same size and a custom Default impl that returns reset value, for example

burrbull: How? All fields are private. You can get register only from peripheral as it was before.

burrbull:
> @burrbull It's no longer accessible from the register documentation pages. Only on upper levels.

Can you show example?

Disasm:
[spi0::RegisterBlock](https://docs.rs/k210-pac/0.1.0/k210_pac/spi0/struct.RegisterBlock.html) -> [CTRLR0](https://docs.rs/k210-pac/0.1.0/k210_pac/spi0/struct.CTRLR0.html) -> [R](https://docs.rs/k210-pac/0.1.0/k210_pac/spi0/ctrlr0/struct.R.html) -> [WORK_MODER](https://docs.rs/k210-pac/0.1.0/k210_pac/spi0/ctrlr0/enum.WORK_MODER.html)

burrbull:
Here is one breaking change (#354). Reader now have `variant` method:
изображение

same as writer:

изображение

And now you can use this enum for both read and write:
```rust
let v = spi.ctrl0.read().work_mode().variant();
...
spi.ctrl0.write(|w| w.work_mode().variant(v)); // Before now incompatible types error was here.
```

burrbull:
@therealprof I can't find where is source of this documentation https://docs.rs/svd2rust ?

src/generate/generic.rs
Disasm: Hmm, now I see. Thanks.

therealprof:
@burrbull You mean `src/lib.rs`?

src/generate/generic.rs
```diff
@@ -10,6 +10,27 @@ pub fn render() -> Result<Vec<Tokens>> {
...
+
+        ///Reset value of the register
+        ///
+        ///This value is initial value for `write` method.
+        ///It can be also directly writed to register by `reset` method.
+        pub trait ResetValue {
+            ///Register size
+            type Size;
+            ///Reset value of the register
+            fn reset_value() -> Self::Size;
```
Disasm: This associated type name is debatable, but I'm not going to force my point of view here.

burrbull: I've changed Size on Type

src/generate/register.rs
```diff
@@ -210,12 +142,38 @@ pub fn render(
...
+        pub type #name_pc = crate::Reg<#rty, #_name_pc>;
+
+        #[allow(missing_docs)]
+        #[doc(hidden)]
+        pub struct #_name_pc;
+    });
 
+    let doc = format!("`read()` method returns [{0}::R]({0}::R) reader structure", &name_sc);
+    if can_read {
+        out.push(quote! {
```
therealprof: Why not move the `let doc = ...` inside the `if` clause? Should be cheaper to compute and avoids the risk of accidentally using it in a wrong context.

src/generate/register.rs
```diff
@@ -210,12 +142,38 @@ pub fn render(
...
+            #[doc = #doc]
+            impl crate::Readable for #name_pc {}
+        });
+    }
+    let doc = format!("`write(|w| ..)` method takes [{0}::W]({0}::W) writer structure", &name_sc);
+    if can_write {
```
therealprof: Same here.

therealprof:
Also please rebase and make sure the commits have a meaningful commit message.

burrbull:
Rebased.

therealprof:
bors try

therealprof:
@Disasm Still happy?

Disasm:
@therealprof yes, looks good.

Disasm:
In general, I'd improve documentation for generic methods (read/write/...), replacing references to the book with short examples.

Issue #353, opened 2019-07-29
Bad news


burrbull:
cc @therealprof 

I think I made regression with `FR` field.
Now there is no access to field enums to compare they manually. It's my mistake. Big mistake.
It doesn't break most of code, but it's regression.
I think we should revert #343.

Alternative is to add `values` method to reader.

therealprof:
Well, the release is out. But I'm not even sure that was intentional that one could do that. Do you see any usecase for that?

PR #354, opened 2019-07-30
add variant to reader

burrbull:
r? @therealprof 

Restore access to `enum`'s by `variant` method that return `enum FIELD` when possible or `Variant`.

Also replace `_to_bits` method in writer with trait.
So now enums for reader and writer are identical (when they don't use different values for read and write) and can be replaced by one.

See #353 .

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -277,7 +277,7 @@ pub fn fields(
             let pc_r = Ident::from(&*format!("{}R", pc));
             let _pc_r = Ident::from(&*format!("{}_R", pc));
             let pc_w = Ident::from(&*format!("{}W", pc));
-            let _pc_w = Ident::from(&*format!("_{}W", pc));
+            let _pc_w = Ident::from(&*format!("{}_W", pc));
             let _sc = Ident::from(&*format!("_{}", sc));
             let bits = if width == 1 {
                 Ident::from("bit")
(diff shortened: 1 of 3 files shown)


burrbull:
изображение

src/generate/register.rs
```diff
@@ -277,7 +277,7 @@ pub fn fields(
             let pc_r = Ident::from(&*format!("{}R", pc));
             let _pc_r = Ident::from(&*format!("{}_R", pc));
             let pc_w = Ident::from(&*format!("{}W", pc));
-            let _pc_w = Ident::from(&*format!("_{}W", pc));
+            let _pc_w = Ident::from(&*format!("{}_W", pc));
```
therealprof: Ugh... It's a pity we didn't notice this earlier. But would you please remove it from this PR?

therealprof:
Which file did you test it on? Don't seem to have a SVD file in reach which actually uses the changed parts.

src/generate/register.rs
burrbull: Ok.
Then
`_pc_r  => "_{}R"` ?

therealprof: No, just remove the change from this PR and put it in a separate one.

burrbull:
from `stm32-rs`
For example LATENCY field: https://docs.rs/stm32f1/0.8.0/stm32f1/stm32f103/flash/acr/index.html

burrbull:
I've moved identical code to `add_from_variants` and `derive_from_base` functions.

therealprof:
Hm, it's a pity I can't see the changes using the unmodified SVD files from `cmsis-svd`. Anyway, please add a CHANGELOG entry and we can land this.

burrbull:
Done.

PR #359, opened 2019-07-31
Test RISC-V SVDs on stable, add FU540 SVD

Disasm:


therealprof:
cc #358

therealprof:
Let's see whether bors agrees to merge #358. If it does we'll stage this one right behind to not break the build.

Issue #360, opened 2019-07-31
Generic mod in separate file


burrbull:
I think it makes sense to have option to separate `generic` module in different file when it's needed like `build.rs` and `device.x`.

PR #362, opened 2019-08-02
separate generic mod

burrbull:
Adds option `-g` which when enabled store contents of `generic.rs` in separate file.

It's useful for common crates like `stm32f1` where one crate for several `pac`s.

Uses `syn::parse` instead of vector of `quote!`.


burrbull:
r? @therealprof 

#360

PR #369, opened 2019-08-09
Fix crash on missing register description in SVD file

aurabindo:
SVDs available for DA1469x from https://www.keil.com/dd2/pack have few registers with missing `<description>` and this causes svd2rust to crash.

This PR enables to continue with a dummy description  and print name of the
problematic register instead of just bailing out.

### DIFF
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -679,7 +679,7 @@ fn expand_register(
     match register {
         Register::Single(info) => register_expanded.push(RegisterBlockField {
             field: convert_svd_register(register, name),
-            description: info.description.clone().unwrap(),
+            description: info.description.clone().unwrap_or_else(|| "".to_string()),
             offset: info.address_offset,
             size: register_size,
         }),
@@ -699,7 +699,7 @@ fn expand_register(
             if array_convertible {
                 register_expanded.push(RegisterBlockField {
                     field: convert_svd_register(&register, name),
-                    description: info.description.clone().unwrap(),
+                    description: info.description.clone().unwrap_or_else(|| "".to_string()),
                     offset: info.address_offset,
                     size: register_size * array_info.dim,
                 });
@@ -707,7 +707,7 @@ fn expand_register(
                 for (field_num, field) in expand_svd_register(register, name).iter().enumerate() {
                     register_expanded.push(RegisterBlockField {
                         field: field.clone(),
-                        description: info.description.clone().unwrap(),
+                        description: info.description.clone().unwrap_or_else(|| "".to_string()),
                         offset: info.address_offset + field_num as u32 * array_info.dim_increment,
                         size: register_size,
                     });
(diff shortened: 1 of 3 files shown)


src/generate/peripheral.rs
```diff
@@ -679,7 +679,10 @@ fn expand_register(
     match register {
         Register::Single(info) => register_expanded.push(RegisterBlockField {
             field: convert_svd_register(register, name),
-            description: info.description.clone().unwrap(),
+            description: info
+                .description
+                .clone()
+                .unwrap_or("No description available".to_string()),
```
Disasm: Maybe it's better to use `unwrap_or_else` in such cases to not create a `String` object every time.

therealprof:
I haven't looked in detail but maybe we could add the no description text earlier than when generating the output. Oh, and would be great if the no-description description was a bit shorter; no need to be verbose here...

Disasm:
@therealprof That `description` field isn't mandatory in SVD spec, so it's ok to get `None` from `svd` parser. What do you mean by "earlier" then?

therealprof:
@Disasm It's a bit unfortunate that we're using the parsed SVD, patch it up and turn it into code all in the same place. Instead it'd be great if we had a pass which used the parsed SVD, checks whether it makes sense, cleans up the odds bits and fills in missing bits so we finally just dump the consistent information into code.

aurabindo:
@therealprof  how about an empty String (`""`) or do you prefer `No description`  ?

src/generate/peripheral.rs
Disasm: There are two functions with pretty the same purpose: `unwrap_or` and `unwrap_or_else`. `unwrap_or` takes a value (and you need to construct it) every time, `unwrap_or_else` constructs this value only if needed by calling a function.

therealprof: The `"No description available".to_string()` will always be executed independently of whether `info.description` is `None` or `Some(...)`. If the description is `None` then the result of the operation will be dropped so you're constructing the same string over and over again just to drop it a moment later.

aurabindo: Oh I see. I didnt know that. Thanks for the pointer. I'll make the change

therealprof: There's a clippy lint about that: https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call

aurabindo:
Well that was my version of **_Slight_** hint ;) I'll just put "Dummy description" ?

therealprof:
Well, descriptions are optional so "Bad SVD" would be technically wrong. It should be something which hints to the "problem" at hand, so "Nothing" or just "No description" would work for me. While you are at it, please add a line to the CHANGELOG and rebase the commit.

therealprof:
Hm, coming to thing of it an empty String would also be acceptable (if Rust accepts that).

PR #371, opened 2019-08-10
register Defaults -> svd-parser

burrbull:
r? @therealprof 

Depends on https://github.com/japaric/svd/pull/73

I decided it is more appropriate to move the forwarding of `Defaults` (common `size`, `access`, etc.) to the parsing stage.

This simplify handling of parsed device.
Also I'm sure this fix a lot of bugs. (for example https://github.com/riscv-rust/k210-pac/blob/c70d429c45e25f15a5bddcd6c25d4470eab66842/k210.svd#L1327)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ error-chain = "0.11.0"
 inflections = "1.1.0"
 log = { version = "~0.4", features = ["std"] }
 quote = "0.3.15"
-svd-parser = "0.7"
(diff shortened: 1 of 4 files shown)


Disasm:
I don't think this `svd` PR will be approved: parsers shouldn't process semantic information in general. Maybe it's time to create the layer mentioned here: https://github.com/rust-embedded/svd2rust/pull/369#issuecomment-519923151

burrbull:
I can split both these PR on 2 parts.

First will only fix inconsistency with SVD specification.

PR #375, opened 2019-08-14
Change 'use crate::' to 'use::super' in generated code

aurabindo:
This fixes a regression starting from f8e7692a3ed98eca5a1a9100f1

Some peripheral access crates like those maintained by stm32-rs have
multiple PACs within same crate. For example, stm32f1 has PACs for
stm32f101, stm32f102, etc.

Using the pattern 'use crate::' in generated code should be avoided
since during compliation, the assumptions are invalidated. When
multiple PACs are merged into one crate, there is one more level to
navigate so as to get the actual symbols meant with `use crate`.

As a solution, instead of resolving top down, do bottom up approach
with 'use super::'

Signed-off-by: Aurabindo Jayamohanan <mail@aurabindo.in>

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -54,7 +54,7 @@ pub fn render(
         let desc = format!("Reader of register {}", register.name);
         mod_items.push(quote! {
             #[doc = #desc]
-            pub type R = crate::R<#rty, super::#name_pc>;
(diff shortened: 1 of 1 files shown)


aurabindo:
Closing since there is a better solution in https://github.com/stm32-rs/stm32-rs/pull/271

PR #379, opened 2019-08-15
enum descriptions

burrbull:
r? @therealprof 
cc @Disasm 

Before:
изображение

After:
after1


burrbull:
Done.

Disasm:
I wonder if this "Bit 0" in the description is not confusing here. The same enum can be used for multiple locations with different bit offsets.

burrbull:
You are right. I'll try to fix it.

PR #364, opened 2019-08-03
Newsyn

burrbull:
r? @therealprof 

New `syn`, `quote` and `proc_macro2`.

PR #383, opened 2019-08-24
Option-like API for variant

burrbull:
r? @therealprof 
cc @Disasm

### DIFF
--- a/src/generate/generic.rs
+++ b/src/generate/generic.rs
@@ -221,15 +221,6 @@ impl<REG: RegisterSpec> W<REG> {
     }
 }
 
-/// Used if enumerated values cover not the whole range.
-#[derive(Clone, Copy, PartialEq)]
-pub enum Variant<U, T> {
-    /// Expected variant.
-    Val(T),
-    /// Raw bits.
-    Res(U),
-}
-
 /// Field reader.
 ///
 /// Result of the `read` methods of fields.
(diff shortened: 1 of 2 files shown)


therealprof:
Interesting, what's the intended use for that?

src/generate/generic.rs
```diff
@@ -227,3 +227,63 @@ pub enum Variant<U, T> {
...
+
+impl<U, T> Variant<U, T> where U: core::fmt::Debug {
+    /// Moves the value `v` out of the `Variant` if it is `Val(v)`.
+    ///
+    /// Panics if the self value equals `Res`
+    #[inline]
+    pub fn unwrap(self) -> T {
+        match self {
+            Val(v) => v,
+            Res(u) => panic!("Unexpected variant: {:?}", u),
```
therealprof: I'm not to hot about this `panic!` and the `expect` method below. If this ends up in use in real code it'd add significant bloat.

burrbull:
Sometimes enumeratedValues does not cover all variants of bits (number of variants less then 2^field_width).
In these cases we have to return `Variant<_,T>` with reserved bits field instead of `T` (~10% of enums).
изображение
It is rare used feature. Usually we simply call `is_something()`.

But if you want to get variant as `enum`, you need to call `variant` and unwrap it.
Getting reserved bits usually means something go wrong.

src/generate/generic.rs
burrbull: I can remove `unwrap` and keep `expect` as it is more explicit.

burrbull:
And what with #364 ?

src/generate/generic.rs
therealprof: I'm fine with with those methods per se but let's get rid of the formatting and the `Debug` bound, please.

burrbull: Ok

burrbull: Is now better?

src/generate/generic.rs
```diff
@@ -227,3 +227,62 @@ pub enum Variant<U, T> {
     ///Raw bits
     Res(U),
 }
+
+use Variant::*;
+impl<U, T> Variant<U, T> {
+    /// Check if the variant is expected
+    pub fn is_value(&self) -> bool {
```
Disasm: This naming is strange. At least function name does not correspond to the docs. I can't suggest anything nicer, though.

src/generate/generic.rs
```diff
@@ -227,3 +227,62 @@ pub enum Variant<U, T> {
...
+            Val(v) => v,
+            Res(u) => f(u),
+        }
+    }
+
+    /// Unwraps a result, yielding the content of an `Val`.
+    ///
+    /// Panics if the value is an `Res`, with a panic message including the
+    /// passed message, and the content of the `Res`.
+    pub fn expect(self, msg: &'static str) -> T {
```
Disasm: Do you really need `'static` bound here?

src/generate/generic.rs
burrbull: Do you have suggestions?

src/generate/generic.rs
burrbull: It doesn’t compile without `'static`.

therealprof: Weird. The original expect doesn't have it.

burrbull: I think it's because original `expect` use formatted `panic!("{}", msg)`.
With 2 arguments.

https://doc.rust-lang.org/src/core/macros.rs.html#12

therealprof:
@Disasm Are you happy with the current state?
@burrbull Can you fix the conflict, please?

burrbull:
Rebased.

burrbull:
See #496

Issue #287, opened 2019-03-27
Register access is not inlined during compile


agalakhov:
Hi,

I recently found out that the code generated by `svd2rust` (the one of `stm32f1` crate) is very slow if compiled for STM32F103 as --release (about five times slower than expected). Disassembling the resulting code has shown that:
* All `Deref` implementations for all registers are called as functions and not inlined.
* All `.ptr()` methods of all registers are called as functions and not inlined.

Adding `#[inline(always)]` manually to these methods solved the issue (I got correct timings and well-optimized assembly code).

Perhaps `svd2rust` should generate `#[inline]` or even `#[inline(always)]` attribute for such calls, as otherwise the compiler sometimes fails to optimize here.

Regards,
Alexey

therealprof:
That is very odd. Can you provide some assembly for comparison?

agalakhov:
Of course. Relevant part of the code being compiled:
```
let mut led = gpioc.pc13.into_push_pull_output(&mut gpioc.crh);
loop {
    led.set_high();
    led.set_low();
}
```
Compiling with rustc 1.33 (stable) for thumbv7m-none-eabi with --release option and linking with lld. The following assembler code is generated:
```
80001f6:       f000 fa1d       bl      8000634 <_ZN7stm32f19stm32f1035GPIOC3ptr17he8d61d57514c9137E>
80001fa:       6105            str     r5, [r0, #16]
80001fc:       f000 fa1a       bl      8000634 <_ZN7stm32f19stm32f1035GPIOC3ptr17he8d61d57514c9137E>
8000200:       6104            str     r4, [r0, #16]
8000202:       e7f8            b.n     80001f6 <main+0xc6>
```
In more complex examples (involving timer delays) I also observed calls to `deref` instead of `ptr`.
The today's nightly version generates the same code.

Adding `#[inline]` to `ptr()` like that:
```
impl GPIOC {
    #[doc = r" Returns a pointer to the register block"]
    #[inline]
     pub fn ptr() -> *const gpioa::RegisterBlock { 
        1073811456 as *const _
    }
}
```
resulted in the following machine code:
```
80001fa:       6020            str     r0, [r4, #0]
80001fc:       6025            str     r5, [r4, #0]
80001fe:       e7fc            b.n     80001fa <main+0xca>
```
which is the expected one. For more complex example I had to add `#[inline]` to `Deref` as well.

[...]

PR #389, opened 2019-10-09
Add `#[inline(always)]` to deref implementations.

m-ou-se:
See #287.


m-ou-se:
Is this worth a new minor release?

PR #372, opened 2019-08-10
Peripheral Cluster defaults

burrbull:
r? @Disasm 

Depends on https://github.com/japaric/svd/pull/74

Instead of #371

PR #394, opened 2019-11-03
use repr(fty) for enums

burrbull:
enums are now in C-style with `#repr(..)`

Should slightly reduce code size in debug mode when enums are used.

r? @Disasm
cc @therealprof

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -636,14 +636,23 @@ impl Variant {
 fn add_from_variants(mod_items: &mut Vec<TokenStream>, variants: &Vec<Variant>, pc: &Ident, f: &F, desc: &str, reset_value: Option<u64>) {
     let fty = &f.ty;
 
+    let mut repr = quote! { #[repr(#fty)] };
+    let mut cast = quote! { variant as _ };
+
+    if f.ty == "bool" {
+        repr = quote! { };
+        cast = quote! { variant as u8 != 0 };
+    }
+
     let vars = variants
         .iter()
         .map(|v| {
             let desc = util::escape_brackets(&format!("{}: {}", v.value, v.doc));
             let pcv = &v.pc;
+            let pcval = &util::unsuffixed(v.value);
             quote! {
                 #[doc = #desc]
-                #pcv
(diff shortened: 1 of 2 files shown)


burrbull:
bors try

burrbull:
As `msp430g2553` contains empty `<enumeratedValues/>`, we should use new `svd-parser` with https://github.com/rust-embedded/svd/pull/69 patch.

burrbull:
bors try

Emilgardis:
Why retry? Is error E0084 not valid?

burrbull:
Because it should be fixed in #372 .

@Emilgardis What do you think about topic? Is it right way?

burrbull:
Change example:
изображение

therealprof:
Uhm, which MCUs are affected by the change? I just tried it on a STM32F042 but there's literally no change.

burrbull:
Rebased. `repr` attribute removed for `bool` fields.
bors try

Emilgardis:
I'm not sure I like this solution. I don't have anything against merging it, but it just seems weird.

burrbull:
> but it just seems weird.

What exactly do you not like? `repr` attributes?

Emilgardis:
My bad, I thought we already did this, misunderstood what this pr does. This is very sensible to do

src/generate/register.rs
```diff
@@ -636,14 +636,23 @@ impl Variant {
 fn add_from_variants(mod_items: &mut Vec<TokenStream>, variants: &Vec<Variant>, pc: &Ident, f: &F, desc: &str, reset_value: Option<u64>) {
     let fty = &f.ty;
 
+    let mut repr = quote! { #[repr(#fty)] };
+    let mut cast = quote! { variant as _ };
+
+    if f.ty == "bool" {
+        repr = quote! { };
+        cast = quote! { variant as u8 != 0 };
+    }
```
Emilgardis: Nitpick: This could be if else instead, have no real preference however.

burrbull: Done

PR #397, opened 2019-11-12
Ignore non snake case types that can exist in SVD files

mvertescher:
Hi! :wave: Thanks for this awesome project!

I've run into a build error generating bindings for the [psoc6-pac](https://github.com/psoc-rs/psoc6-pac). Here's a example of one of the many build failures:

```
error: structure field `data_list_sent_update__status` should have a snake case name
   --> src/ble.rs:187:9
    |
187 |     pub data_list_sent_update__status: self::blell::DATA_LIST_SENT_UPDATE__STATUS,
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: convert the identifier to snake case: `data_list_sent_update_status`
```

This issue is caused by `#![deny(warnings)]` and should be fixed by adding the `#![allow(non_snake_case)]` exception. There may be alternative fixes, I'm open to suggestions!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Enum items now associated with values (C-style), enums annotated with `repr(fty)`
 - Bump `svd-parser` dependency (0.8.1)
+- Switched from denying all warnings to only a subset.
 
 ## [v0.16.1] - 2019-08-17
 
--- a/src/generate/device.rs
+++ b/src/generate/device.rs
@@ -45,9 +45,29 @@ pub fn render(
 
     out.push(quote! {
         #![doc = #doc]
+        // Deny a subset of warnings
+        #![deny(const_err)]
+        #![deny(dead_code)]
+        #![deny(improper_ctypes)]
+        #![deny(legacy_directory_ownership)]
         #![deny(missing_docs)]
-        #![deny(warnings)]
+        #![deny(no_mangle_generic_items)]
+        #![deny(non_shorthand_field_patterns)]
+        #![deny(overflowing_literals)]
+        #![deny(path_statements)]
+        #![deny(patterns_in_fns_without_body)]
+        #![deny(plugin_as_library)]
+        #![deny(private_in_public)]
+        #![deny(safe_extern_statics)]
+        #![deny(unconditional_recursion)]
+        #![deny(unions_with_drop_fields)]
+        #![deny(unused_allocation)]
+        #![deny(unused_comparisons)]
+        #![deny(unused_parens)]
+        #![deny(while_true)]
+        // Explicitly allow a few warnings that may be verbose
         #![allow(non_camel_case_types)]
+        #![allow(non_snake_case)]
         #![no_std]
     });
 


therealprof:
The dreaded `#![deny(warnings)]` strikes again. I'd rather remove that and add explicit `deny` rules for the things we really do not want to allow.

therealprof:
cf. https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md

mvertescher:
>  I'd rather remove that and add explicit deny rules for the things we really do not want to allow.

SGTM, I'll update this PR to instead deny a (hopefully safe) list of warnings.

therealprof:
If you would be so kind to update the CHANGELOG.md and rebase I'd give this a CI run...

mvertescher:
@therealprof I've updated to only deny a subset of warning adopted from the [patterns repo]( https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md) (the warnings referenced there are a little out of date, I'll [fix that up](https://github.com/rust-unofficial/patterns/issues/89) later). 

I _think_ the set of denied warnings should not affect users, but I could be wrong.

mvertescher:
Hrm, taking a closer look at this, there are some rustc lints denied in this change that are explicitly allowed (`rustc -W help`) by the Rust 1.39.0 compiler: 

```
missing_debug_implementations,
trivial_casts,
unused_import_braces,
unused_results,
```

I believe that by denying these, we could potentially break future PAC builds. What do you think @therealprof ?

therealprof:
No idea to be honest. Can you rebase, then we can give it a spin? I'm totally expecting that we will need to tweak those...

mvertescher:
Cool, that sounds good. This branch should be up to date with master. I think we can run CI?

burrbull:
bors try

therealprof:
> Cool, that sounds good. This branch should be up to date with master. I think we can run CI?

It's still separate commits. ;) Maybe I should have said rebase and squash. It looks a weird to merge a PR that does one thing in a commit and reverses course in the next.

mvertescher:
> It's still separate commits. ;) Maybe I should have said rebase and squash. It looks a weird to merge a PR that does one thing in a commit and reverses course in the next.

Ah yes, that makes more sense. :) I'll rebase and squash!

therealprof:
bors try

therealprof:
bors try

therealprof:
I think it got confused by the rebase in the middle of the try run. ;)

therealprof:
As expected tons of errors here:
https://travis-ci.org/rust-embedded/svd2rust/builds/611328824

Not sure whether this is the only one but it's certainly the most common one:
`error: trivial cast: `&uart0::RegisterBlock` as `*const uart0::RegisterBlock`

therealprof:
@mvertescher Can you have a look at the logs and adjust the PR accordingly?

mvertescher:
So, I've removed any deprecated (`bad-style`) or now allowed lints from the denied lint list. I believe   this new configuration should not break any existing crates, but we'll see!

mvertescher:
bors try

mvertescher:
@therealprof can you give CI a run?

therealprof:
bors d+

therealprof:
Just in case something fails.

PR #400, opened 2019-12-14
field array support

burrbull:


burrbull:
bors try

burrbull:
bors try

burrbull:
Rebased.

bors try

src/generate/register.rs
```diff
@@ -234,26 +235,43 @@ pub fn fields(
...
+                            let sequential_indexes = dim_index
+                                .iter()
+                                .map(|element| element.parse::<u32>())
+                                .eq((first..de.dim+first).map(Ok));
+                            if !sequential_indexes {
+                                return Err(format!("unsupported array indexes in {}", f.name))?;
+                            }
+                            (first, None)
+                        } else {
+                            (0, de.dim_index.clone())
```
Disasm: Which case is processed by this branch?

src/generate/register.rs
```diff
@@ -554,11 +606,18 @@ pub fn fields(
                 }
             });
 
+            let o_entry = if f.dim.is_some() {
+                quote! {o: usize,}
```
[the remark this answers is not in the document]
burrbull: Replace with `offset` ?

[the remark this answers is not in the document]
Disasm: Yes. Or at least document it somehow, so it will be clear that `o` represents an offset of the bit[s] inside the register word.

src/generate/register.rs
```diff
@@ -234,26 +235,43 @@ pub fn fields(
...
             };
 
+            let dim = match f {
+                Field::Array(_, de) => {
+                    let (first, index) = if let Some(dim_index) = &de.dim_index {
+                        if let Ok(first) = dim_index[0].parse::<u32>() {
+                            let sequential_indexes = dim_index
+                                .iter()
+                                .map(|element| element.parse::<u32>())
+                                .eq((first..de.dim+first).map(Ok));
```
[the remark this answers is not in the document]
burrbull: This check is to ensure that `dim_index` consist of correct sequence (e.g. `4,5,6,7`)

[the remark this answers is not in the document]
Disasm: Oh, indeed, I misunderstood the spec.

burrbull:
Rebased and fixed.

burrbull:
bors try

burrbull:
Now it must be easier to review.

burrbull:
cc @Disasm

Disasm:
Started reviewing again.
This change produces the same problem with docs (https://github.com/riscv-rust/k210-pac/commit/4f7330528951b6d1f54ed5030d7f17b0078cdce5):
```rust
#[doc = "Reader of field `pin%s`"]
pub type PIN_R = crate::R<bool, bool>;
```

Disasm:
Also,
```rust
        impl R {
            #[doc = "Enable channel %s"]
            #[inline(always)]
            pub unsafe fn ch_en(&self, n: usize) -> CH_EN_R {
                CH_EN_R::new(((self.bits >> n - 1) & 0x01) != 0)
            }
```

burrbull:
@Disasm try with last commits.

Disasm:
Thanks, it seems to be fixed for `R`, now it happens for `W`:
```rust
        impl W {
            #[doc = "Enable channel %s"]
            #[inline(always)]
            pub unsafe fn ch_en(&mut self, n: usize) -> CH_EN_W {
                CH_EN_W {
                    w: self,
                    offset: n - 1,
                }
            }
```

burrbull:
Rebased

Disasm:
Please, don't rebase without any good reason. It's quite difficult to review when you do this.

burrbull:
Must be fixed now.

Disasm:
Could you rebase your changes? Seems like there is a conflict with #423

burrbull:
> Could you rebase your changes? Seems like there is a conflict with #423

Done.

src/generate/register.rs
```diff
@@ -216,26 +213,19 @@ pub fn fields(
     let can_write = access != Access::ReadOnly;
 
     // TODO enumeratedValues
+    let inline = quote! { #[inline(always)] };
```
therealprof: Have you checked whether this is worth the harder to read quote sections below? If we do something like this we ideally wouldn't repeat the quoting for each register we process but do it exactly once and then reuse it.

Disasm:
I'd rather prefer readability over speed, because we already have relatively fast processing, but the code is quite difficult to understand.

therealprof:
> I'd rather prefer readability over speed, because we already have relatively fast processing, but the code is quite difficult to understand.

I agree. Also we could do a lot better explaining what the code is doing, especially for the more obscure and less common parts.

src/generate/register.rs
```diff
@@ -476,33 +558,74 @@ pub fn fields(
...
 
-            let doc = format!("Write proxy for field `{}`", f.name);
+            let (doc, offset_entry) = if let Some((_, _, _, _, suffixes_str)) = &field_dim {
+                (
+                    format!("Write proxy for fields `{}{}`", &name, &suffixes_str),
+                    quote! {offset: usize,},
+                )
+            } else {
+                (format!("Write proxy for field `{}`", f.name), quote! {})
+            };
```
Disasm: You can use the following syntax to make things nicer:
```rust
let doc;
let offset_entry;
if ... {
    doc = format!(...);
    offset_entry = quote! {...};
} else {
    doc = format!(...);
    offset_entry = quote! {};
}
```

burrbull: Done.

src/generate/register.rs
```diff
@@ -246,7 +239,6 @@ pub fn fields(
         let fty = width.to_ty()?;
         let evs = &f.enumerated_values;
         let quotedfield = String::from("`") + &f.name + "`";
```
Disasm: Maybe this `let quotedfield` line should be moved too. @therealprof could you comment on this?

src/generate/register.rs
```diff
@@ -414,7 +484,9 @@ pub fn fields(
                 if Some(evs) != evs_r.as_ref() {
                     pc_w = &new_pc_w;
                     if let Some(base) = base {
-                        let pc = base.field.to_sanitized_upper_case();
+                        let pc = util::replace_suffix(base.field, "")
+                            .to_sanitized_upper_case()
+                            .to_string();
```
Disasm: Do you need this `.to_string();`?

Disasm:
Hmm, there are still buggy doc comments:
```rust
        #[doc = "Reader of field `ch%s_abort`"]
        pub type CH_ABORT_R = crate::R<bool, bool>;
        #[doc = "Write proxy for field `ch%s_abort`"]
        pub struct CH_ABORT_W<'a> {
            w: &'a mut W,
        }
```

Disasm:
Also
```rust
        #[doc = "Reader of fields `ch_abort_we(1-6)`"]
        pub type CH_ABORT_WE_R = crate::R<bool, bool>;
        #[doc = "Write proxy for fields `ch_abort_we(1-6)`"]
        pub struct CH_ABORT_WE_W<'a> {
            w: &'a mut W,
            offset: usize,
        }
```
But sometimes it generates correct docs:
```rust
            #[doc = "Enable write to ch(1-6)_susp bit"]
            #[inline(always)]
            pub unsafe fn ch_susp_we(&self, n: usize) -> CH_SUSP_WE_R {
                CH_SUSP_WE_R::new(((self.bits >> n - 1 + 24) & 0x01) != 0)
            }
```

src/generate/register.rs
burrbull: > ```rust
> CH_ABORT_R
> ```

This field does not contain <dim> 
https://github.com/riscv-rust/k210-pac/blob/4f7330528951b6d1f54ed5030d7f17b0078cdce5/k210.svd#L1123
Please add there:
```
                            <dim>6</dim>
                            <dimIncrement>1</dimIncrement>
                            <dimIndex>1-6</dimIndex>
```

src/generate/register.rs
```diff
@@ -264,7 +256,41 @@ pub fn fields(
...
+                };
+                let suffixes_str = format!("({}-{})", first, first + de.dim - 1);
+                Some((first, de.dim, de.dim_increment, suffixes, suffixes_str))
+            }
+            Field::Single(_) => None,
+        };
+
         if can_read {
+            let readerdoc = if let Some((_, _, _, _, suffixes_str)) = &field_dim {
+                format!("Reader of fields `{}{}`", &name, &suffixes_str)
```
Disasm: Maybe this fixes the last issue
```suggestion
                format!("Reader of fields `{}`", util::replace_suffix(&f.name, &suffixes_str))
```

Disasm: The same for "Write proxy for fields"

src/generate/register.rs
```diff
@@ -476,33 +560,75 @@ pub fn fields(
...
                         self.w
                     }
                 }
             });
 
-            let doc = format!("Write proxy for field `{}`", f.name);
+            let doc;
+            let offset_entry;
+            if let Some((_, _, _, _, suffixes_str)) = &field_dim {
+                doc = format!("Write proxy for fields `{}`", util::replace_suffix(&name, suffixes_str));
```
Disasm:
```suggestion
                doc = format!("Write proxy for fields `{}`", util::replace_suffix(&f.name, suffixes_str));
```

src/generate/register.rs
therealprof: No preference. I planned to do something with it but that plan is already gone out the window with this PR anyway. 😅

Disasm: Ok! @burrbull, could you move it then to line 296?

burrbull:
I've already shut down my PC. But you can do it yourself. :)

Disasm:
Indeed!

burrbull:
`cargo fmt` and squash commits

PR #401, opened 2019-12-23
MSP430 Improvements

cr1901:
This PR brings MSP430 PAC generation up to parity with Cortex M PAC generation. Among other things, a `device.x` file is now generated for each MSP430 PAC, and the `interrupt!` macro has been replaced with an attribute macro from the not-yet-published `msp430-rt v0.2.0`. 

The remaining changes are docs and cleanups of the generated code; a number of features have become stable. Only `abi_msp430_interrupt` is not stable yet, and I have a workaround for this that I've not written yet.

Switching to an `interrupt` attribute macro is a breaking change, so I incremented the "major" version number.


src/lib.rs
```diff
@@ -71,17 +71,68 @@
...
+//! bare-metal = "0.2.0"
+//! msp430 = "0.1.0"
+//! vcell = "0.1.0"
+//!
+//! [dependencies.msp430-rt]
+//! optional = true
+//! version = "0.2.0"
+//!
+//! [features]
+//! rt = ["msp430-rt/device"]
```
Emilgardis: Haven't tested this but I think this needs to be added to svd2rust-regress

cr1901: Could you elaborate on what you mean by "adding this to sv2rust-regress"?

cr1901:
@Emilgardis (X-post from IRC) How did you specifically run svd2rust-regress (command line options) to break my PR? I've never actually used the regression suite before, so I'm unsure of what to look for.

cr1901:
@Emilgardis Could you please request a review from @awygle as well?

We also found an [issue](https://github.com/YuhanLiin/msp430fr2355-quickstart/issues/4#issuecomment-569164734) that will prevent us from merging as-is. There appears to be a bug somewhere between what TI provides us (`.dslite` files) and SVD generation of interrupt vectors. 

I'll have to think about how to properly handle this...

YuhanLiin:
Running `cargo run -- -a Msp430 -v` in the `svd2rust-regress` directory fails with `Expected content in <enumeratedValues> tag, found none`. This happens because the MSP430 SVD files used for regression tests are outdated and need to be regenerated.

pftbest:
@YuhanLiin 

svd2rust-regress has [old link in the code](https://github.com/rust-embedded/svd2rust/blob/master/ci/svd2rust-regress/src/tests.rs#L4223)

Here is a proper link for updated svd file:

https://raw.githubusercontent.com/pftbest/msp430g2553/v0.1.3-svd/msp430g2553.svd

YuhanLiin:
After updating the SVD file, svd2rust-regress for MSP430 fails with `error[E0554]: #![feature] may not be used on the stable release channel` on the line `#![cfg_attr(feature = "const-fn", feature(const_fn))]`. This happens on master as well.

cr1901:
>After updating the SVD file, svd2rust-regress for MSP430 fails with error[E0554]: #![feature] may not be used on the stable release channel on the line #![cfg_attr(feature = "const-fn", feature(const_fn))]. This happens on master as well.

I make no guarantees that `msp430 0.1.x` or `msp430-rt 0.1.x` compile on stable. If `svd2rust-regress` _must_ be run on stable, this means that my `svd2rust` improvements rely on `msp430 0.2.0` being published. 

I had intended _all_ the crates (`msp430`, `msp430-rt`, PACs) to be updated in lockstep _after_ the `svd2rust` changes were merged. I wanted to wait to update the crates just in case some bugs in the `msp430` crates surfaced while I made changes to `svd2rust`. This introduces a circular dependency.

Not sure how to proceed here. I guess I should publish `msp430 0.2.0` now to shut up `svd2rust-regress` and release a patch version if it becomes apparent there are bugs.

Emilgardis:
@cr1901 I think that's fine, don't push the 0.2.0 unless it feels like it's ready. svd2rust will choose whatever toolchain is being used by default, but it can also use nightly if passed `--toolchain=nightly`.

burrbull:
Rebase and squash your changes instead of merge master.

src/lib.rs
cr1901: @Emilgardis Okay, I think I have a grasp on this now.

I fixed the regression tests so the `msp430` test passes. As of `svd2rust` 17.0 plus this PR (which actually should probably be `18.0` since this is a breaking change), `svd2rust` will generate a PAC that is _not_ compatible with `msp430-rt v0.1.x`. Specifically:

* `INTERRUPTS` array has been renamed `__INTERRUPTS`; this matters to the linker script provided with `msp430-rt`.
* `interrupts!` macro replaces with attribute macro.

Right now, it seems nothing enables the `rt` feature of _any_ PACs (`lib.rs`) generated by the regression suite. Consequently, adding `msp430-rt v0.1.0` as a dep still passes- it's basically a no-op. Same situation would happen if I decided to publish and use `msp430-rt v0.2.0` as a prereq for this PR.

Since I'm going to be upgrading `msp430` packages in lockstep very soon, and it seems that `cortex-m` is also affected by not enabling the `rt` feature,  do you object if we hold off on adding this functionality to `svd2rust-regress` for now?

cr1901:
@burrbull Done. I kept the old branch locally in case it becomes necessary to use it again.

burrbull:
bors try

src/generate/interrupt.rs
```diff
@@ -136,7 +118,7 @@ pub fn render(
                 #[link_section = ".vector_table.interrupts"]
                 #[no_mangle]
                 #[used]
-                pub static INTERRUPTS:
+                pub static __INTERRUPTS:
```
burrbull: How will this affect other architectures?

cr1901: I changed the name for parity with `CortexM`. It doesn't affect other architectures because the match arm only runs when the arch is `Msp430`. The name should've been changed a while ago, but I only got around to it now.

burrbull:
Format with `cargo +stable fmt`, please.

therealprof:
@burrbull Is it possible to add a "fast fail" attribute to the rustfmt check only?

burrbull:
> @burrbull Is it possible to add a "fast fail" attribute to the rustfmt check only?

You mean stop all jobs when `rustfmt` fails?
I'm not an expert in `travis`.

cr1901:
@burrbull Pushed w/ formatting changes. Github decided for me to dismiss your review- I didn't do it.

src/lib.rs
Emilgardis: Fine for me :+1:

PR #402, opened 2019-12-24
Simplify Interrupt::nr

### LINKED ISSUE #211: Use custom discriminant values for Interrupts enum
I wondered why the `Nr` trait is implemented so complicated and long-winded with the `match` lookup. Wouldn't it be easier, and more performant (or: less reliant on compiler optimisations), to just use [Custom Discriminant Values](https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-field-less-enumerations) in the `enum` declaration - to allow a direct cast to/from an integer? Maybe even `#repr(u8)`?
If this sounds reasonable, I'll try to make a PR, it seems doable for a beginner like me.

Or maybe it would be nice to go even further and not use the "*IRQ number*" (as defined by CMSIS and represented in the [`Nr` trait](https://docs.rs/bare-metal/0.1.2/bare_metal/trait.Nr.html)) but the raw "*exception number*" as in the `IPSR`, `ICSR.VECTPENDING`, and `ICSR.VECTACTIVE` register fields. This would require an offset of 16, but I know only about Cortex-M4 machines. How about other processor types supported by svd2rust?

### LINKED ISSUE #370: Suboptimal code generated for Interrupt numbers
To contextualize, I'm using the nrf52840-pac crate version 0.8.0 (generated by svd2rust v0.14.0 but I think the issue still applies to the latest) and compiled using rustc 1.37.0-nightly (0af8e872e 2019-06-30) with `codegen-units=1`, `debug=true`, `lto=true`, and `opt-level="z"`. Now, with that out of the way.

The Interrupt enum has a variant for each hardware interrupt with a doc tag containing its number and lets the compiler assign the discriminant. To get the number from the enum variant, the `nr(&self) -> u8` function matches each variant to the hardware interrupt number. Unfortunately, the compiler turns this lookup into a jump table instead of encoding the interrupt number in the discriminant (perhaps because the interrupt numbers on the mentioned chip are discontinuous), wasting five bytes per interrupt plus the overhead of having a function do the work.

I manually changed the crate to set the discriminant of each variant to the interrupt number, but the compiled code did not change. I then changed the body of `nr(&self) -> u8` to `*self as u8` instead of the match and reduced code size by 304 bytes. Perhaps the generation should be changed so the code is generated in this form? That's a significant savings for a small transformation.

roblabla:
Make the Interrupt enum descriminant match the interrupt nr directly. This allows simplifying Interrupt::nr from being a complicated match to a simple cast to an u8.

According to #370, this should lead to a code reduction of as much as 304 bytes.

Fixes #211
Fixes #370

### DIFF
--- a/src/generate/interrupt.rs
+++ b/src/generate/interrupt.rs
@@ -149,12 +149,13 @@ pub fn render(
     let interrupt_enum = quote! {
         ///Enumeration of all the interrupts
         #[derive(Copy, Clone, Debug)]
+        #[repr(u8)]
         pub enum Interrupt {
             #(#variants)*
         }
 
         unsafe impl bare_metal::Nr for Interrupt {
-            #[inline]
+            #[inline(always)]
             fn nr(&self) -> u8 {
                 *self as u8
             }
--- a/src/generate/interrupt.rs
+++ b/src/generate/interrupt.rs
@@ -146,10 +146,11 @@ pub fn render(
         Target::None => {}
     }
 
-    let enum_repr = if variants.is_empty() {
-        quote!()
+    let self_token = quote!(self);
+    let (enum_repr, nr_expr) = if variants.is_empty() {
+        (quote!(), quote!(match #self_token {}))
     } else {
-        quote!(#[repr(u8)])
+        (quote!(#[repr(u8)]), quote!(*#self_token as u8))
     };
 
     let interrupt_enum = quote! {
@@ -162,8 +163,8 @@ pub fn render(
 
         unsafe impl bare_metal::Nr for Interrupt {
             #[inline(always)]
-            fn nr(&self) -> u8 {
-                *self as u8
+            fn nr(&#self_token) -> u8 {
+                #nr_expr
             }
         }
     };
--- a/src/generate/interrupt.rs
+++ b/src/generate/interrupt.rs
@@ -25,7 +25,6 @@ pub fn render(
     interrupts.sort_by_key(|i| i.value);
 
     let mut root = TokenStream::new();
-    let mut arms = vec![];
     let mut from_arms = vec![];
     let mut elements = vec![];
     let mut names = vec![];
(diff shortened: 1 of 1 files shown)


src/generate/interrupt.rs
```diff
@@ -161,9 +156,7 @@ pub fn render(
         unsafe impl bare_metal::Nr for Interrupt {
             #[inline]
```
burrbull:
```suggestion
             #[inline(always)]
```

src/generate/interrupt.rs
```diff
@@ -161,9 +156,7 @@ pub fn render(
         unsafe impl bare_metal::Nr for Interrupt {
```
burrbull:
```suggestion
         #[repr(u8)]
         unsafe impl bare_metal::Nr for Interrupt {
```

roblabla: I suppose this should go on the enum and not on the impl?

burrbull: yes

roblabla:
bors try

therealprof:
@roblabla You should now have try permissions.

burrbull:
bors try

roblabla:
Repr u8 seems to not work on uninhabited enums. Guess I’ll have to put the repr only if the interrupt enum is not empty?

burrbull:
> Repr u8 seems to not work on uninhabited enums. Guess I’ll have to put the repr only if the interrupt enum is not empty?

reasonably

roblabla:
bors try

roblabla:
bors try

roblabla:
bors try

roblabla:
bors try

burrbull:
bors retry

therealprof:
@roblabla Sorry I missed it earlier, but please squash your commits the next time so we don't have broken commits in the history.

PR #412, opened 2020-01-01
Simplify register generation code

therealprof:


therealprof:
bors try

src/generate/register.rs
```diff
@@ -223,113 +223,77 @@ pub fn fields(
...
+            evs,
             fields,
             parent,
             all_registers,
             peripheral,
             all_peripherals,
         )?;
 
-        let pc_r = &f.pc_r;
-        let mut pc_w = &f.pc_r;
```
burrbull: It's important!

therealprof: It's kind of hard to see without any comments in the code. ;) I'll have a look at the build results and try to incorporate it in my tests.

burrbull: Reader and writer must use one common `Enum_A` unless field have `enumeratedValues`: one for `read-only` and other for `write-only` (`Enum_AW`).

src/generate/register.rs
```diff
@@ -459,29 +422,28 @@ pub fn fields(
...
             if let Some((evs, base)) = lookup_filter(&lookup_results, Usage::Write) {
                 let variants = Variant::from_enumerated_values(evs)?;
 
-                if variants.len() == 1 << f.width {
+                if variants.len() == 1 << width {
                     unsafety = None;
                 }
 
                 if Some(evs) != evs_r.as_ref() {
-                    pc_w = &f.pc_w;
```
burrbull: And this.

therealprof:
Thanks for the explanation. I added a comment to that effect.

therealprof:
bors try

PR #416, opened 2020-01-01
More de vectorisation

therealprof:


burrbull:
bors try

therealprof:
No need to try it. If the build doesn't work it will just fail.

burrbull:
bors retry

Issue #418, opened 2020-01-02
Padding of registers with non 8-bit multiple sizes is incorrectly generated


david-sawatzke:
See https://github.com/im-tomu/fomu-pac-rs/issues/1, seems to stem from 
https://github.com/rust-embedded/svd2rust/blob/325a3044bdd2f4e8771ccd4a3a23e3a3df983de1/src/generate/peripheral.rs#L326 truncating instead of rounding up.

david-sawatzke:
Fixed in #419

PR #419, opened 2020-01-02
Round up when calculating register size in bytes

Tiwalun:
This fixes the issue in #418.

### DIFF
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -323,13 +323,13 @@ impl FieldRegions {
         let mut indices = Vec::new();
 
         let field_start = field.offset;
-        let field_end = field_start + field.size / BITS_PER_BYTE;
+        let field_end = field_start + (( field.size as f32 ) / (BITS_PER_BYTE as f32)).ceil() as u32;
 
         // The region that we're going to insert
         let mut new_region = Region {
             fields: vec![field.clone()],
             offset: field.offset,
-            end: field.offset + field.size / BITS_PER_BYTE,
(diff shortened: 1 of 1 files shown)


burrbull:
bors try

therealprof:
Seems a bit heavy and weird to do this with floating point operations.

src/generate/peripheral.rs
```diff
@@ -323,13 +323,13 @@ impl FieldRegions {
         let mut indices = Vec::new();
 
         let field_start = field.offset;
-        let field_end = field_start + field.size / BITS_PER_BYTE;
+        let field_end = field_start + (( field.size as f32 ) / (BITS_PER_BYTE as f32)).ceil() as u32;
```
therealprof: The more obvious way to achieve the same would be

```suggestion
        let field_end = field_start + (field.size + BITS_PER_BYTE - 1) / BITS_PER_BYTE;
```

Tiwalun:
You're right, it's neater this way. I applied your suggestion.

Tiwalun:
I'm wondering if this still has an issue with larger fields, for example if I declare a 17 bit wide register,
this will use a 24 bit width to calculate the offset.

However, the register width is mapped to to the next power of two, and the register field will actually be 32 bit wide.

https://github.com/rust-embedded/svd2rust/blob/325a3044bdd2f4e8771ccd4a3a23e3a3df983de1/src/generate/register.rs#L34

It might make more sense to use the actual width of the register in this calculation, instead of calculating it again.

PR #420, opened 2020-01-02
more TokenStream extend

burrbull:
in addition to #416 

r? @therealprof


burrbull:
bors try

therealprof:
Have you checked the performance? It seems that too much TokenStream extension makes everything slower...

therealprof:
Okay, this one doesn't negatively affect performance, though it doesn't improve it either. ;)

PR #423, opened 2020-01-06
Deduplicate some code and move some out of the inner loop

therealprof:

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -211,11 +211,14 @@ pub fn fields(
     r_impl_items: &mut TokenStream,
     w_impl_items: &mut TokenStream,
 ) -> Result<()> {
+    let span = Span::call_site();
+    let can_read = [Access::ReadOnly, Access::ReadWriteOnce, Access::ReadWrite].contains(&access);
+    let can_write = access != Access::ReadOnly;
+
     // TODO enumeratedValues
-    for f in fields.into_iter() {
+    for f in fields.iter() {
         // TODO(AJM) - do we need to do anything with this range type?
         let BitRange { offset, width, .. } = f.bit_range;
-        let span = Span::call_site();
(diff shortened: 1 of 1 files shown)


src/generate/register.rs
```diff
@@ -234,18 +237,19 @@ pub fn fields(
...
+        let can_write = can_write && (f.access != Some(Access::ReadOnly));
 
         let mask = 1u64.wrapping_neg() >> (64 - width);
         let hexmask = &util::hex(mask);
         let offset = u64::from(offset);
         let rv = reset_value.map(|rv| (rv >> offset) & mask);
         let fty = width.to_ty()?;
         let evs = &f.enumerated_values;
+        let quotedfield = String::from("`") + &f.name + "`";
+        let readerdoc = String::from("Reader of field ") + &quotedfield;
```
Disasm: Maybe it's better to have just one line for this. Is the performance impact significant compared to `format!()`?

therealprof: `core::fmt::write` makes up 17.5% of all CPU cycles used in svd2rust so I'm looking for ways to reduce its usage in addition to trying to have canned code available instead of regenerating everything with every invocation.

burrbull: I think this code is not faster.

therealprof: It's completely in the noise.

Disasm: What about merging into one line? It looks more readable and should be (insignificantly) faster:
```rust
let readerdoc = String::from("Reader of field `") + &f.name + "`";
```
If you prefer to leave things as they are, just bors r= me

therealprof: I have other plans with the quoted field name. ;) Moving out the readerdoc was just to get rid of the three separate definitions.

Still working on figuring out how to make `proc_macro2` play ball with respect to being able to pre-generate the TokenTrees/TokenStreams and simply change the field names

PR #428, opened 2020-01-11
fix async sanitizing

### LINKED ISSUE #386: `async` is a reserved keyword
It's easily fixed by search/replace whole word only and case sensitive `async` to `r#async` but I though it would be good to report this.

encountered when running `svd2rust` on http://www.keil.com/dd2/siliconlaboratories/efr32fg1p133f256gm48/

burrbull:
Fix #386


burrbull:
bors try

burrbull:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e4fa00961880e30527a7e57990e4eb56

jacobrosenthal:
This seems to have fixed our problem

PR #429, opened 2020-01-12
Add RegisterBlock trait

burrbull:
r? @therealprof 

cc @TeXitoi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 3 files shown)


burrbull:
Rebased.

cc @therealprof 

We need this if we want to have generic implementation of some code (without macros), but `Deref` is unavailable for some reason. For example when we split `USARTx` on `Tx` and `Rx`.

burrbull:
The idea was proposed by @TeXitoi  and instead of using macro let us to write something like
```rust
impl<USART> Tx<USART> where USART: RegisterBlock<RB=pac::USART1> {
    pub fn listen(&mut self) {
        unsafe { (*USART::rb()).cr1.modify(|_, w| w.txeie().set_bit()) };
    }
}
```

PR #430, opened 2020-01-14
Modules (`form`)

### LINKED ISSUE #424: Intergate `form`
What are you think about integrate `form` functionality in `svd2rust` (optional)?

When this option is enabled we can do not merge mod TokenStreams in one (and then split with `form`) and write them in different files directly. I think this can give us significant speed-up in this case.

@therealprof

burrbull:
Closes #424 .

r? @therealprof 

Speed tests are needed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- Option `-f`(`--form`) let you split `lib.rs` on modules, as the `form` tool does
+- Option `-o`(`--output`) let you specify output directory path
+
 ### Fixed
 
 - Keyword sanitizing (`async`)
--- a/src/generate/device.rs
+++ b/src/generate/device.rs
@@ -5,6 +5,7 @@ use std::fs::File;
 use std::io::Write;
 
 use crate::errors::*;
+use crate::modules::Module;
 use crate::util::{self, ToSanitizedUpperCase};
 use crate::Target;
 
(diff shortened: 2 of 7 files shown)


therealprof:
That's very nice! Need to have a closer look this evening.

burrbull:
@therealprof  you promised to look

therealprof:
Didn't get to it, sorry. Maybe I can squeeze it in later today.

therealprof:
I had a look. `svd2rust` runtime goes through the roof with the `-f`option, it takes 10x as long in my tests to dump the code. There's no change in build/check times of the generated code.

It would be great to see some noticeable benefit somewhere.

therealprof:
The `Module` is a rather weird implementation causing quite a few clippy lints and warnings in case the superfluous `#[allow(dead_code)]` are removed.

We should also get rid of all the panics and instead implement proper error handling.

duskmoon314:
Any plan on this?

Recently I tried the logic of `xtask` in `esp-pacs` on `d1-pac`, which first generates `lib.rs` and then transforms it into modules in `src`. Much simpler than this implementation in my mind.
[esp-pacs](https://github.com/esp-rs/esp-pacs/blob/main/xtask/src/main.rs#L145-L161)
[d1-pac](https://github.com/duskmoon314/aw-pac/blob/main/xtask/src/main.rs#L126-L143)

> it takes 10x as long in my tests to dump the code.

I am not familiar with improving performance. Taking a simple look at the flame graph of executing `xtask gen d1`, I found that `form::create_directory_structure` is indeed costly. But I don't feel it annoying since I always use `form` after generating code with `svd2rust`.

The flame graph:
flamegraph

rmsyn:
@burrbull you mentioned that you would like me to take a look at bringing this PR over the finish line.

Happy to do so. Would you mind rebasing on the latest `HEAD` to make finishing the work a bit easier?

burrbull:
> @burrbull you mentioned that you would like me to take a look at bringing this PR over the finish line.
> 
> Happy to do so. Would you mind rebasing on the latest `HEAD` to make finishing the work a bit easier?

I need to remember what I've tried to do and try to solve perfomance issues.

PR #431, opened 2020-01-15
split out register size type from ResetValue trait

burrbull:


therealprof:
What's the benefit of this change?

burrbull:
@therealprof  This change is very important in fact.

The advantages are:
1) `RawType::Ux` is auto-implemented for all registers and can't be implemented in incorrect way somewhere else.
     Same time not all registers can have `ResetValue`.
2) Slightly less of generated code.
3) **`RawType` trait can be used in other interfaces.**
     For example, several months ago I experimented with wrapper `Read/Write/Modify/SetBits` traits    (draft is [here](https://github.com/rust-embedded/svd2rust/blob/pa-traits/pa/src/lib.rs)). The truth is I have not come up with a real application for them.

PR #436, opened 2020-04-18
Use standard wording for the MSRV note

ra-kete:
Part of [wg/#445](https://github.com/rust-embedded/wg/issues/445).

svd2rust already has its MSRV documented and tested, but @adamgreig suggested updating it to use the standard wording. Since this is a special case in that the MSRV applies to the generated code, not svd2rust itself, the standard wording doesn't fully apply, so I changed it in a (hopefully) sensible way.

An open question is: Should we also document the MSRV of svd2rust itself? Looking at how the CI works, currently it's implicitly guaranteed to be the same version als the MSRV for the generated code, so we could just extend the note in the README to also include svd2rust itself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,8 +58,6 @@ matrix:
     # - env: TARGET=x86_64-unknown-linux-musl
 
     # OSX
-    - env: TARGET=i686-apple-darwin
-      os: osx
     - env: TARGET=x86_64-apple-darwin
       os: osx
 
--- a/README.md
+++ b/README.md
@@ -11,9 +11,9 @@ This project is developed and maintained by the [Tools team][team].
 
 # [API](https://docs.rs/svd2rust)
 
-## Requirements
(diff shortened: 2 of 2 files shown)


ra-kete:
The built failed because the i686-apple-darwin Target is no longer fully supported: https://blog.rust-lang.org/2020/01/03/reducing-support-for-32-bit-apple-targets.html

I suggest we simply remove it from the CI, unless there is a reason svd2rust still needs to support it?

therealprof:
@ra-kete Yeah, let's remove it. That target is pretty silly anyway: pretty much only the very first MacBooks had only 32bit Intel MCUs, everything else had 64Bit MCUs from the start and 64Bit support was there since Lion (2011). That target was useful for legacy code bases which I highly doubt even exist since Rust 1.0 was released 4 years after Apple started actually **requiring** 64bit capable CPUs...

PR #438, opened 2020-04-23
Remove unnecessary coercions to 'String' + small fixes to doc comments

JOE1994:
Remove unnecessary coercions to 'String' + small fixes to doc comments

-------------------------------------
Thank you for reviewing this PR 😸

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,8 +3,8 @@
 //! [CMSIS-SVD]: http://www.keil.com/pack/doc/CMSIS/SVD/html/index.html
 //!
 //! A SVD file is an XML file that describes the hardware features of a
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -240,15 +240,15 @@
 //! # `read` / `modify` / `write` API
 //!
 //! Each register in the register block, e.g. the `cr1` field in the `I2C` struct, exposes a
-//! combination of the `read`, `modify` and `write` methods. Which methods exposes each register
+//! combination of the `read`, `modify`, and `write` methods. Which method exposes each register
```
burrbull, line 243: Looks good to me except these strange commas before `and`.

therealprof: Those are Oxford commas.

PR #439, opened 2020-04-27
Add commit and date info to generated crates

cr1901:
I don't need this feature as much now, but I know more than once in the past while adding msp430 feature I've forgotten that my changes haven't been incorporated into a release yet.

This means the docs.rs docs are out of date relative to the PAC. I've added some info to the doc string to mention the commit and date that svd2rust was compiled with, and explicitly how to generate up-to-date svd2rust docs.


src/generate/device.rs
```diff
@@ -20,13 +20,29 @@ pub fn render(
 ) -> Result<TokenStream> {
     let mut out = TokenStream::new();
 
+    let commit_info = {
+        let tmp = include_str!(concat!(env!("OUT_DIR"), "/commit-info.txt"));
+
+        if tmp.is_empty() {
+            " (unknown commit)"
```
Emilgardis, line 27: I'd rather this be ` (untracked)`  or ` (???????)`

src/generate/device.rs
```diff
@@ -20,13 +20,29 @@ pub fn render(
...
+        }
+    };
+
     let doc = format!(
         "Peripheral access API for {0} microcontrollers \
-         (generated using svd2rust v{1})\n\n\
-         You can find an overview of the API [here].\n\n\
-         [here]: https://docs.rs/svd2rust/{1}/svd2rust/#peripheral-api",
+         (generated using svd2rust v{1}{2})\n\n\
+         You can find an overview of the API as of release v{1} [here].\n\n\
```
Emilgardis, line 36:
```suggestion
         You can find an overview of the generated API [here].\n\n\
```

src/generate/device.rs
```diff
@@ -20,13 +20,29 @@ pub fn render(
...
+         Docs containing API features to be included in the [next] \
+         release can be generated by cloning the svd2rust [repository], \
+         checking out the above commit, and running `cargo doc`.\n\n\
```
Emilgardis, line 39:
```suggestion
         API features to be included in the [next] svd2rust \
         release can be generated by cloning the svd2rust [repository], \
         checking out the above commit, and running `cargo doc --open`.\n\n\
```

cr1901:
@Emilgardis 

>Also, I would like it if we switched over to println!("cargo:rustc-env=REV={}", git_rev), but that's another pr.

FWIW, `msp430_svd` uses the [`vergen`](https://crates.io/crates/vergen) crate to do what you described. See its [`build.rs`](https://github.com/pftbest/msp430_svd/blob/master/build.rs).

PR #442, opened 2020-05-07
Support as_ptr on a register for DMA and such

petrohi:


src/generate/generic.rs
```diff
@@ -32,6 +32,21 @@ pub struct Reg<U, REG> {
 
 unsafe impl<U: Send, REG> Send for Reg<U, REG> {}
 
+impl<U, REG> Reg<U, REG>
+where
+    U: Copy,
+{
+    /// Gets a memory address of a register.
```
Emilgardis, line 39: Wording, maybe something like

    Returns the underlying memory address of register

petrohi: Adjusted.

burrbull:
bors try

therealprof:
@burrbull No need to "try" anymore. GHA is automatically running CI on every commit, cf. https://github.com/rust-embedded/svd2rust/runs/653370311

burrbull:
reflex :smile:

therealprof:
bors r=Emilgardis

PR #444, opened 2020-05-12
Cluster name qualification

gkelly:
This contains two commits, the first is a simple typo fix. The second ensures that the complete-pathed type name is generated for clusters.

### DIFF
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -155,7 +155,7 @@ pub fn render(
         mod_items.extend(cluster_block(c, &defaults, p, all_peripherals, nightly)?);
     }
 
-    // Push all regsiter realted information into the peripheral module
+    // Push all register related information into the peripheral module
     for reg in registers {
         mod_items.extend(register::render(
             reg,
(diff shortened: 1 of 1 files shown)


burrbull:
Thanks.
bors r+

burrbull:
Something strange with `dimable_name_check` regex:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cc2c9f848d4144df27f48af6ed6c516e

It's always `true`!!!


`derived_from` fields also can contain dots inside

gkelly:
> Something strange with `dimable_name_check` regex:
> https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cc2c9f848d4144df27f48af6ed6c516e
> 
> It's always `true`!!!
> 
> `derived_from` fields also can contain dots inside

Hmm. I wonder if the `^$` are binding to the `|` groups instead of the whole pattern. [Looks like it](https://regexper.com/#%5E%28%28%25s%29%7C%28%25s%29%5B_A-Za-z%5D%7B1%7D%5B_A-Za-z0-9%5D*%29%7C%28%5B_A-Za-z%5D%7B1%7D%5B_A-Za-z0-9%5D*%28%5C%5C%5B%25s%5C%5C%5D%29%3F%29%7C%28%5B_A-Za-z%5D%7B1%7D%5B_A-Za-z0-9%5D*%28%25s%29%3F%5B_A-Za-z0-9%5D*%29%24). I can do a follow up PR to fix that.

gkelly:
Created https://github.com/rust-embedded/svd/pull/122 to address the naming issue.

PR #445, opened 2020-05-20
Add ESP32 support

arjanmels:
Adding support for ESP32 as used in new pull requests on https://github.com/esp-rs.

ESP32 has 2 level interrupt handling. First level in the cores. Second level in the periphery/software.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
         rust: [stable]
 
         # All vendor files we want to test on stable
-        VENDOR: [rustfmt, Atmel, Freescale, Fujitsu, Holtek, Nordic, Nuvoton, NXP, RISC-V, SiliconLabs, Spansion, STMicro, Toshiba, Espressif]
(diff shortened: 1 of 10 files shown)


burrbull:
can you add any tests for esp32?

arjanmels:
Ok, will look into it, but will be next week.

arjanmels:
@burrbull @therealprof Thanks for the lightning fast review!

therealprof:
bors ping

therealprof:
bors retry

burrbull:
with new `svd-parser`:
[ERROR svd2rust] In device Espressif

Caused by:
    0: In peripheral `AES`
    1: Peripheral have `registers` tag, but it is empty

`registers` tag can't be empty according to https://www.keil.com/pack/doc/CMSIS/SVD/html/elem_registers.html

arjanmels:
@burrbull, @mabezdev Ok, I will look into fixing teh example svd, hopefully later today, but cannot promise.

arjanmels:
@burrbull I updated the svd file, so now should be ok. (I tested it with branch svd_new, but it also passed before the updates, so I might have done something wrong.)

PR #447, opened 2020-06-02
update svd parser

burrbull:


Emilgardis:
Converted to draft

burrbull:
Problems (updatable):
1.
[ERROR svd2rust] In device `nrf51`
    
    Caused by:
        0: In peripheral `WDT`
        1: In register `RR[%s]`
        2: Parsing field #0
        3: In field `RR`
        4: Value 1850885685 out of range 0..0

burrbull:
> @burrbull I updated the svd file for Espressif/ESP32, so now should be ok. (I tested it with branch svd_new, but it also passed before the updates, so I might have done something wrong.)

I still see empty `<registers />` in the file. Just remove them.

burrbull:
cc @therealprof

therealprof:
Why are we removing the Spansion tests again?

burrbull:
> Why are we removing the Spansion tests again?

Spansion SVDs are same as Fujitsu's ones, but more old versions of them with lot of bugs, that are fixed in newer.
So I've left only not doubled files.

therealprof (approved):
Great then, thanks!

bors r+

PR #460, opened 2020-08-05
Rework `Interrupt` enum for MSP430

therealprof:
This also removes the `bare-metal` dependency from PACs created for
MSP430, as requested in
https://github.com/rust-embedded/svd2rust/pull/455#issuecomment-665214017

### DIFF
--- a/src/generate/interrupt.rs
+++ b/src/generate/interrupt.rs
@@ -160,41 +160,54 @@ pub fn render(
         (quote!(#[repr(u8)]), quote!(*#self_token as u8))
     };
 
-    let interrupt_enum = quote! {
-        ///Enumeration of all the interrupts
-        #[derive(Copy, Clone, Debug, PartialEq, Eq)]
-        #enum_repr
-        pub enum Interrupt {
-            #variants
-        }
-
-        unsafe impl bare_metal::Nr for Interrupt {
-            #[inline(always)]
-            fn nr(&#self_token) -> u8 {
-                #nr_expr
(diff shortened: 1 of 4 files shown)


therealprof:
CC @cr1901

cr1901:
@therealprof Unable to test right now... it looks fine to me. Since `msp430` crate at present relies on `bare_metal` to import `CriticalSection` and `Mutex`, I think I'll need to bump the patch version and require that anyone generating a PAC use at least a (to-be-published) `v0.2.3` version of `msp430` crate.

However, this doesn't affect the public API and it's an impl detail that `msp430` still depends on `bare_metal`, so no rush.

therealprof:
Since the generated PACs don't depend on `bare-metal` anymore you are free to use any version you'd like in your application or other dependencies without conflict.

src/generate/interrupt.rs
```diff
@@ -176,7 +189,7 @@ pub fn render(
         }
     };
 
-    if target == Target::CortexM || target == Target::Msp430 {
+    if target == Target::CortexM {
         root.extend(interrupt_enum);
     } else {
```
adamgreig, line 194: Won't this `else` branch still trigger, causing `#interrupt_enum` to end up in `mod_items` along with the `try_from` impl etc?

therealprof: Good catch. I wonder why this didn't cause any CI problems...

therealprof:
@adamgreig Are we good to go now?

PR #463, opened 2020-08-06
Register trait improvements

### LINKED ISSUE #459: Awkward to write code generic over registers
I am writing code that is generic over registers, and I'm getting a bit frustrated by the awkward bounds that need to be applied.

The ideal case would look something like this:

```rust
fn read_bit<U: Copy, Register: Readable>(reg: &Reg<U, Register>) -> U {
    // ...
}
```

Unfortunately, the `Readable` and `Writable` traits are impl'd on the `Reg` struct typedef itself, not on the inner type.  Also unfortunately, the traits don't provide any functionality directly, so I can't rely completely on the traits and forget about the `Reg` struct.  This means the bounds need to be written in the much more awkward form:

```rust
fn read_bit<U: Copy, Register>(reg: &Reg<U, Register>) -> U
where
    Reg<U, Register>: Readable,
{
    // ...
}
```

In my case, I'd like to apply the bounds to associated types in a trait.  Though I can add the relevant where bounds on `Reg<...>` to the trait definition, they don't "propagate" to uses of that trait, meaning I need to repeat them on any block that uses the trait.

Ideally it would look something like:

```rust
trait Port {
    type U: Sized + Copy + BitAnd + BitOr + ...;
    type DDR: Readable + Writable;
    type Data: Readable + Writable;

    fn ddr() -> Reg<Self::U, Self::DDR>;
    fn data() -> Reg<Self::U, Self::Data>;
}

impl<P: Port> Component<P> {
    fn configure() {
        P::ddr().modify(|r, w| {
            // ...
        });
    }

    fn do_something() {
        P::data().modify(|r, w| {
            // ...
        });
    }
}
```

But, given the above constraints, I actually need to write:

```rust
trait Port { // I could put a where clause here but it really wouldn't help
    type U: Sized + Copy + BitAnd + BitOr + ...;
    type DDR;
    type Data;

    fn ddr() -> Reg<Self::U, Self::DDR>;
    fn data() -> Reg<Self::U, Self::Data>;
}

impl<P: Port> Component<P>
where
    Reg<P::U, P::DDR>: Readable + Writable
{
    fn configure() {
        P::ddr().modify(|r, w| {
            // ...
        });
    }
}

impl<P: Port> Component<P>
where
    Reg<P::U, P::Data>: Readable + Writable
{
    fn do_something() {
        P::data().modify(|r, w| {
            // ...
        });
    }
}
```

I believe that this situation could be resolved by generating the `impl Readable`/`impl Writable` for the hidden inner register type rather than directly on the outer register type alias.  For documentation purposes, there could be a blanket impl on `Reg`, like:

```rust
impl<U, REG: Readable> Readable for Reg<U, REG> {}
impl<U, REG: Writable> Writable for Reg<U, REG> {}
```

Which I believe would be allowed due to coherence.

If you are open to these changes, I would be willing to work on a PR implementing them.  Or if you have suggestions for another strategy to make this usage more ergonomic, I'm all ears.

I spent a while looking around for related discussion and found none.  Feel free to point me to any existing documentation or prior discussion on this topic.

Thanks for all your hard work on this!

couchand:
Shuffling around type and trait definitions with the goal of improving the generated API & documentation.  Some highlights:

- Moves relevant trait impls from register type alias to register spec ZST, which fixes #459 and improves documentation output.
- Removes internal use of register type alias in favor of transparent usage of `Reg<..>`.  While the type alias is convenient, it hides critical implementation details.
- Wraps register reader/writer and field readers in newtype wrappers, which significantly improves the documentation output.

My testing so far has been admittedly much too limited given the scope of the changes.  However, for the binaries I've checked, the assembly output is identical before and after these changes, suggesting that they are all indeed only type-level cleanup.  I've also not run into any examples of user code breakage, despite the public API changes (though I would not be suprised to find some).  Spot checking with `scd2rust-regress` has found no issues so far.

Some notable design decisions worth discussing:

- The register spec ZST becomes much more important, so I've renamed it to remove the leading underscore.  To avoid conflicting with the register type alias, I've moved it into the register module.  Should the name have a suffix?  Should it be located in the parent module alongside the (now less-important) type alias?
- I've removed internal use (notably on generated `RegisterBlock`s) of the register type alias, as I think it adds a speed bump to understanding how the generated code works.  Should the alias continue to be used there instead?
- If breaking changes were on the table, I'd suggest changing the API of `Reg::write`.  Requiring the closure to return the input doesn't seem like the right thing to do.  It sows doubt about how exactly the changes propagate, and it makes one-liners slightly shorter at the expense of making multi-line closures an extra line.

Some documentation highlights:

- Boolean readers now show accessor options directly.
svd2rust-bool-field

- Field readers and writers show `bits` accessors.
svd2rust-reg-reader
svd2rust-reg-writer-2

- Generic `R` type **doesn't** show every specific type's impls.
svd2rust-generic-r-impls

### DIFF
--- a/src/generate/generic.rs
+++ b/src/generate/generic.rs
@@ -9,14 +9,19 @@ pub trait Register {
 /// Trait implemented by readable registers to enable the `read` method.
 ///
 /// Registers marked with `Writable` can be also `modify`'ed.
-pub trait Readable: Register {}
+pub trait Readable: Register {
+    /// Result from a call to `read` and argument to `modify`.
+    type Reader: core::convert::From<R<Self>> + core::ops::Deref<Target = R<Self>>;
+}
 
 /// Trait implemented by writeable registers.
 ///
 /// This enables the  `write`, `write_with_zero` and `reset` methods.
 ///
 /// Registers marked with `Readable` can be also `modify`'ed.
-pub trait Writable: Register {}
+pub trait Writable: Register {
+    type Writer: core::convert::From<W<Self>> + core::ops::Deref<Target = W<Self>>;
+}
 
 /// Reset value of the register.
 ///
@@ -61,11 +66,11 @@ impl<REG: Readable> Reg<REG> {
     /// let flag = reader.field2().bit_is_set();
     /// ```
     #[inline(always)]
-    pub fn read(&self) -> R<REG> {
-        R {
+    pub fn read(&self) -> REG::Reader {
+        REG::Reader::from(R {
             bits: self.register.get(),
             _reg: marker::PhantomData,
-        }
+        })
     }
 }
 
@@ -96,13 +101,13 @@ impl<REG: Resettable + Writable> Reg<REG> {
     #[inline(always)]
     pub fn write<F>(&self, f: F)
     where
-        F: FnOnce(&mut W<REG>) -> &mut W<REG>,
+        F: FnOnce(&mut REG::Writer) -> &mut REG::Writer,
     {
         self.register.set(
-            f(&mut W {
+            f(&mut REG::Writer::from(W {
                 bits: REG::reset_value(),
                 _reg: marker::PhantomData,
-            })
+            }))
             .bits,
         );
     }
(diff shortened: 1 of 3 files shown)


couchand:
@therealprof tagging you since you commented on #459.

@burrbull tagging you because I was inspired by a number of your ideas in https://github.com/rust-embedded/svd2rust/blob/pa-traits/pa/src/lib.rs

couchand:
Well I'll be.  Some architectures have registers named `R`.  That sure throws a wrench in my naming scheme.  I'm open to any suggestions.

burrbull:
I see 3 big changes in your PR. Can you split them in different PRs, please? If it is possible. I will try to review tomorrow.

couchand:
Closing in favor of #464, #465, and #466.

PR #464, opened 2020-08-07
Register spec & traits

### LINKED ISSUE #459: Awkward to write code generic over registers

couchand:
Part 1 of #463.

- Publishes the register spec zero-sized type and moves all relevant register traits to that struct, which fixes #459 and improves generated documentation.
- Removes the extra type parameter on `Reg`, making the register spec the sole authority on the shape of the register (but not its address).

One thing that's changed since the previous PR, this renames the spec struct so that it won't cause a conflict on NXP, Freescale, and possibly other architectures.  I've gone with `register::REGISTER_SPEC` since that seems reasonably clear and I can't think of how it would cause a conflict (but I'm ready to be proven wrong since I've tried several ideas only to be stymied by various architectures' oddball register naming patterns).

I'd prefer the spec to be in the parent module, but I can't come up with a way to do that which guarantees that the generated code will be free from naming conflicts.  This seems good enough, plus it maintains the rule that when we make up a name it belongs in the child mod (except for `RegisterBlock`).

There is potential for breaking changes here, though the impact seems pretty minimal.  Specifically:
- If users referred to the traits in `mod generic`, their usage would break.  Writing code referring to those traits was previously pretty awkward (see #463), and this PR makes such usage much more ergonomic, so it would seem to be a net benefit.
- If users named the full reader/writer incantation the reference would break.  It doesn't seem users are likely to name the readers/writers at all, but if they do it seems very likely they would use the type alias.
- If users named the previously `#[docs(hidden)] ` spec struct the reference would break.  I don't see why they would have done so.
- If users patched something in to the generated module with the same name as our new register spec, it would break.  That case seems highly unlikely.

r? @burrbull

### DIFF
--- a/src/generate/generic.rs
+++ b/src/generate/generic.rs
@@ -1,38 +1,28 @@
 use core::marker;
 
+/// Raw register type
 pub trait Register {
+    /// Raw register type (`u8`, `u16`, `u32`, ...).
     type Ux: Copy;
-}
-
-pub trait ReadableRegister: Register {}
-pub trait WritableRegister: Register {}
-pub trait ResettableRegister: Register {
-    fn reset_value() -> Self::Ux;
 }
 
 /// Trait implemented by readable registers to enable the `read` method.
 ///
 /// Registers marked with `Writable` can be also `modify`'ed.
-pub trait Readable {}
+pub trait ReadableRegister: Register {}
 
 /// Trait implemented by writeable registers.
 ///
 /// This enables the  `write`, `write_with_zero` and `reset` methods.
 ///
 /// Registers marked with `Readable` can be also `modify`'ed.
-pub trait Writable {}
+pub trait WritableRegister: Register {}
 
-
-/// Raw register type (autoimplemented for `Reg` type)
-pub trait RawType {
-    /// Raw register type (`u8`, `u16`, `u32`, ...).
-    type Ux: Copy;
-}
 /// Reset value of the register.
 ///
 /// This value is the initial value for the `write` method. It can also be directly written to the
 /// register by using the `reset` method.
-pub trait ResetValue: RawType {
+pub trait ResettableRegister: Register {
     /// Reset value of the register.
     fn reset_value() -> Self::Ux;
 }
@@ -41,15 +31,6 @@ pub trait ResetValue: RawType {
 pub struct Reg<REG: Register> {
     register: vcell::VolatileCell<REG::Ux>,
     _marker: marker::PhantomData<REG>,
-}
-
-impl<REG: ReadableRegister> Readable for Reg<REG> {}
-impl<REG: WritableRegister> Writable for Reg<REG> {}
-impl<REG: ResettableRegister> ResetValue for Reg<REG> {
-    #[inline(always)]
-    fn reset_value() -> Self::Ux {
-        REG::reset_value()
-    }
 }
 
 unsafe impl<REG: Register> Send for Reg<REG> where REG::Ux: Send {}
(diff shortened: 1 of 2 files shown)


burrbull:
Looks good in first view. One question. Have you tested this for result firmware size? In release and debug mode.

couchand:
> Have you tested this for result firmware size? In release and debug mode.

In my testing of binaries built using code generated from the previous PR, the assembly produced (`RUSTFLAGS='--emit asm'`) was identical to master.

I'll repeat that testing here to make sure I've maintained that property.

couchand:
arduino uno blink example

bytes | master | #464 
-|---------|---
*debug* | 1090816 | 1094812
*release* | 7016 | 7016

couchand:
After `strip --strip-unneeded`, the numbers are:

bytes | master | #464
--|--|--
debug | 11148 | 11148
release | 2444 | 2444

PR #465, opened 2020-08-07
Accessor newtypes

couchand:
Part 2 of #463, builds on #464.

- Wraps register reader/writer and field readers in newtype wrappers, which significantly improves the documentation output.

Again there's some minor potential for breaking changes:

- If users named the complete type of the previous type alias, it would break.  It seems more likely that they would use the alias, which will silently upgrade to a reference to the newtype.

r? @burrbull

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -57,7 +57,21 @@ pub fn render(
         let desc = format!("Reader of register {}", register.name);
         mod_items.extend(quote! {
             #[doc = #desc]
-            pub type R = crate::R<#name_uc_spec>;
+            pub struct R(crate::R<#name_uc_spec>);
+
+            impl core::ops::Deref for R {
+                type Target = crate::R<#name_uc_spec>;
+
+                fn deref(&self) -> &Self::Target {
+                    &self.0
+                }
+            }
+
+            impl core::convert::From<crate::R<#name_uc_spec>> for R {
+                fn from(reader: crate::R<#name_uc_spec>) -> Self {
+                    R(reader)
+                }
+            }
         });
         methods.push("read");
     }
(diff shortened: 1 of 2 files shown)


couchand:
bytes | master | #464 | #465 
--|--|--|--
debug | 1090816 | 1094812 | 1471756
release | 7016 | 7016 | 105180

Oh my this really blows up the size of the release build!  I thought that the newtype pattern was basically optimized away by the compiler, but it seems that either something is preventing that from happening or the compiler's handling of it is not as good as I thought it was.

couchand:
Upon closer inspection, most of that extra junk is in `.symtab`.  After `strip --strip-unneeded`, the numbers are:

bytes | master | #464 | #465 
--|--|--|--
debug | 11148 | 11148 | 11772 
release | 2444 | 2444 | 2732

My intuition is that `.symtab` won't get shipped to flash anyway, so these numbers would be more representative of the real impact.  Is that right?

As a general matter, what's the best way to do this sort of profiling?

couchand:
Ok I've stumbled onto `size`, which seems appropriate for the job.  Running that against the binaries here gives me:

debug | master | #464 | #465 
--|--|--|--
text | 9386 | 9386 | 10010
data | 1250 | 1250 | 1250
bss | 1 | 1 | 1

release | master | #464 | #465 
--|--|--|--
text | 1862 | 1862 | 2150
data | 74 | 74 | 74
bss | 1 | 1 | 1

src/generate/register.rs
```diff
@@ -50,36 +51,62 @@ pub fn render(
...
         let desc = format!("Reader of register {}", register.name);
         mod_items.extend(quote! {
             #[doc = #desc]
-            pub type R = crate::R<#rty, super::#name_pc>;
+            pub struct R(crate::R<#name_uc_spec>);
+
+            impl core::ops::Deref for R {
+                type Target = crate::R<#name_uc_spec>;
+
+                fn deref(&self) -> &Self::Target {
```
burrbull, line 65: Add `#[inline(always)]` for each `Deref` and `DerefMut` first and retest.
Do you use `cargo size` from `cargo-binutils` or something else?

couchand: That did it!  Now it's exactly the same size.

These numbers are from my system `avr-size`.  I just tried installing `cargo-binutils` and I'm getting 'Failed to parse crate metadata'.

couchand:
Rebased on top of master, please take another look.

---

Re: https://github.com/rust-embedded/svd2rust/pull/466#issuecomment-670553172

For `Rahix/avr-device`:

```
$ du --summarize src-*
8884	src-465
7920	src-master
```

For `lpc-rs/lpc-pac`:

```
$ du --summarize lpc*-*
4540	lpc11uxx-465
4168	lpc11uxx-master
7072	lpc176x5x-465
6296	lpc176x5x-master
278760	lpc54606-465
277548	lpc54606-master
4720	lpc82x-465
4212	lpc82x-master
118212	lpc845-465
117632	lpc845-master
```

There's some increase in the generated code size, but not a lot.

therealprof:
Size increase looks reasonable. There's still a lot of (medium high hanging) fruit to pick.

I'll soft approve and let give @burrbull have a second look.

PR #466, opened 2020-08-07
Avoid register type aliases

couchand:
Part 3 of #463, builds on #465.

- Removes generated use of the register type aliases in favor of directly referencing `Reg<REGISTER_SPEC>`.  I think this helps clarify how the generated and generic types work together.

No breaking change concerns on this change set.

r? @burrbull

### DIFF
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -799,7 +799,7 @@ fn expand_svd_register(
             for (idx, _i) in indices.iter().zip(0..) {
                 let nb_name = util::replace_suffix(&info.name, idx);
 
-                let ty = name_to_ty(&ty_name, name)?;
+                let ty = name_to_wrapped_ty(&ty_name, name)?;
 
                 out.push(new_syn_field(&nb_name.to_sanitized_snake_case(), ty));
             }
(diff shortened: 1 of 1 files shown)


therealprof:
I'm a bit worried about generated source code size here. Do you have any numbers how much the source code blows up with that change? The compiler is already not too swift (yeah, sorry for the euphemism) thanks to the massive amount of generated code; we should try our best not to make it worse...

burrbull:
Yes, code size and compilation time questions are opened. We can reduce code size using `derive_more::Deref/DerefMut/From`, but this will require `proc_macro2` in dependencies.

couchand:
Rebased on master, leaving out the commits from #465 since the two PRs are completely independent.  Please take another look.

@therealprof I suspect your question is more geared towards #465 rather than here, but I'll answer it in both places.

For `Rahix/avr-device`:

```
$ du --summarize src-*
7928	src-466
7920	src-master
```

For `lpc-rs/lpc-pac`:

```
> du --summarize lpc*-*
4176	lpc11uxx-466
4168	lpc11uxx-master
6304	lpc176x5x-466
6296	lpc176x5x-master
277572	lpc54606-466
277548	lpc54606-master
4228	lpc82x-466
4212	lpc82x-master
117640	lpc845-466
117632	lpc845-master
```

therealprof:
> @therealprof I suspect your question is more geared towards #465 rather than here, but I'll answer it in both places.

Oops, yes you're right.

PR #468, opened 2020-08-17
Improve documentation on generated registers and fields

couchand:
Previously, the descriptions for fields were only listed on the docs for the accessor methods on the `R` and `W` structs.  This adds the descriptions to the field accessor structs so they appear on the register module page, too.


therealprof:
bors retry

PR #469, opened 2020-08-29
Remove `_mut` version of alternate register access function

burrbull:
As we use `vcell` to access register contents, we don't need mutable references for this.


therealprof (approved):
Nice catch. LGTM.

bors r+

PR #472, opened 2020-09-20
Update svd2rust-regress once again

therealprof:
This bumps the arch dependencies to the current level, removes the
"const-fn" flag support and uses newer reqwest and rayon dependencies
to get rid of some 50 crate dependencies.


adamgreig (approved):
bors merge

PR #473, opened 2020-09-20
Implement new InterruptNumber trait from next cortex-m release

therealprof:
This eliminates the bare-metal and cortex-m-rt dependencies for Cortex-M
and thus changes CI and svd2rust-regress accordingly.

Replaces #455

### DIFF
--- a/src/generate/interrupt.rs
+++ b/src/generate/interrupt.rs
@@ -155,9 +155,9 @@ pub fn render(
 
     let self_token = quote!(self);
     let (enum_repr, nr_expr) = if variants.is_empty() {
-        (quote!(), quote!(match *#self_token {}))
+        (quote!(), quote!(match #self_token {}))
     } else {
-        (quote!(#[repr(u8)]), quote!(*#self_token as u8))
+        (quote!(#[repr(u16)]), quote!(#self_token as u16))
     };
 
     if target == Target::Msp430 {
@@ -180,9 +180,9 @@ pub fn render(
                 #variants
             }
 
-            unsafe impl bare_metal::Nr for Interrupt {
+            unsafe impl cortex_m::interrupt::InterruptNumber for Interrupt {
                 #[inline(always)]
-                fn nr(&#self_token) -> u8 {
+                fn number(#self_token) -> u16 {
                     #nr_expr
                 }
             }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Don't generate pre Edition 2018 `extern crate` statements anymore
 
+- [breaking-change] Cortex-M PACs now rely on
+  `cortex_m::interrupt::InterruptNumber` instead of `bare_metal::Nr` for
+  interrupt number handling. The minimum supported `cortex-m` version is now
+  **0.7** and `bare-metal` is not a dependency anymore.
+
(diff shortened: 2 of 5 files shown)


adamgreig:
This LGTM to me, except I don't understand why we were using `cortex_m_rt::interrupt` before anyway.

therealprof:
I have no idea why we're not getting random errors in for different vendors. Need to investigate after cooking...

therealprof:
@adamgreig If we're happy with this we'll only need the cortex-m release to finalize this PR.

src/generate/device.rs
```diff
@@ -112,7 +112,7 @@ pub fn render(
         out.extend(quote! {
             pub use cortex_m::peripheral::Peripherals as CorePeripherals;
             #[cfg(feature = "rt")]
-            pub use cortex_m_rt::interrupt;
```
adamgreig, line 115: I think this was here to ensure the svd2rust generated PAC exports an `interrupt` macro, since the cortex-m-rt documentation says you should use `interrupt` from your PAC and not from c-m-rt. So, I think we should keep the `pub use cortex_m_rt::interrupt`, but we don't need to `pub use cortex_m::interrupt::InterruptNumber` at all, since the impl writes the full path anyway.

adamgreig (approved):
Looks good. I think we'll have to try it out a bit with cortex-m and PACs to be sure it's working as we expect but PR looks fine to me. Could you add a CHANGELOG entry?

therealprof:
@adamgreig I've updated the PR to use the release and added a CHANGELOG. I also ran `svd2rust-regress` locally but I haven't tried adapting a HAL impl yet.

adamgreig (approved):
bors merge

src/generate/interrupt.rs
```diff
@@ -180,9 +180,9 @@ pub fn render(
                 #variants
             }
 
-            unsafe impl bare_metal::Nr for Interrupt {
+            unsafe impl cortex_m::interrupt::InterruptNumber for Interrupt {
```
Rahix, line 183: This change breaks the build for [`avr-device`](https://github.com/Rahix/avr-device).  The problem is that you modified the `else` branch which is for everything but `msp430` ...

I suppose this code needs to be refactored such that this line is only emitted for `cortex-m`.  I would assume other platforms (like `riscv`) will also run into the same issue.

Cc: @trembel

therealprof:
@Rahix Thanks for noticing. Indeed we has some discussion in #455 about other architectures and the approach to be taken there.

Since this doesn't break RISC-V in CI, this would imply that either it doesn't break or that it is not exercised properly.

AVR is completely disabled in CI which is why there's no indication of breakage: https://github.com/rust-embedded/svd2rust/blob/c68b71c09c87ff17270e9e8e6cb63f45276b5068/ci/script.sh#L78

Rahix:
> Since this doesn't break RISC-V in CI, this would imply that either it doesn't break or that it is not exercised properly.

Hmm, from reading the code in `interrupt.rs` I cannot tell how RISC-V manages to evade this problem (unless all tested SVD files contain no interrupts at all which I would assume is not the case).  Would need another closer look ...

> AVR is not even tested in CI which would be a good first thing to add.

I can look into adding that.  For AVR, the vendor files are in a format called ATDF which we first convert to SVD using a small utility called [`atdf2vd`](https://github.com/Rahix/atdf2svd).  I suppose I should upload a pre-generated SVD somewhere and reference that for the CI tests?

PR #477, opened 2020-09-23
fix W::bits signature

### LINKED ISSUE #475: Breaking change in W::bits method
I changed the `W::bits` method signature in  #465 at https://github.com/rust-embedded/svd2rust/commit/d9e06a747942acc8ea5681911304230d1fd56b27#diff-b0b528563af56ba9f8af6cb6bb3d2539L213.  It really looks like an oversight, but offhand I can't recall if there was a specific reason to do it.  I didn't call it out in the PR, which makes me think it was a mistake.

In any case, I'd better fix it.

couchand:
Fixes #475.

### DIFF
--- a/src/generate/generic.rs
+++ b/src/generate/generic.rs
@@ -102,7 +102,7 @@ impl<REG: Resettable + Writable> Reg<REG> {
     #[inline(always)]
     pub fn write<F>(&self, f: F)
     where
-        F: FnOnce(&mut REG::Writer) -> &mut REG::Writer,
+        F: FnOnce(&mut REG::Writer) -> &mut W<REG>
     {
         self.register.set(
             f(&mut REG::Writer::from(W {
@@ -124,7 +124,7 @@ where
     #[inline(always)]
     pub fn write_with_zero<F>(&self, f: F)
     where
-        F: FnOnce(&mut REG::Writer) -> &mut REG::Writer,
+        F: FnOnce(&mut REG::Writer) -> &mut W<REG>
     {
         self.register.set(
             (*f(&mut REG::Writer::from(W {
@@ -157,7 +157,7 @@ impl<REG: Readable + Writable> Reg<REG> {
     #[inline(always)]
     pub fn modify<F>(&self, f: F)
     where
-        for<'w> F: FnOnce(&REG::Reader, &'w mut REG::Writer) -> &'w mut REG::Writer,
+        for<'w> F: FnOnce(&REG::Reader, &'w mut REG::Writer) -> &'w mut W<REG>
     {
         let bits = self.register.get();
         self.register.set(
@@ -216,8 +216,9 @@ pub struct W<REG: RegisterSpec + ?Sized> {
 impl<REG: RegisterSpec> W<REG> {
     /// Writes raw bits to the register.
     #[inline(always)]
-    pub unsafe fn bits(&mut self, bits: REG::Ux) {
+    pub unsafe fn bits(&mut self, bits: REG::Ux) -> &mut Self {
         self.bits = bits;
+        self
     }
 }
 


burrbull:
This code doesn't fix problem in my view, as REG:Writer and W<REG> are different types.
https://github.com/rust-embedded/svd2rust/blob/1a2bf65b8f08c0a9a6f3c276e393c511ed236cf9/src/generate/register.rs#L84

burrbull:
I don't see any way (without traits) to implement generic `bits` function that will be work as before.

If we don't found better solution, I think it is better just delete generic `unsafe fn bits`.
And replace
https://github.com/rust-embedded/svd2rust/blob/29129c0e4e261bf4c38057f7585e7ddfc05fafd0/src/generate/register.rs#L168
with:
`self.bits = bits`

couchand:
I'd agree that it is a bit surprising at first that this works.  I submitted the PR and went off to bed, but then had one of those moments lying in bed and thinking about it when I felt a sinking feeling that the fix would break everything.  I got back up and did a bunch more testing to verify that nothing breaks.

The key insight is that there's a deref coercion happening.  Since `REG::Writer` implements `DerefMut<Target=W>`, a mutable reference to the more specific type is just as good as a mutable reference to the more general type.  However, that's not true the other direction (we have to explicitly `.into()` for that to work).  This is why we need to relax the bounds at [1], [2], and [3].

[1]: https://github.com/rust-embedded/svd2rust/pull/477/files#diff-b0b528563af56ba9f8af6cb6bb3d2539L105
[2]: https://github.com/rust-embedded/svd2rust/pull/477/files#diff-b0b528563af56ba9f8af6cb6bb3d2539L127
[3]: https://github.com/rust-embedded/svd2rust/pull/477/files#diff-b0b528563af56ba9f8af6cb6bb3d2539L160

A collection of resources I assembled while verifying that this was correct:
* [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=82a9e748108170b875b409f464e48224)
* [`DerefMut` trait ref](https://doc.rust-lang.org/std/ops/trait.DerefMut.html#more-on-deref-coercion) (specifically: *Values of type &mut T are coerced to values of type &mut U*)
* [the book Chapter 15.2](https://doc.rust-lang.org/book/ch15-02-deref.html)

couchand:
(As an aside, all this chicanery would be avoided with #478).

burrbull:
See what problem I see:

```rust
    reg.write(|w| { w
         .bits(10)    // <-- DerefMut here
         .field1().set_bit()  // <-- error! generic W<REG> doesn't have `field1`
    });
```

couchand:
Ah, that's a fair point.  Though it would be a fairly marginal use-case, wouldn't it?  Setting the bits for the whole register explicitly and then overriding just one field?

couchand:
Oh, thanks @burrbull for bringing up the specialized version of `bits` at https://github.com/rust-embedded/svd2rust/blob/29129c0e4e261bf4c38057f7585e7ddfc05fafd0/src/generate/register.rs#L167-L170, in all the hubbub I had forgotten about it.  That would certainly explain why in any normal code #475 didn't manifest as a bug, since the compiler will pick the impl directly on the specific type before even trying to deref.

So #475 will only manifest in code written to be generic over `Writable`, so I'm almost certainly the only one affected.

To close the loop, your example at https://github.com/rust-embedded/svd2rust/pull/477#issuecomment-697427710 will be fine, since `w` refers to the specific register writer, so `w.bits` refers to the version that returns `self`, not the generic version implemented here.

burrbull:
> Is that example from real code somewhere?

I already saw something like this.

On the other hand this should work:
```rust
reg.write(|w| {
      w.bits(10);    // <-- DerefMut here, but drop ref
      w.field1().set_bit()
});
```
Can you test this?

couchand:
> I already saw something like this.

Do you mean just that you've seen that pattern before or that it exhibited a bug?  My current thinking is that there's no bug in that pattern, since the `w.bits` that gets called there is not the one added by this PR.

I've not been able to trigger #475 on master or any suspected bug from this PR in any testing (and it does seem like it would have to be both or neither, not one or the other).  If you have an example of somewhere that has an issue, please point me to it.

couchand:
Ping @burrbull.  I don't believe there's a bug in any code except my own.  Happy to look at any counter-example that you might come up with.

Are there any other outstanding issues with this PR?  Thanks!

couchand:
@burrbull, have you been able to identify any issues with this change?  To reiterate, I'm pretty confident that the only one affected by the bug is me.

@ryankurte it looks like you're still the @rust-highfive reviewer here, would you by chance be able to take a look at this PR?  Thanks!

burrbull (approved):

ryankurte (approved):
thanks for the fix! i am not 100% across the related issue but, looks okay to me.

bors r+

couchand:
Thanks for the reviews!

PR #483, opened 2020-12-02
Only use cortex-m when target is cortex-m. Fixes #482.

### LINKED ISSUE #482: Support for platform=none broken
In https://github.com/rust-embedded/svd2rust/pull/473#pullrequestreview-539243757 we inadvertently added a dependency on cortex-m for all platforms except msp430, which has broken svd2rust use for AVR platforms (and possibly others).

Hopefully this specific issue can be easily fixed by updating that `else` branch to target only cortex-m. We could consider adding tested support for AVR devices separately too.

adamgreig:
There might be more we can do here but this simple fix should at least stop us using `cortex-m` for targets other than `cortex-m`.

cc @Rahix

### DIFF
--- a/src/generate/interrupt.rs
+++ b/src/generate/interrupt.rs
@@ -173,23 +173,25 @@ pub fn render(
         root.extend(interrupt_enum);
     } else {
         let interrupt_enum = quote! {
-            ///Enumeration of all the interrupts
+            ///Enumeration of all the interrupts.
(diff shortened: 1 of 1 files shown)


Rahix (approved):
This makes the AVR build work again, thanks!

I agree, the large amount of nested branching in this function should probably be refactored at some point in the future ...

therealprof (approved):
LGTM, thanks for picking this up @adamgreig.

bors r+

PR #484, opened 2020-12-02
Update the min `vcell` version to v0.1.2

david-sawatzke:
`src/generate/generic.rs:53` contains an `as_ptr()` call on a vcell.
This was only added in `vcell` v0.1.1 (since yanked).

Thus, the generated crates only build with vcell v0.1.2

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -70,7 +70,7 @@ main() {
     cargo init --name foo $td
     echo 'cortex-m = "0.7.0"' >> $td/Cargo.toml
     echo 'cortex-m-rt = "0.6.13"' >> $td/Cargo.toml
-    echo 'vcell = "0.1.0"' >> $td/Cargo.toml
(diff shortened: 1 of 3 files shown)


therealprof:
Hm, a specification of "0.1.0" should automatically pick up "0.1.2", so this should not cause a build problem unless there's an outdated `Cargo.lock` in your project.

therealprof (approved):
LGTM, thanks.

bors r+

david-sawatzke:
> Hm, a specification of "0.1.0" should automatically pick up "0.1.2", so this should not cause a build problem unless there's an outdated Cargo.lock in your project.

(you already merged this, but I wanna get this in here) That works, but it should build with all allowed combinations, not just the most obvious ones. This could (but pretty unlikely) also lead to problems if some other dependency pins it's vcell version with "=0.1.0".

PR #485, opened 2020-12-04
Expand derived clusters

### LINKED ISSUE #462: Unexpected token error processing SVD file
[RA6M2.zip](https://github.com/rust-embedded/svd2rust/files/5033887/RA6M2.zip)

I am trying to process the RA6M2.svd file.

This requires changes to the svd-parser that I have already reported and to register.rs file, which I have attached, to filter out enumerated values without a value element.

When I run svd2rust with these modification I get a [ERROR svd2rust] unexpected token

smindinvern:
This PR addresses two issues encountered when attempting to process the svd file [here](https://raw.githubusercontent.com/renesas/fsp/master/ra/fsp/src/bsp/cmsis/Device/RENESAS/SVD/RA.svd).

1.  The SVD file defines a cluster in the 'R_GPT_ODC' peripheral that derives from a sibling cluster.  AFAICT from the SVD schema, this is allowed, and we should expand the cluster definition in the same way registers are handled.
2.  The SVD file contains many `EnumeratedValue`s that are "reserved", but aren't filtered out because they're named "other".  They have an `<isDefault>` child element, which is mutually exclusive with the `<value>` child element according to the SVD schema, i.e. they don't represent a value that we would want to define in the generated code, and so can be skipped.

When running `svd2rust-regress` locally 3 tests fail, but the same tests also fail on `master`, so it seems there's no indication of a regression here.

This seems to fix #462.

### DIFF
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -87,59 +87,78 @@ pub fn render(
 
     // erc: *E*ither *R*egister or *C*luster
     let ercs = p.registers.as_ref().map(|x| x.as_ref()).unwrap_or(&[][..]);
-    let registers: &[&Register] = &util::only_registers(&ercs)[..];
 
-    // make a pass to expand derived registers.  Ideally, for the most minimal
+    // make a pass to expand derived registers and clusters.  Ideally, for the most minimal
     // code size, we'd do some analysis to figure out if we can 100% reuse the
     // code that we're deriving from.  For the sake of proving the concept, we're
     // just going to emit a second copy of the accessor code.  It'll probably
     // get inlined by the compiler anyway, right? :-)
 
     // Build a map so that we can look up registers within this peripheral
-    let mut reg_map = HashMap::new();
-    for r in registers {
-        reg_map.insert(&r.name, svd_parser::svd::register::Register::clone(r));
+    let mut erc_map = HashMap::new();
+    for erc in ercs {
+        erc_map.insert(util::erc_name(erc), erc.clone());
     }
 
-    // Build up an alternate erc list by expanding any derived registers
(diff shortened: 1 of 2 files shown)


burrbull:
LGTM.
Rustfmt, please.

smindinvern:
I noticed that this causes invalid rust code to be generated for fields which only have a single `isDefault` value.  I have a patch for that, but I'm not sure that it's correct.  Should I include that here for consideration, or address that separately?

burrbull:
I think it's better to split this PR. And place c9b18c7165523c610efb84bd2eed61658dbe05eb in other one.

src/generate/register.rs
```diff
@@ -771,9 +771,9 @@ impl Variant {
         let span = Span::call_site();
         evs.values
             .iter()
-            // filter out all reserved variants, as we should not
+            // filter out all reserved or default variants, as we should not
             // generate code for them
-            .filter(|field| field.name.to_lowercase() != "reserved")
+            .filter(|field| field.name.to_lowercase() != "reserved" && field.is_default == None)
```
Emilgardis, line 776: This will fail in some cases as noted. A suggestion is to run svd2rust-regress and then diff the outputs, that can help with determining validity of change.

Emilgardis (changes requested):
Nice!

Rebase to squash rustfmt and remove c9b18c7165523c610efb84bd2eed61658dbe05eb

burrbull (approved):

burrbull:
@Emilgardis

Emilgardis (approved):

Emilgardis:
bors r=burrbull

therealprof:
@smindinvern And of course I missed that this is missing a CHANGELOG entry. Would you mind PRing one? 😅

PR #486, opened 2020-12-07
Fix handling of `isDefault' `enumeratedValue's

smindinvern:
Currently, an `enumeratedValue` is ignored if (and only if) it's name matches "reserved", and all remaining `enumeratedValue`s are expected to have a `value` child element.  However, the SVD schema allows an `enumeratedValue` to contain *either* a `value` element or an `isDefault` element.  When `svd2rust` hits an `isDefault` element that isn't named "reserved", an error is thrown.

In general, filtering these out should be safe, as they don't represent specific values that can be assigned to a field, but rather a *range* of values.  The ability to read and write such arbitrary values is already handled with the `bits()` method defined on field readers/writers.  The one corner case, however, is for fields with only a *single* `enumeratedValue` which has an `isDefault` child element.  Attempting to handle these in the usual way results in emitting empty `enum`s which causes compilation failures.  Instead, emit a `newtype` encapsulating the field's base type.

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -442,79 +442,82 @@ pub fn fields(
                 } else {
                     let has_reserved_variant = evs.values.len() != (1 << width);
                     let variants = Variant::from_enumerated_values(evs)?;
-
-                    add_from_variants(mod_items, &variants, &name_pc_a, &fty, &description, rv);
-
                     let mut enum_items = TokenStream::new();
 
-                    let mut arms = TokenStream::new();
-                    for v in variants.iter().map(|v| {
-                        let i = util::unsuffixed_or_bool(v.value, width);
-                        let pc = &v.pc;
+                    if variants.is_empty() {
+                        add_with_no_variants(mod_items, &name_pc_a, &fty, &description, rv);
+                    } else {
+                        add_from_variants(mod_items, &variants, &name_pc_a, &fty, &description, rv);
+
+                        let mut arms = TokenStream::new();
+                        for v in variants.iter().map(|v| {
+                            let i = util::unsuffixed_or_bool(v.value, width);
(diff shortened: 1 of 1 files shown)


Emilgardis:
Comes initially from https://github.com/rust-embedded/svd2rust/pull/485

therealprof:
Is there a specific svd file that exhibits this conflict?

smindinvern:
@therealprof: it was mentioned in #485, but specifically it happens with [this file](https://raw.githubusercontent.com/renesas/fsp/master/ra/fsp/src/bsp/cmsis/Device/RENESAS/SVD/RA.svd).  One example is the `FSA` field in `R_FACI_HP.FSADDR`:

```xml
<register>
  <name>FSADDR</name>
  <description>Flash Start Address</description>
  <addressOffset>0x0030</addressOffset>
  <size>32</size>
  <access>read-write</access>
  <resetValue>0x00000000</resetValue>
  <resetMask>0xFFFFFFFF</resetMask>
  <fields>
    <field>
      <name>FSA</name>
      <description>Start Address of Flash Sequencer Command Target Area
  These bits can be written when FRDY bit of FSTATR register is "1". Writing to these bits in FRDY = "0" is ignored.</description>
      <lsb>0</lsb>
      <msb>31</msb>
      <access>read-write</access>
      <enumeratedValues>
        <enumeratedValue>
          <name>others</name>
          <description>Specifies start address for each command processing.</description>
          <isDefault>true</isDefault>
        </enumeratedValue>
      </enumeratedValues>
    </field>
  </fields>
</register>
```

smindinvern:
Anything I could do to make this easier/more ready to review?

therealprof (approved):
Sorry for the delay, LGTM. Thanks a lot.

bors r+

PR #489, opened 2021-02-02
Fix type path for cluster fields. Allows nested clusters

### LINKED ISSUE #249: error: cannot find type `MODE2_ALARM` in this scope
The library generated from a SAM L11 svd fails with the following error:
```
error[E0412]: cannot find type `MODE2_ALARM` in this scope
   --> src/rtc/mod.rs:125:23
    |
125 |     pub mode2_alarm0: MODE2_ALARM,
    |                       ^^^^^^^^^^^ not found in this scope
help: possible candidate is found in another module, you can import it into scope
    |
1   | use rtc::mode2::MODE2_ALARM;
    |
```

Somehow this field in the MODE2 register block fails to reference the `mode2` module like all the others do: 
```
    #[doc = "0x18 - MODE2 Clock Value"]
    pub clock: self::mode2::CLOCK,
    _reserved2: [u8; 4usize],
    #[doc = "0x20 - MODE2 Alarm"]
    pub mode2_alarm0: MODE2_ALARM,
    _reserved3: [u8; 27usize],
    #[doc = "0x40 - General Purpose"]
    pub gp: [self::mode2::GP; 2],
    _reserved4: [u8; 24usize],
```

I noticed that in the svd this seems to be defined as a cluster, rather than a register. Does this help explain the issue?

I am able to workaround through patches.

You can find the svd I used here - https://github.com/evq/atsaml11xxx/blob/master/ATSAML11E16A.orig.svd

theli-ua:
Fixes #249

Tested on following test svd: https://gist.github.com/theli-ua/a1b9c17e94442fbd56c774bb504795a0

### DIFF
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -633,7 +633,7 @@ fn expand_cluster(
 
     match cluster {
         Cluster::Single(info) => cluster_expanded.push(RegisterBlockField {
-            field: convert_svd_cluster(cluster)?,
+            field: convert_svd_cluster(cluster, name)?,
             description: info.description.as_ref().unwrap_or(&info.name).into(),
             offset: info.address_offset,
             size: cluster_size,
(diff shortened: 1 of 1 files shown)


Emilgardis (approved):
bors r+

PR #491, opened 2021-02-05
Fix type paths for arrays of nested clusters

theli-ua:
This is a followup to https://github.com/rust-embedded/svd2rust/pull/489 , this time for arrays


burrbull (approved):

theli-ua:
@burrbull bors seem to still pend on approval, do you need to poke it?

PR #495, opened 2021-03-27
use u8 for field array offset instead of usize

burrbull:


burrbull:
cc @adamgreig

adamgreig:
Is the idea that field arrays can never have >255 elements because registers are only 32 (or maybe 64) bits wide, so there's no point using a large usize when a u8 will always work?

therealprof:
I'd also like to see what the concrete benefits are. Switching from `usize` to something else might require a bit of casting during use which is annoying (and error prone).

burrbull:
> Is the idea that field arrays can never have >255 elements because registers are only 32 (or maybe 64) bits wide, so there's no point using a large usize when a u8 will always work?

Yes, I forget this when implemented. As field arrays are not really arrays, I don't think someone will use big usize accessing them.

adamgreig:
> I'd also like to see what the concrete benefits are. Switching from `usize` to something else might require a bit of casting during use which is annoying (and error prone).

It's only used in bit shifting arithmetic, not as a slice index; this PR changes the generated code from something like:

```rust
            pub unsafe fn pin(&self, n: usize) -> PIN_R {
                PIN_R::new(((self.bits >> n) & 0x01) != 0)
            }
```

to something like:

```rust
            pub unsafe fn pin(&self, n: u8) -> PIN_R {
                PIN_R::new(((self.bits >> n) & 0x01) != 0)
            }
```

where in this case `self.bits` is 32-bit (or in general is the register width).

@burrbull, do any of the tests in the CI actually use field arrays?

therealprof:
> It's only used in bit shifting arithmetic, not as a slice index; this PR changes the generated code from something like:

Fair enough, but in that case it's also unlikely to provide any benefit.

adamgreig:
We haven't actually shipped an svd2rust since field array support was added in #400, so I see this as more cleaning up the API for it before release. But, yes, I don't know if there's any benefit to swapping to u8 for this. If anything maybe it should use the same type as the registers to make it line up with self.pins etc?

burrbull:
> do any of the tests in the CI actually use field arrays?

risk-v k210 should use them, but I'm not sure is test contain actual version of SVD.

PR #496, opened 2021-03-27
Remove useless `Variant`, use Option instead

burrbull:
As each  `FieldReader` now has `bits` method, I think it is not an issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


burrbull:
See #496

therealprof (dismissed):
Using `Option` makes a lot of sense and lets us profit from upstream optimisations and methods which is a good thing. Would you mind removing the `u8` change from this PR so we can get that merged?

burrbull:
> Using `Option` makes a lot of sense and lets us profit from upstream optimisations and methods which is a good thing. Would you mind removing the `u8` change from this PR so we can get that merged?

Done.

adamgreig (approved):
bors merge

adamgreig:
bors merge

PR #500, opened 2021-04-13
Add CHANGELOG.md entry checker to GHA workflows

therealprof:
r? @adamgreig

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,19 @@
+on:
+  pull_request:
+
+name: Changelog check
+
+jobs:
+  changelog:
+    name: Changelog check
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Changelog updated
+        uses: Zomzog/changelog-checker@v1.1.0
+        with:
+          fileName: CHANGELOG.md
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,19 @@
+on:
+  pull_request_target:
```
adamgreig, line 2: Does it need to be `pull_request_target`? On the example for the action they seem to just use a normal `pull_request`.

therealprof: Not sure, it's a verbatim copy from the `stm32f4xx-hal`. 🤷🏻‍♂️

adamgreig: Could you try with just pull_request? If that works it's somewhat more straightforward and safer.

therealprof, line 2:
```suggestion
  pull_request:
```

adamgreig:
Thanks, looks like it works. Currently it won't stop bors at all (if I bors+ this now, it should merge, I think) because it doesn't run on bors' push to staging and also bors isn't configured to care about it. I think this sort of makes sense - it should alert reviewers/users that the changelog is missing, but doesn't stop a change being r+'d if the reviewer thinks it's fine. I don't think the "apply label to allow it to pass without changelog" would work under bors anyway, so this is probably the best config. What do you think? Happy to merge as-is.

adamgreig (dismissed):

adamgreig:
Oh, sorry, I didn't notice the force push. What did you change?

adamgreig (approved):
bors merge

therealprof:
As requested I changed `pull_request_target` to `pull_request`.

therealprof:
> I don't think the "apply label to allow it to pass without changelog" would work under bors anyway, so this is probably the best config. What do you think? Happy to merge as-is.

I agree. Alerting the reviewer to a missing CHANGELOG entry should be okay.

PR #501, opened 2021-04-14
don't add underscore to end of "is_keyword"

### LINKED ISSUE #497: If reserved name is altered and becomes non reserved, do not add underscore
Reference issue: <https://github.com/stm32-rs/stm32-rs/issues/523>

Reserved names, like `match` cannot be used as function name, for this reason, an underscore is added at the end, but when it is a boolean for example, it ends up as `is_match_`, but in this case, it could be left as `is_match`.

burrbull:
fixes #497

not tested yet

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
-- Keyword sanitizing (`async`)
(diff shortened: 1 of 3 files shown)


burrbull:
should work correct now
cc @therealprof

therealprof (approved):
LGTM, thanks!

bors r+

PR #502, opened 2021-04-14
unsafe write_with_zero

burrbull:
As 0 is not always correct value, `write_with_zero` can't be safe for all registers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- [breaking-change] make `write_with_zero` method `unsafe` because the way it is
+
 - [breaking-change] remove `Variant<U, ENUM_A>`, use `Option<ENUM_A>` instead
 
 - split out register size type (`RawType`) from `ResetValue` trait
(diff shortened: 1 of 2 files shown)


adamgreig:
I agree, this seems a shame but it fits our current safety model. Please could you add a note to the changelog?

therealprof:
> I agree, this seems a shame but it fits our current safety model. Please could you add a note to the changelog?

How about merging #500? 😉

adamgreig:
@burrbull maybe we could leave it safe if the reset value is 0 and not otherwise? or also if all fields are covered, the same way the register-level `bits()` becomes safe?

burrbull:
> @burrbull maybe we could leave it safe if the reset value is 0 and not otherwise? or also if all fields are covered, the same way the register-level `bits()` becomes safe?

In this way we need to implement own `write_with_zero` on each register.

therealprof:
I think this is fine and makes sense.

@adamgreig Any objections to merging it?

adamgreig:
> In this way we need to implement own `write_with_zero` on each register.

Oh, yea, you're right of course. Never mind.

@therealprof yep let's merge.

therealprof (approved):
Alright then.

bors r+

PR #503, opened 2021-04-14
implement both owning and const generic field array writers

burrbull:


burrbull:
Now we can see something like this:
image

.github/workflows/ci.yml
```diff
@@ -25,7 +25,7 @@ jobs:
 
         include:
           # Test MSRV
-          - rust: 1.40.0
+          - rust: 1.51.0
```
therealprof, line 28: For me requiring 1.51 at this point in time is not ideal. Other opinions, @adamgreig maybe?

adamgreig: Is this to generate crates or to use the generated crates? 1.51 is _way_ too new for building PACs, but I wouldn't mind requiring it to generate them.

therealprof: Generated crates. It's providing a const generic writer. I would not terribly object if this was hidden behind a feature gate but I fully agree that enabled for all is a non-starter.

burrbull: > behind a feature gate

It's idea

burrbull: Done

burrbull: Now this PR doesn't change anything if you don't use `const-generic` feature.

burrbull:
rebased

therealprof:
@adamgreig Any objections to the approach?

therealprof (changes requested):
We should still test this in CI with 1.51.

burrbull:
@therealprof 
I've added test with rust 1.51 and enabled `strict` and `const-generic` features. RISC-V `k210` uses field arrays

adamgreig (approved):
LGTM

therealprof (approved):
Thanks a lot!

burrbull:
rebased

adamgreig (approved):
bors merge

Issue #504, opened 2021-04-15
Register definitions with the same <name> but different <alternateGroup> generates duplicate structs.


john-terrell:
I'm unsure I this is a dupe of #357.   The SVD [here](https://github.com/atsam-rs/atsam-pac/blob/john/sam4l_support/svd/ATSAM4LC2A.svd) creates duplicate structs.

error[E0428]: the name `CMR` is defined multiple times
   --> pac/atsam4lc2a/src/tc0.rs:151:1
    |
147 | pub type CMR = crate::Reg<cmr::CMR_SPEC>;
    | ----------------------------------------- previous definition of the type `CMR` here
...
151 | pub type CMR = crate::Reg<cmr::CMR_SPEC>;
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `CMR` redefined here
    |
    = note: `CMR` must be defined only once in the type namespace of this module

And the code does indeed contain duplicates:

#[doc = "Channel Control Register Channel"]
pub mod ccr;
**#[doc = "CMR register accessor: an alias for `Reg<CMR_SPEC>`"]
pub type CMR = crate::Reg<cmr::CMR_SPEC>;
#[doc = "Channel Mode Register Channel"]
pub mod cmr;**
**#[doc = "CMR register accessor: an alias for `Reg<CMR_SPEC>`"]
pub type CMR = crate::Reg<cmr::CMR_SPEC>;
#[doc = "Channel Mode Register Channel"]
pub mod cmr;**
#[doc = "CV register accessor: an alias for `Reg<CV_SPEC>`"]
pub type CV = crate::Reg<cv::CV_SPEC>;

burrbull:
Can you try different versions of `svd2rust` an investigate what version (commit) causes this?

burrbull:
I see such output when run:
```sh
[WARN  svd2rust::generate::peripheral] Found type name conflict with region Some("cmr"), renamed to Some("cmr0")
[WARN  svd2rust::generate::peripheral] Found type name conflict with region Some("cmr"), renamed to Some("cmr1")
[WARN  svd2rust::generate::peripheral] Found type name conflict with region Some("cmr"), renamed to Some("cmr2")
```

john-terrell:
I tried master and the head of the v0.18 branch.

burrbull:
CMR is twice described. Check your SVD more carefully. 
[...]

PR #506, opened 2021-04-15
support for alternateGroup registers

burrbull:
See #504

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


therealprof (commented):
Looks good to me in general. Where is this needed?

burrbull:
> Looks good to me in general. Where is this needed?

Some vendors (like Atmel, or Freescale) use `alternateGroup` to specify registers with that can use different set of fields in different modes.
I've just added this group name as prefix for register name. So now there are no conflicts.

burrbull:
cc @adamgreig  @john-terrell

therealprof (approved):
LGTM

bors r+

PR #507, opened 2021-04-15
Emit code to bring generic module into scope when using `-g` option.

cr1901:
The -g option emits a separate file called generic.rs to split the generic code from the peripherals.

Right now the -g option requires a user to manually add something like the following lines to `lib.rs`:

```rust
use generic::*;
pub mod generic;
```

Otherwise compilation will fail with errors like the following:

```sh
     error[E0433]: failed to resolve: maybe a missing crate `FieldReader`?
   --> src\port_1_2\p2ifg.rs:111:25
    |
111 |         P2IFG2_R(crate::FieldReader::new(bits))
    |                         ^^^^^^^^^^^ maybe a missing crate `FieldReader`?
```

This patch automates adding the above code.

### DIFF
--- a/src/generate/device.rs
+++ b/src/generate/device.rs
@@ -146,6 +146,13 @@ pub fn render(
     let generic_file = std::str::from_utf8(include_bytes!("generic.rs"))?;
     if generic_mod {
         writeln!(File::create("generic.rs")?, "{}", generic_file)?;
+
+        out.extend(quote! {
+            #[allow(unused_imports)]
+            use generic::*;
+            ///Common register and bit access and modify traits
+            pub mod generic;
+        });
     } else {
         let tokens = syn::parse_file(generic_file)?.into_token_stream();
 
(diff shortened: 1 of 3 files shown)


src/generate/device.rs
```diff
@@ -146,6 +146,13 @@ pub fn render(
     let generic_file = std::str::from_utf8(include_bytes!("generic.rs"))?;
     if generic_mod {
         writeln!(File::create("generic.rs")?, "{}", generic_file)?;
+
+        out.extend(quote! {
+            #[allow(unused_imports)]
+            use generic::*;
+            ///Common register and bit access and modify traits
```
adamgreig, line 153: Annoyingly plain comments like this don't make it through to the generated code, so to keep it in the output we need to use the `doc` attribute instead.

```suggestion
            #[doc="Common register and bit access and modify traits"]
```

adamgreig (approved):
Thanks!

bors merge

PR #510, opened 2021-04-17
Add -m switch to create a module instead of a lib

adamgreig:
Currently projects like stm32-rs have to strip out all the attributes during build to create a file suitable for use as a module. This PR adds a new `-m` switch instead, which causes a `mod.rs` file to be created without the inner attributes or the `pub mod generic`/`use generic::*` lines which would live in the top-level `lib.rs`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- New `-m` switch generates a `mod.rs` file instead of `lib.rs`, which can
+  be used as a module inside a crate without further modification.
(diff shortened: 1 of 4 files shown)


burrbull (approved):

burrbull:
Have you tested all 4 modes?
-m and -g  disables/enabled all/only one

adamgreig:
I guess ideally we should add a CI test for both -g and -m (and combinations) but it didn't seem too easy to add to the current setup.

svd2rust works OK with nothing, `-m`, `-g`, and `-m -g`, and stm32-rs builds a good PAC with `-m` or `-m -g`.

PR #512, opened 2021-04-30
register iterator

burrbull:


src/generate/register.rs
```diff
@@ -1101,10 +1101,10 @@ fn lookup_in_peripheral<'p>(
     base_register: &'p str,
     base_field: &str,
     base_evs: &str,
-    all_registers: &[&'p Register],
+    mut all_registers: impl Iterator<Item = &'p Register>,
```
therealprof, line 1104: Hm, why is this `mut` now?

burrbull: Because iteration is mutable one-off operation.
If it is harder to read, I can bring vector and slice back.

therealprof: Hm, don't we have a concrete borrowable type for all Registers to pass in? I would expect that can build something like a `RegisterMap` and pass a reference to it around allowing called functions to obtain their own `iter` to it.

burrbull: I've restored previous behavior.

therealprof (approved):
LGTM

bors r+

PR #518, opened 2021-05-09
output-dir

burrbull:


CHANGELOG.md
```diff
@@ -7,10 +7,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- Option `-o`(`--output-path`) let you specify output directory path
+- Option `ignore_groups` for optional disabling #506
```
Emilgardis, line 13: This should describe what it does, instead of simply linking to an issue

CHANGELOG.md
```diff
@@ -7,10 +7,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+
 ### Changed
 
 - options can be set now with `svd2rust.toml` config
-- option `ignore_groups` for optional disabling #506
```
Emilgardis, line 13: ditto

src/lib.rs
```diff
@@ -516,7 +516,7 @@ pub fn generate(xml: &str, config: &Config) -> Result<Generation> {
     let device = svd::parse(xml)?;
     let mut device_x = String::new();
     let items =
-        generate::device::render(&device, &config, &mut device_x).or(Err(SvdError::Render))?;
+        generate::device::render(&device, &config, &mut device_x, ".").or(Err(SvdError::Render))?;
```
Emilgardis, line 519: shouldn't this be configurable for the lib as well?

Emilgardis (changes requested):

CHANGELOG.md
Emilgardis: (probably don't need two changelog entries)

src/lib.rs
burrbull: Added to config

Emilgardis (approved):
no concerns from me. bors r=me if there's nothing else

burrbull:
bors r=Emilgardis

PR #524, opened 2021-05-25
multiline descriptions

### LINKED ISSUE #523: Properly handle newlines in descriptions
The [RP2040 SVD](https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2040/hardware_regs/rp2040.svd) contains extensive inline documentation, but svd2rust escapes all newlines (doc strings in the output contain `\\n`), which messes up the generated documentation (see https://docs.rs/rp2040-pac/0.1.1/rp2040_pac/). An example from the spec shows the same usage of `\n` to "force" line breaks (https://siliconlabs.github.io/Gecko_SDK_Doc/CMSIS/SVD/html/svd__example_pg.html).

burrbull:
Fixes #523

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- `\n` in descriptions for multiline 
 - `_rererved` fields in `RegisterBlock` now hexidemical usize
 - options can be set now with `svd2rust.toml` config
 - option `ignore_groups` for optional disabling #506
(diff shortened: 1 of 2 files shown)


burrbull:
@9ary Do this works as you expected?

9ary:
Look like it works, yes. Output is still not great because most of the descriptions are not actually valid markdown so single newlines get eaten, but I'm not sure it is svd2rust's job to fix that. The multiline descriptions are formatted for C headers, to be read in a plaintext editor.

Emilgardis (approved):
bors r+

Issue #526, opened 2021-05-29
device.x generation for `--target riscv`


allexoll:
Currently, the riscv option [does not](https://github.com/rust-embedded/svd2rust/blob/e94fc7b795ac57002cd06244d4ed2d92e093af1f/src/generate/interrupt.rs#L128) produce a device.x linker file for weak interrupt symbols.

I think that it would benefit from that kind of file.

Riscv is a bit different than ARM in its handling of interrupt request. Instead of having the branch address in the vector table, riscv only has 1 "external" interrupt line. once we are in the external machine interrupt handler, having that vector table can ease the dispatching of the irq.

that would be helpful for both PLIC-implementing SoCs, and for custom interrupt managers. Since the symbols are weak, not using them wouldn't break linking either.

I'd be open to work on a PR for this purpose, but i wanted to have some input before, i'll link to this issue on the riscv-rt repo as well.

Remarks welcome

PR #527, opened 2021-05-29
add device.x  for riscv targets, and provides __EXTERNAL_INTERRUPTS

allexoll:
This PR is a proposal to address #526 . 

`__EXTERNAL_INTERRUPTS` is the symbol used since `__INTERRUPTS` is used by [riscv-rt](https://github.com/rust-embedded/riscv-rt/blob/47ece5f5163a2e38ce5e4685b5d3145713d7954a/src/lib.rs#L505)

I think it fits because the peripheral interrupts are supposed to be pending through the external interrupt. 

This provides hals with information to manage interrupt request. either through a PLIC or trough the vectored interrupt when that is implemented (in [riscv-rt](https://github.com/rust-embedded/riscv-rt/issues/1) i suppose).

Remarks welcome

### DIFF
--- a/src/generate/interrupt.rs
+++ b/src/generate/interrupt.rs
@@ -125,7 +125,35 @@ pub fn render(
                     ];
             });
         }
-        Target::RISCV => {}
+        Target::RISCV => {
+            for name in &names {
+                writeln!(
+                    device_x,
+                    "PROVIDE({} = DefaultExternalInterruptHandler);",
+                    name
+                )?;
+            }
+
+            root.extend(quote! {
+                #[cfg(feature = "rt")]
+                extern "C" {
+                    #(fn #names();)*
+                }
+
+                #[doc(hidden)]
+                pub union Vector {
+                    pub _handler: unsafe extern "C" fn(),
+                    pub _reserved: usize,
+                }
+
+                #[cfg(feature = "rt")]
+                #[doc(hidden)]
+                #[no_mangle]
+                pub static __EXTERNAL_INTERRUPTS: [Vector; #n] = [
+                    #elements
+                ];
+            });
+        }
         Target::XtensaLX => {
             for name in &names {
                 writeln!(device_x, "PROVIDE({} = DefaultHandler);", name)?;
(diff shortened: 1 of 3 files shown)


allexoll:
( i'm unable to test with `svd2rust-regresison`, unsure if this is expected)

burrbull:
Looks good in first view. Add this to changelog, please.

cc @Disasm

burrbull (dismissed):

burrbull:
> ( i'm unable to test with `svd2rust-regresison`, unsure if this is expected)

We did not use `svd2rust-regression` a long time. So it can be broken.

burrbull:
move text in changelog under `Unreleased`

allexoll:
sorry, there you go

burrbull (approved):

src/generate/interrupt.rs
```diff
@@ -125,7 +125,35 @@ pub fn render(
                     ];
             });
         }
-        Target::RISCV => {}
+        Target::RISCV => {
+            for name in &names {
+                writeln!(
+                    device_x,
+                    "PROVIDE({} = DefaultExternalInterruptHandler);",
```
Disasm, line 132: Do we need a separate handler for this? I think we can reuse `DefaultInterruptHandler`.

allexoll: I just tested it and that works too. I'm not sure if it makes sense to have them point to the same handler.

Maybe the interrupt controller handler needs to do some housekeeping even if there is no handler declared for this interrupt line. 

The intention with this different symbol was to allow the hal to implement its own interrupt controller, since as of now, the IRQ management of riscv is not well fixed, where it is for NVIC/ARM.

For my soc pointing to the same handler is fine, but i don't know if that might be to restrictive for some implementation...

Disasm: I have no idea what kind of housekeeping can be done in this case. If there is a need for such, maybe it can be moved to the interrupt controller driver out of the handler functions.

allexoll: I'm fine switching it, we'll see if anyone needs that kind of functionality later

Disasm (approved):
Thanks!

burrbull (approved):

therealprof (approved):
Uhm, if you're happy with it  you might as well merge it. ;)

bors r+

PR #529, opened 2021-06-07
Allow multiple level of deriveFrom peripherals.

ithinuel:
Adds an alias to the base module to allow for multiple level of `deriveFrom`.


therealprof (dismissed):
Missing CHANGELOG.md entry.

burrbull (approved):

PR #534, opened 2021-07-19
Emit more Cluster arrays as arrays instead of lists of elements.

rcls:
(This obsoletes https://github.com/rust-embedded/svd2rust/pull/531)
Some Cluster arrays cannot be emitted as Rust (built-in) arrays due to padding between elements.
Add a helper zero-sized type `ArrayProxy` that overloads `[]` to do pointer-arithmetic internally.

I believe, although I'm not 100% confident, that I'm not breaking Rust safety, as follows: There is no way for (safe) code to create an `ArrayProxy`. Code that produces a (reference to) an `ArrayProxy` has to use `unsafe` somewhere.

See https://github.com/rcls/svd2rust-example/commit/b8645fecb05b2a38b59fe29df9fcf643b7d4ac74 for any example of the difference this makes (or does not make):
- STM32L051 and LPC408x get no change.
- Cypress PSOC gets a significant difference.

This uses const generics, so the new functionality requires the `--const_generic` command line optoin.


burrbull:
>     * 
> 
> 
> One obvious issue with the current code: It uses const-generics, and does not build with the current minimum supported Rust version (1.40.0). Marking as draft until at least that is fixed.

`svd2rust` has `--const_generic` that could be uses in your situation also.

src/generate/array_proxy.rs
```diff
@@ -0,0 +1,46 @@
+
```
burrbull, line 1: Add
```
#![allow(unused)]
```

rcls: Isn't `allow(unused)` redundant on `pub` definitions? E.g., nothing in `generic.rs` currently has `allow(unused)`?
Or am I missing something?

rcls:
> `svd2rust` has `--const_generic` option that could be uses in your situation also.

Thanks. The changes updated to rely on that option.

burrbull (approved):

PR #537, opened 2021-08-20
Don't use a mask when not needed

thibautvdv:
When the mask has the same size as the parent register of a field, then
a mask is not required. This removes the clippy warnings/errors where
for example an operation of inverting a mask results in zero

```
error: this operation will always return zero. This is likely not the intended outcome
   |
61 |         self.w.bits = (self.w.bits & !0xffff_ffff) | (value as u32 & 0xffff_ffff);
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
```

or when the operation of using a mask is ineffective.

```
warning: the operation is ineffective. Consider reducing it to `value as u32`
   |
61 |         self.w.bits = (self.w.bits & !0xffff_ffff) | (value as u32 & 0xffff_ffff);
   |                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
```

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -302,6 +302,12 @@ pub fn fields(
         let fty = width.to_ty()?;
         let evs = &f.enumerated_values;
 
+        let use_mask = if let Some(size) = parent.size {
+            size != width
+        } else {
+            false
+        };
+
         let lookup_results = lookup(
             evs,
             fields,
@@ -374,9 +380,13 @@ pub fn fields(
                 quote! {
                     ((self.bits >> #offset) & #hexmask) #cast
                 }
-            } else {
+            } else if use_mask {
                 quote! {
                     (self.bits & #hexmask) #cast
+                }
+            } else {
+                quote! {
+                    self.bits #cast
                 }
             };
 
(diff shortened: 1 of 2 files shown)


src/generate/register.rs
```diff
@@ -679,12 +697,23 @@ pub fn fields(
...
+                            }
+                        }
+                    } else {
+                        quote! {
+                            ///Writes raw bits to the field
+                            #inline
+                            pub #unsafety fn #bits(self, value: #fty) -> &'a mut W {
+                                self.w.bits |= value as #rty;
+                                self.w
+                            }
```
Emilgardis, line 716: Could these two be joined, instead using one quote for the fn decl, and two quotes for the assignment in an if else

eg 

```rust
let op = if use_mask {
    quote!{ self.w.bits = (self.w.bits & !#hexmask) | (value as #rty & #hexmask); }
else {
    quote!{ self.w.bits |= value as #rty; }
}
....
```

thibautvdv: Totally!

burrbull (dismissed):

src/generate/register.rs
```diff
@@ -679,11 +693,17 @@ pub fn fields(
                         }
                     }
                 } else {
+                    let op = if use_mask {
+                        quote!{ self.w.bits = (self.w.bits & !#hexmask) | (value as #rty & #hexmask); }
+                    } else {
+                        quote!{ self.w.bits |= value as #rty; }
```
burrbull, line 699: Looks like just assign must be there:
```rust
quote!{ self.w.bits = value as #rty; }
```

thibautvdv: Ah, you're right!

burrbull:
Squash commits, please.

burrbull (approved):

Emilgardis (approved):
Thank you!

bors r+

Issue #540, opened 2021-10-01
Counterintuitive handling of W1C fields


BigPeteB:
The way W1C (write-1-to-clear) fields are handled is counterintuitive. Basically, they're not handled differently from normal fields, which makes them counterintuitive from a programmer's perspective.

Suppose you have a register with a mix of RW and RW1C fields, like so:

```
Offset | Field    | Access | Description
-------|----------|--------|-------------
7      | irq_stat | RW1C   | 1 = Interrupt has occurred. Write 1 to clear the interrupt.
...    | ...      | ...    | ...
1      | foo_en   | RW     | 0 = Foo is disabled. 1 = Foo is enabled.
0      | irq_en   | RW     | 0 = Interrupts are disabled. 1 = Interrupts are enabled.
```

Suppose a programmer wants to implement "Change this register to set `foo_en` to 1." They write this in the obvious way as `reg.modify(|_, w| w.foo_en().set_bit())`. svd2rust does the simple and 'obvious' thing: it reads the register, computes `val | 1 << 1`, and writes that value to the register.

However, this naïve API and implementation has led to the wrong result. If `irq_stat` contained 1 when it was read, we would write back a 1, which would clear that field! This is not what the programmer wanted!

To work with such a register, the programmer needs to ensure that _every time they `modify()` that register_, they handle the W1C register correctly. To implement "Change this register to set `foo_en` to 1," they actually need to do `reg.modify(|_, w| w.foo_en().set_bit().irq_stat().clear_bit())`. This is counterintuitive, since their intention was to _not_ change the value of `int_stat`. Even the name `clear_bit()` is misleading, because the effect on the register will be to _not_ clear that field. When the programmer wants to "clear" the field, they must use the `set_bit()` function.

(Note: all of this applies to W0S ("write-0-to-set") fields as well. When the programmer wants to "set" the field, they must use the `clear_bit()` function, etc.)

This is backwards, confusing, and dangerous. I'll grant that one could argue that svd2rust is simply implementing the obvious actions, and that the onus is on the programmer to know what the fields in the register are and how to use them correctly. However, given how svd2rust provides so much other protection such as not accessing reserved bits, or only writing allowed enumerated values, I'd argue that this naïve handling of W1C fields is _dangerous_. The behavior if the programmer makes a careless omission is exactly what they wouldn't want, and the only remedy is to check all of their register accesses _by hand_ to ensure that they handled all W1C and W0S fields correctly.

[...]

Issue #176, opened 2018-02-18
Generate API for "write 1 to toggle" registers


Serisium:
Hey all,

Over in rust-embedded/rfcs#40, user @kinnison  brought up that some of the STM32's USB control registers have a "write 0 to leave alone; write 1 to toggle" interface, and that svd2rust isn't generating the proper API for these registers. Since he hasn't filed an issue here yet, this is my attempt to document the feature.

In particular, the STM32F0x2 has a register for each USB endpoint with six of these toggleable bits. These registers are described in [RM0091](http://www.st.com/content/ccc/resource/technical/document/reference_manual/c2/f8/8a/f2/18/e6/43/96/DM00031936.pdf/files/DM00031936.pdf/jcr:content/translations/en.DM00031936.pdf) under the name `USB_EPnR`:

> When the application software writes ‘0, the value of DTOG_TX remains unchanged, while writing ‘1 makes the bit value toggle. This bit is read/write but it can only be toggled by writing 1.

[Here](https://raw.githubusercontent.com/GGreenwood/stm32f042/master/src/svd.rs) is a link to the current output generated by svd2rust v0.12.0 from [STM32F0x2.svd ](http://www.st.com/resource/en/svd/stm32f0_svd.zip)

**STM32f0x2.svd**
```
            <field>
              <name>DTOG_TX</name>
              <description>Data Toggle, for transmission
              transfers</description>
              <bitOffset>6</bitOffset>
              <bitWidth>1</bitWidth>
            </field>
```

**svd2rust API**
```
            #[doc = "Bit 6 - Data Toggle, for transmission transfers"]
            #[inline]
            pub fn dtog_tx(&self) -> DTOG_TXR {
                let bits = {
                    const MASK: bool = true;
                    const OFFSET: u8 = 6;
                    ((self.bits >> OFFSET) & MASK as u32) != 0
                };
                DTOG_TXR { bits }
            }
```

```
            #[doc = "Bit 6 - Data Toggle, for transmission transfers"]
            #[inline]
            pub fn dtog_tx(&mut self) -> _DTOG_TXW {
                _DTOG_TXW { w: self }
            }
```

Let me know if there's any other information I can dig up! I'm still learning embedded on Rust, and am really excited to see where things are headed.

Emilgardis:
To implement this we need to enable parsing of  `<modifiedWriteValues>` as [defined in the spec][elem_field].

To make it work you'd have to patch this element into the SVD aswell

```
            <field>
[...]

PR #541, opened 2021-10-05
modifiedWriteValues

burrbull:
#176, #540

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -673,10 +673,6 @@ pub fn fields(
             }
 
             if width == 1 {
-                if mwv != ModifiedWriteValues::Modify {
-                    unsafety = Some(Ident::new("unsafe", Span::call_site()));
-                }
-
                 match mwv {
                     ModifiedWriteValues::Modify => {
                         proxy_items.extend(quote! {
@@ -697,7 +693,7 @@ pub fn fields(
                             ///Sets the field bit
                             #inline
                             pub fn set_bit(self) -> &'a mut W {
-                                unsafe { self.bit(true) }
+                                self.bit(true)
                             }
                         });
                     }
@@ -706,7 +702,7 @@ pub fn fields(
                             ///Clears the field bit
                             #inline
                             pub fn clear_bit(self) -> &'a mut W {
-                                unsafe { self.bit(false) }
+                                self.bit(false)
                             }
                         });
                     }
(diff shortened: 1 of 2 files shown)


burrbull:
cc @rust-embedded/tools

src/generate/register.rs
```diff
@@ -652,19 +656,80 @@ pub fn fields(
...
 
             if width == 1 {
-                proxy_items.extend(quote! {
-                    ///Sets the field bit
-                    #inline
-                    pub #unsafety fn set_bit(self) -> &'a mut W {
-                        self.bit(true)
-                    }
+                if mwv != ModifiedWriteValues::Modify {
+                    unsafety = Some(Ident::new("unsafe", Span::call_site()));
```
adamgreig, line 660: What's the purpose of this? It seems like it just makes `bits()` for this field `unsafe` when it's not `Modify`, but that won't affect people calling `update` and also doesn't affect the `set_bit()` etc methods, right?

burrbull: What is `update`?

adamgreig: Sorry, I meant `modify()` ([talking about this](https://github.com/rust-embedded/svd2rust/issues/540#issuecomment-934580846)).

You're right, it will only affect `fn bit(bool)`. I'm not sure about making `bit()` unsafe though, since it's still safe to call and you can always write one of the other methods to get the same effect, which is why we make `bits()` safe if there's an enumeratedValue for every possible value. What would the safety requirements on `bit()` be, since you can safely call the other method to set it true or false?

src/generate/register.rs
```diff
@@ -652,19 +656,80 @@ pub fn fields(
...
+                                unsafe { self.bit(true) }
+                            }
+                        });
+                    }
+                    ModifiedWriteValues::ZeroToSet => {
+                        proxy_items.extend(quote! {
+                            ///Sets the field bit by passing zero
+                            #inline
+                            pub fn set_bit_by_zero(self) -> &'a mut W {
+                                unsafe { self.bit(false) }
```
adamgreig, line 710: For `ZeroToSet`, how can the user set a `1` to _not_ set the bit, if only `set_bit_by_zero()` exists, if the reset value (for `write()`) or current value (for `modify`) is currently 0 already?

Same thing with `ZeroToToggle` below.

src/generate/register.rs
burrbull: I've not thought about this. But if incorrect bit just do nothing we could make `bit()` safe everywhere.

burrbull:
Closed in favor of #603

PR #542, opened 2021-10-05
svd-parser 0.11

burrbull:


Emilgardis (commented):
Could we try using `--const-generics` and `--strict` on everything as option in the feature matrix? Would increase CI but I think it's worth it

burrbull:
> Could we try using `--const-generics` and `--strict` on everything as option in the feature matrix? Would increase CI but I think it's worth it

Not all SVDs feel good with `--strict`. But we can maximize usage of these flags.
Could you fix issue with multiple flags? I'm not good in bash.

Emilgardis:
Ill get to it later this today CET

Emilgardis:
Maybe just removing the bash script and completely use svd2rust-regress is a good thing to do as well. Not sure if that would be in the scope of this pr though.

therealprof (dismissed):
FWIW, looks good to me.

Emilgardis:
https://github.com/rust-embedded/svd2rust/blob/69a869c06a5b171028b5cf04fd3a9d7e63b2c9d5/.github/bors.toml

needs to be changed, I'm not sure if we have a policy on how to gather multiple results and gate bors on it, but I'd prefer to have a workflow that `needs: ci-linux`

burrbull:
const-generics require Rust 1.51, so you can't use it with 1.46

Emilgardis:
bors try

testing the combined ci check

.github/bors.toml
```diff
@@ -2,21 +2,6 @@ block_labels = ["needs-decision", "S-waiting-on-team"]
...
-  "ci-linux (stable, NXP, x86_64-unknown-linux-gnu, linux)",
-  "ci-linux (stable, RISC-V, x86_64-unknown-linux-gnu, linux)",
-  "ci-linux (stable, SiliconLabs, x86_64-unknown-linux-gnu, linux)",
-  "ci-linux (stable, Spansion, x86_64-unknown-linux-gnu, linux)",
-  "ci-linux (stable, STMicro, x86_64-unknown-linux-gnu, linux)",
-  "ci-linux (stable, Toshiba, x86_64-unknown-linux-gnu, linux)",
-  "ci-linux (1.46.0, Nordic, x86_64-unknown-linux-gnu, linux)",
-  "ci-linux (stable, x86_64-apple-darwin, osx)",
-  "ci-linux (stable, x86_64-pc-windows-msvc, windows)",
+  "CI", "Changelog check"
```
Emilgardis, line 5: actually this is wrong, changelog check should probably not be here, sometimes no change-log is valid

Emilgardis:
for clarity, the try run shouldn't fail because of https://github.com/esp-rs/esp32/pull/47 recently being merged

Emilgardis:
ugh, changelog check is not ran on trying 🤷
bors try-

Emilgardis (approved):
lgtm

leaving bors r=me for @burrbull, not sure if this is considered done, but looks to be to me

burrbull:
bors r=Emilgardis

PR #545, opened 2021-11-01
Possible fix for derived register arrays

### LINKED ISSUE #361: Support derivedFrom when target is register array
#357

robamu:
Possibly fixes #361 . 
I haven't tried the regression tool yet to ensure nothing went wrong, but I was able to generate a PAC for the VA108XX family of devices with these adaptions without any `syn` errors.

The issue was that the SVD file had a register array which was derived from a single register. In the previous implementation, only the other register is checked which meant that a single array was generated even though the derived one is still an array

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Bumped `xtensa-lx` and add `xtensa_lx::interrupt::InterruptNumber` implementation.
 - Don't use a mask when the width of the mask is the same as the width of the parent register.
 
+### Fixed
+
+- When expanding register arrays which were derived from a single
+  register, a single register was returned even though the derived type
+  was an array. This has been fixed by checking the derived and ancestor
+  type now
+
 ## [v0.19.0] - 2021-05-26
 
 ### Added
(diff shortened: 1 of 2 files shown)


burrbull:
I don't think this is correct behavior.
`dim` values of register should not be replaced by derived in my view

burrbull:
I don't think this is correct behavior.
`dim` values of register should not be replaced by derived in my view

burrbull:
cc @adamgreig @therealprof 
Need decision on how it works

burrbull:
https://github.com/rust-embedded/svd/blob/master/svd-rs/src/derive_from.rs

We just need to implement:
```
impl DeriveFrom for Register {
```
and
```
impl DeriveFrom for Cluster {
```

burrbull:
I've created 2 versions:
https://github.com/rust-embedded/svd/pull/158 and https://github.com/rust-embedded/svd/pull/159
Don't know which is right.

burrbull:
@robamu Try this https://github.com/rust-embedded/svd2rust/tree/derive_register

robamu:
I just tried it. It worked and the code has not changed (except the part with the used svd2rust commit)

PR #548, opened 2021-11-09
Added debug and trace output

robamu:
The application already has log levels and these additional logs will be disabled by default. They can be helpful when debugging something or tracing where an error/warning happens

### DIFF
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -6,7 +6,7 @@ use crate::svd::{
     Cluster, ClusterInfo, DeriveFrom, DimElement, Peripheral, Register, RegisterCluster,
     RegisterProperties,
 };
-use log::warn;
+use log::{debug, trace, warn};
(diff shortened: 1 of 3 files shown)


Emilgardis (approved):
Output:
image

I think this helps a bit, thank you!

bors r+

src/generate/peripheral.rs
```diff
@@ -168,16 +168,24 @@ pub fn render(
...
 
     // Push any register or cluster blocks into the output
+    debug!(
+        "Pushing {} register or cluster blocks into output",
+        ercs.len()
+    );
     let mut mod_items = TokenStream::new();
     mod_items.extend(register_or_cluster_block(&ercs, &defaults, None, config)?);
 
+    debug!("Pushing cluster information into output");
```
Emilgardis, line 178: always printing this even when there's no output is a bit strange, but maybe that's a good thing.

PR #549, opened 2021-11-09
fix bug in #537

### LINKED ISSUE #547: Reader for bit range [0:0] incorrect
I am working on a HAL implementation for a UART peripheral and came cross odd behaviour:

```rs
            ...
            if status_reader.rdavl().bit_is_set() {
                let data = uart.data.read().bits();
                Ok((data & 0xff) as u8)
            } else {
                Err(nb::Error::WouldBlock)
            }
            ...
```

The `rdavl().bit_is_set()` call always returns `true`,  even when the actual bit in the hardware was not set (I checked in peripheral view and by reading the whole register value as `u32`).

I then checked the `rdavl()` code in the PAC:

```rs
impl R {
    #[doc = "Bit 0 - Read Data Available"]
    #[inline(always)]
    pub fn rdavl(&self) -> RDAVL_R {
        RDAVL_R::new(self.bits != 0)
    }
    #[doc = "Bit 1 - Read Fifo NOT Full"]
    #[inline(always)]
    pub fn rdnfull(&self) -> RDNFULL_R {
        RDNFULL_R::new(((self.bits >> 1) & 0x01) != 0)
    }
    #[doc = "Bit 2 - RX Busy Receiving"]
    #[inline(always)]
    pub fn rxbusy(&self) -> RXBUSY_R {
        RXBUSY_R::new(((self.bits >> 2) & 0x01) != 0)
    }
```

This is odd. The call does not look correct, it should be `RDAVL_R::new((self.bits & 0x01) != 0)`

I checked the SVD file: 

```xml
          <name>RXSTATUS</name>
          <description>Status Register</description>
          <addressOffset>0x010</addressOffset>
          <access>read-only</access>
          <resetValue>0x0000000</resetValue>
          <fields>
            <field>
              <name>RDAVL</name>
              <description>Read Data Available</description>
              <bitRange>[0:0]</bitRange>
            </field>
            <field>
              <name>RDNFULL</name>
              <description>Read Fifo NOT Full</description>
              <bitRange>[1:1]</bitRange>
            </field>
```

Might this be a bug?

burrbull:
resolves #547

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
+- Bug with `use_mask`
 - Correct derive for register (cluster) array (needs `svd-rs` 0.11.2) 
 
 ### Added
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -310,7 +310,7 @@ pub fn fields(
         let use_mask = if let Some(size) = parent.properties.size {
             size != width
         } else {
-            false
+            true
         };
 
         let lookup_results = lookup(


Emilgardis (approved):

Emilgardis (approved):
I was more leaning towards the commit message, but it's fine. I linked the issue where this was found.

Leaving r=me for you if there's maybe something else

burrbull:
bors r=Emilgardis

PR #551, opened 2021-11-09
add tracing

Emilgardis:
This PR does some changes to how logging is done and displayed.

More work maybe needed.

Depends on rust-embedded/svd#172


Emilgardis:
currently looks like this

image

burrbull:
What is default log level?

Emilgardis:
Default log level is set on https://github.com/rust-embedded/svd2rust/blob/0db48384869e62072f2321bfb21ba32edc1cbedd/src/main.rs#L204

burrbull:
Any progress on this?

src/main.rs
```diff
@@ -193,29 +192,25 @@ fn run() -> Result<()> {
...
+        None => tracing_subscriber::EnvFilter::from_default_env()
+            .add_directive(tracing::Level::INFO.into()),
+    };
 
-    builder.init();
+    tracing_subscriber::fmt()
+        .without_time()
+        .with_target(false)
+        .with_env_filter(filter)
+        .compact()
```
Emilgardis, line 211: with compact
image

with full
image

I prefer compact, looks cleaner imo

Emilgardis: I'd want to remove the `render:expand:etcetc` things, but that would require a custom formatter, simple to do but maybe not neccessary, this shouldn't be used often

burrbull:
@Emilgardis should I make new release or wait for this PR?

Emilgardis:
Go ahead with a release, this is breaking but can be made to not be :)

PR #554, opened 2021-12-04
Generate safe bits writers when possible

newAM:
The 'bits' method of register writers can be safe if:
* there is a single field that covers the entire register
* that field can represent all values

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -168,14 +168,40 @@ pub fn render(
 
         mod_items.extend(w_impl_items);
 
-        mod_items.extend(quote! {
-            #[doc = "Writes raw bits to the register."]
-            #[inline(always)]
-            pub unsafe fn bits(&mut self, bits: #rty) -> &mut Self {
-                self.0.bits(bits);
-                self
+        // the writer can be safe if:
+        // * there is a single field that covers the entire register
+        // * that field can represent all values
+        let can_write_safe = match register
+            .fields
+            .as_ref()
+            .and_then(|fields| fields.iter().next())
+            .and_then(|field| field.write_constraint)
+        {
+            Some(WriteConstraint::Range(range)) => {
+                range.min == 0 && range.max == u64::MAX >> (64 - rsize)
             }
-        });
+            _ => false,
+        };
+
+        if can_write_safe {
+            mod_items.extend(quote! {
+                #[doc = "Writes raw bits to the register."]
+                #[inline(always)]
+                pub fn bits(&mut self, bits: #rty) -> &mut Self {
+                    unsafe { self.0.bits(bits) };
+                    self
+                }
+            });
+        } else {
+            mod_items.extend(quote! {
+                #[doc = "Writes raw bits to the register."]
+                #[inline(always)]
+                pub unsafe fn bits(&mut self, bits: #rty) -> &mut Self {
+                    self.0.bits(bits);
+                    self
+                }
+            });
+        }
 
         close.to_tokens(&mut mod_items);
     }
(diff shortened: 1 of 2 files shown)


burrbull:
`map(fn).flatten()` could be replaced with `flat_map(fn)`

newAM:
> `map(fn).flatten()` could be replaced with `flat_map(fn)`

I think `flat_map` is only an `Iterator` method: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map

The `map`/`flatten` chains are just on single `Option<T>` types to access deeply nested data in the case, I couldn't find a `flat_map` method for `Option<T>`.

> Explicit `deref` also unneeded

Done!  Good catch.

burrbull:
> I think `flat_map` is only an `Iterator` method: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map

I missed this is `Option`. What about `and_then`?

adamgreig (approved):
LGTM, @burrbull ?

burrbull:
> LGTM, @burrbull ?

This will not compile.

burrbull:
> > LGTM, @burrbull ?
> 
> This will not compile.

```rust
if register
            .fields
            .as_ref()
            .and_then(|fields| fields.iter().next())
            .and_then(|field| field.write_constraint)
            .and_then(|constraint| match constraint {
                WriteConstraint::Range(range)
                    if range.min == 0 && range.max == 2_u64.pow(rsize) - 1 =>
                {
                    None
                }
                _ => Some(()),
            })
            .is_some()
        {
            mod_items.extend(quote! {
                #[doc = "Writes raw bits to the register."]
                #[inline(always)]
                pub unsafe fn bits(&mut self, bits: #rty) -> &mut Self {
                    self.0.bits(bits);
                    self
                }
            });
        } else {
            mod_items.extend(quote! {
                #[doc = "Writes raw bits to the register."]
                #[inline(always)]
                pub fn bits(&mut self, bits: #rty) -> &mut Self {
                    unsafe { self.0.bits(bits); }
                    self
                }
            });
        }
```

thejpster:
I've looked at this twice now and I can't for the life of me work out what's going on here.

If it's just me, then that's ok. But if anyone else is struggling, maybe some comments would be useful, or to break up the big chain into smaller pieces.

But I love the theory, and this will clean up a bunch of code.

newAM:
> > LGTM, @burrbull ?
> 
> This will not compile.

Can you explain more?

I am testing this with the STM32WL and stm32-rs, I can compile the generated crate with these changes.

> I missed this is `Option`. What about `and_then`?

Good catch, that's much better.  Changed.

burrbull:
> Can you explain more?

`self.0.bits(bits)` is `unsafe`

newAM:
> > Can you explain more?
> 
> `self.0.bits(bits)` is `unsafe`

Fixed!  I forgot a `set -e` in my bash script that I used for checking this :(

> I've looked at this twice now and I can't for the life of me work out what's going on here.
> 
> If it's just me, then that's ok. But if anyone else is struggling, maybe some comments would be useful, or to break up the big chain into smaller pieces.

I broke it up to separate the logic of "can the write be safe" and the code generated as a result.  Let me know if that helped :)

src/generate/register.rs
```diff
@@ -168,14 +168,40 @@ pub fn render(
...
+        // * there is a single field that covers the entire register
+        // * that field can represent all values
+        let can_write_safe = match register
+            .fields
+            .as_ref()
+            .and_then(|fields| fields.iter().next())
+            .and_then(|field| field.write_constraint)
+        {
+            Some(WriteConstraint::Range(range)) => {
+                range.min == 0 && range.max == 2_u64.pow(rsize) - 1
```
burrbull, line 181: And last note:
We are support 64-bit registers. This `2_u64.pow(rsize)-1` will overflow for them.
Use `u64::MAX >> (64 - rsize)` instead.

newAM: Good catch!  Thanks for all the code review, I really appreciate it!

burrbull (approved):

PR #556, opened 2021-12-08
serde YAML/JSON input

burrbull:


src/util.rs
```diff
@@ -73,6 +73,20 @@ impl Default for Target {
     }
 }
 
+#[allow(unused)]
```
Emilgardis, line 76: Why allow unused? Should this not be defined in `main.rs` maybe?

src/lib.rs
```diff
@@ -532,7 +532,14 @@ pub enum SvdError {
...
 
-    let device = svd_parser::parse(xml)?;
+    let mut parser_config = svd_parser::Config::default();
+    parser_config.validate_level = if config.strict {
+        svd::ValidateLevel::Strict
+    } else {
+        svd::ValidateLevel::Weak
+    };
+
+    let device = svd_parser::parse_with_config(xml, &parser_config)?;
```
Emilgardis, line 542: oops :D

this should probably be mentioned in the changelog. 

> Correctly use config when using `svd2rust::generate` as a library

Emilgardis: also, why set it as default and then override it directly afterwards?

burrbull:
Should be fixed

src/util.rs
```diff
@@ -14,6 +14,26 @@ pub const BITS_PER_BYTE: u32 = 8;
 /// that are not valid in Rust ident
 const BLACKLIST_CHARS: &[char] = &['(', ')', '[', ']', '/', ' ', '-'];
 
+pub(crate) fn load_from(input: &str, config: &crate::util::Config) -> Result<Device> {
```
Emilgardis, line 17: I dont understand why this is living in util.rs instead if lib.rs

burrbull: I've tried to move `load_from` to `lib.rs` but get several errors looking like:

```
    = note: expected reference `&util::Config`
               found reference `&svd2rust::Config`
For more information about this error, try `rustc --explain E0308`.
```

Could you try it yourself?

Emilgardis: done, did some other changes as well.

the issue was that the util mod in main.rs and the util mod in lib.rs were separate things, and that means that the types were not equal.

burrbull:
Rebased

therealprof:
How would one use this? Are there parsable examples we could/should include?

burrbull:
Espressif test was disabled. See https://github.com/esp-rs/xtensa-lx/issues/14

burrbull:
For testing with CI. New `svdtools` release needed first.

adamgreig (approved):
LGTM, nice!

bors merge

Issue #563, opened 2022-01-11
Order of `replace_suffix` and `to_sanitized_snake_case`


duskmoon314:
Currently in the `register.rs`, `replace_suffix` is executed after `to_sanitized_snake_case`:

```rust
# line 443, 808
&util::replace_suffix(&f.name.to_sanitized_snake_case(), suffix),
```

Thus when facing an svd file like this will generate `pX_pwr_mod_sel`

```xml
<field>
  <dim>5</dim>
  <dimIncrement>1</dimIncrement>
  <dimIndex>C,D,E,F,G</dimIndex>
  <name>P%s_PWR_MOD_SEL</name>
  <description>PX_POWER POWER MODE Select</description>
  <bitRange>[2:2]</bitRange>
</field>
```

```rust
    #[doc = "Bit 2 - PX_POWER POWER MODE Select"]
    #[inline(always)]
    pub fn pC_pwr_mod_sel(&self) -> P_PWR_MOD_SEL_R {
        P_PWR_MOD_SEL_R::new(((self.bits >> 2) & 0x01) != 0)
    }
    #[doc = "Bit 3 - PX_POWER POWER MODE Select"]
    #[inline(always)]
    pub fn pD_pwr_mod_sel(&self) -> P_PWR_MOD_SEL_R {
        P_PWR_MOD_SEL_R::new(((self.bits >> 3) & 0x01) != 0)
    }
    #[doc = "Bit 4 - PX_POWER POWER MODE Select"]
    #[inline(always)]
    pub fn pE_pwr_mod_sel(&self) -> P_PWR_MOD_SEL_R {
        P_PWR_MOD_SEL_R::new(((self.bits >> 4) & 0x01) != 0)
    }
    #[doc = "Bit 5 - PX_POWER POWER MODE Select"]
    #[inline(always)]
    pub fn pF_pwr_mod_sel(&self) -> P_PWR_MOD_SEL_R {
        P_PWR_MOD_SEL_R::new(((self.bits >> 5) & 0x01) != 0)
    }
    #[doc = "Bit 6 - PX_POWER POWER MODE Select"]
    #[inline(always)]
    pub fn pG_pwr_mod_sel(&self) -> P_PWR_MOD_SEL_R {
        P_PWR_MOD_SEL_R::new(((self.bits >> 6) & 0x01) != 0)
    }
```

Should we replace the suffix before converting the name to snake case? Or have I made a mistake and this is exactly the desired behavior?

[...]

PR #564, opened 2022-01-12
fix: Replace suffix before converting to snake case

duskmoon314:
## Abstract

In #563, we noticed that converting fields' names to snake cases before replacing the suffix caused the generated methods not to satisfy the snake case. This PR fixes this issue.

## Change

Use `util::replace_suffix(&f.name, suffix).to_sanitized_snake_case()` instead of `util::replace_suffix(&f.name.to_sanitized_snake_case(), suffix)` when generating methods of fields.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Replace suffix in fields' name before converting to snake case when generating methods #563
+
 ## [v0.20.0] - 2021-12-07
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


adamgreig (approved):
Looks good, thank you!

The CI failure on nightly isn't an issue for this.

bors merge

adamgreig:
Oh, boo, since we aggregated all the CI jobs into one "CI", bors requires the nightly tests to pass as well, that's unfortunate. This needs https://github.com/esp-rs/xtensa-lx/issues/14 to be merged and published, but I guess that might not be until after Rust 1.59.

I'll open a PR to disable the Espressif CI in the meantime.

burrbull:
bors retry

PR #572, opened 2022-01-16
optional: structs instead of enums

burrbull:
Depends on #571 

Example:
изображение


Emilgardis:
this is neat, what's the use-case for it though?

burrbull:
> this is neat, what's the use-case for it though?

I don't have answer. It was more experiment than practice goal.

PR #574, opened 2022-01-17
disable the interrupt macro for MIPS as well

kiffie:
r? @burrbull

Yet another small MIPS fix in addition to yesterday's change. Sorry for submitting two separate PR instead of a single one. I did not realize that the interrupt macro can cause name clashes when including multiple device support modules into a single crate. I discovered this problem when trying to build the docs.


burrbull (approved):

kiffie:
It seems that something in the CI job went wrong when downloading https://raw.githubusercontent.com/posborne/cmsis-svd/master/data/Fujitsu/MB9AF1AxL.svd via curl. Can it be re-triggered?

burrbull:
bors retry

Issue #535, opened 2021-08-02
Add support for atomic xor/clear/set register operations for Raspberry Pi RP2040 microcontroller


richardanaya:
As per https://datasheets.raspberrypi.org/rp2040/rp2040-datasheet.pdf "2.1.2. Atomic Register Access"

<img width="632" alt="Screen Shot 2021-08-01 at 10 27 20 PM" src="https://user-images.githubusercontent.com/294042/127808539-91281dea-eb1e-4e6b-b8f4-8767c526b493.png">

henkkuli:
I have described one possible implementation for this in [a comment elsewhere](https://github.com/rp-rs/rp-hal/pull/143#issuecomment-930070759), but I don't know how to gate that for specific registers or peripherals only. I tried to look at the [corresponding SVD file](https://github.com/rp-rs/rp2040-pac/blob/main/svd/rp2040.svd) but haven't found any indication there that some registers should be atomic*. I also couldn't find a way to encode this info in a SVD file, at least based on a surface look at [svd_parser](https://docs.rs/svd-parser/0.10.2/svd_parser/index.html). Does anyone here have an idea how to mark only some peripherals/registers as atomic? Maybe have another file listing all atomic peripherals/registers? Or list each atomic peripheral as a flag?

* All registers except those belonging to SIO peripheral should be atomic.

burrbull:
As SVD doesn't have such info, the only way I see is to implement atomic operation for all registers, but make them `unsafe` as it done for `write_with_zero`.

henkkuli:
That would be unfortunate, but at least it would work.

By the way, I have a POC available in [my fork](https://github.com/henkkuli/svd2rust/tree/rp2040-atomic-poc) if someone wants to take a look. Would something like that fly? I can try to finish a PR, but for that I need some help with how to gate the feature for RP2040 target only.

adfernandes:
FWIW, on the `RP235X` at least here's the [definitive answer](https://github.com/raspberrypi/pico-feedback/issues/423) on "Which peripherals, specifically, support atomic modification?"

I'm trying to add this as a type-safe `trait` right now...

burrbull:
If those `write`s are marked `unsafe` do we really need to restrict them?
Lets just implement `xor/clear/set` for all `Reg`s.

jannic:
Just playing devils advocate: If `write`s are marked `unsafe`, why do we bother with a PAC at all and don't just do volatile writes do raw addresses?

[...]

PR #576, opened 2022-01-28
Add --mark_range flag

henkkuli:
This flag allows implementing a marker trait for registers in specific memory addresses. These can then be used to implement methods on registers. One such example is providing atomic access to only some of the registers directly in the PAC. Currently this is done in HAL, but this loses the ability to use writer proxies. It is also finicky to use raw register addresses. Moreover this allows keeping atomic and non-atomic usage in HAL more similar.

For an example on how to use this, see my [fork of rp2040-pac](https://github.com/henkkuli/rp2040-pac/commit/f701d72ca010b7e5451986718d89ee296230ebda). The most interesting changes regarding the use reside in
- [`generic_extension.rs`](https://github.com/henkkuli/rp2040-pac/commit/f701d72ca010b7e5451986718d89ee296230ebda#diff-5d2c6f640d0464d137f154be92f25979060a6c6c77d398a153fa39adaa79e944)
- [`update.sh`](https://github.com/henkkuli/rp2040-pac/commit/f701d72ca010b7e5451986718d89ee296230ebda#diff-e14594f6bb9b29ac0146c04e78ae33e8e92f2f00709fefab97359dea31b94d11)

Even though I have an example only for the RP2040, I believe that other architectures and devices could also benefit from this. For example on SMT32 MCUs this could be used to implement a nicer API for bit-banding.

Addresses #535

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,7 +12,7 @@ use clap::{App, Arg};
 
 use svd2rust::{
     generate, load_from,
-    util::{build_rs, Config, SourceType, Target},
+    util::{build_rs, parse_address, Config, MarkRange, SourceType, Target},
 };
 
 fn run() -> Result<()> {
@@ -103,6 +103,20 @@ fn run() -> Result<()> {
                 .takes_value(true)
                 .possible_values(&["off", "error", "warn", "info", "debug", "trace"]),
         )
+        .arg(
+            Arg::with_name("mark_range")
+                .long("mark_range")
+                .multiple(true)
+                .number_of_values(3)
+                .help("Mark registers in given range with a marker trait")
+                .long_help(
+                    "Mark registers in given range with a marker trait.
+The first argument after the flag sets the name for the marker trait.
+The second and third argument are the start and end addresses of registers
+which this trait is applied for.",
+                )
+                .value_names(&["trait", "range start", "range end"]),
+        )
         .version(concat!(
             env!("CARGO_PKG_VERSION"),
             include_str!(concat!(env!("OUT_DIR"), "/commit-info.txt"))
@@ -168,6 +182,18 @@ fn run() -> Result<()> {
         source_type = SourceType::from_path(Path::new(file))
     }
 
+    let mut mark_range_iter = cfg.values("mark_range", Filter::Arg).into_iter().flatten();
+    let mut mark_ranges = Vec::new();
+    while let (Some(name), Some(start), Some(end)) = (
+        mark_range_iter.next(),
+        mark_range_iter.next(),
+        mark_range_iter.next(),
+    ) {
+        let start = parse_address(&start).context("Invalid range start")?;
+        let end = parse_address(&end).context("Invalid range end")?;
+        mark_ranges.push(MarkRange { name, start, end });
+    }
+
     let config = Config {
         target,
         nightly,
@@ -179,6 +205,7 @@ fn run() -> Result<()> {
         strict,
         output_dir: path.clone(),
         source_type,
+        mark_ranges,
     };
 
     info!("Parsing device from SVD file");
--- a/src/generate/device.rs
+++ b/src/generate/device.rs
@@ -3,6 +3,7 @@ use proc_macro2::{Ident, Span, TokenStream};
 use quote::{quote, ToTokens};
 
 use log::debug;
+use std::collections::HashSet;
 use std::fs::File;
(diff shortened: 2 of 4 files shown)


src/main.rs
```diff
@@ -103,6 +103,20 @@ fn run() -> Result<()> {
...
+            Arg::with_name("mark_range")
+                .long("mark_range")
+                .multiple(true)
+                .number_of_values(3)
+                .help("Mark registers in given range with a marker trait")
+                .long_help(
+                    "Mark registers in given range with a marker trait.
+The first argument after the flag sets the name for the marker trait.
+The second and third argument are the start and end addresses of registers
+which this trait is applied for.",
```
Emilgardis, line 116: I think this deserves an example of how it could look.

Emilgardis (commented):
really neat!

adamgreig:
I think this is a really nice way for specific PACs to implement chip-specific functionality like you've done, nice work!

Is there any reason in your `update.sh` you have each peripheral's address range spelled out, instead of just putting 0x40000000-0x40070000, 0x50200000-0x50204000, 0x50300000-0x50304000, 0x50400000-0x50404000? It seems like all the peripheral spaces are contiguous?

henkkuli:
> Is there any reason in your `update.sh` you have each peripheral's address range spelled out, instead of just putting 0x40000000-0x40070000, 0x50200000-0x50204000, 0x50300000-0x50304000, 0x50400000-0x50404000?

Not really, other than to show that you can specify the same marker multiple times, and to list all peripherals that allow atomic access. I still need to go through that list once more; it seems that *some* USB peripheral registers allow atomic access, while now *none* implement it. It could still be simplified.

adamgreig:
Could you add an entry to the changelog for this, please?

Also, I wonder if it's a bit unnatural to have command line arguments like `--flag a b c`: it looks like only `a` is part of `--flag` and `b` and `c` are extra positional arguments... maybe they should be comma-separated? Maybe it's not a big deal, I just don't feel like I've seen that sort of command line arguments much before.

burrbull:
Even with new comments it is not obvious how it works. Please add more complete example to CI tests.

henkkuli:
> Could you add an entry to the changelog for this, please?

Sure, I'll add an entry with next batch of changes.



> Also, I wonder if it's a bit unnatural to have command line arguments like `--flag a b c`

I originally did that because I didn't want to write a parser for separating the marker name from the address, especially because I didn't know would this even be something that is wanted for svd2rust. Now that it seems that there is interest, I will write a proper parser. And now that I think of it, it shouldn't be anything more complicated than just a couple of splits.

I was thinking that maybe the following would be natural: `--mark_range MarkerName:0x1234-0x5678`

> Even with new comments it is not obvious how it works. Please add more complete example to CI tests.

I can add a CI test, but I'm not exactly sure how. Should I make it a new option that is passed to rust2svd in [`ci/script.sh:13`](https://github.com/rust-embedded/svd2rust/blob/v0.21.0/ci/script.sh#L13)? And make it part of `OPTIONS=all`?

burrbull:
> I can add a CI test, but I'm not exactly sure how. Should I make it a new option that is passed to rust2svd in [`ci/script.sh:13`](https://github.com/rust-embedded/svd2rust/blob/v0.21.0/ci/script.sh#L13)? And make it part of `OPTIONS=all`?

Not necessary. Any way you feel comfortable. You can just add separate job. We'll do refactoring some other day.

adamgreig:
> I originally did that because I didn't want to write a parser for separating the marker name from the address, especially because I didn't know would this even be something that is wanted for svd2rust. Now that it seems that there is interest, I will write a proper parser. And now that I think of it, it shouldn't be anything more complicated than just a couple of splits.
> 
> I was thinking that maybe the following would be natural: `--mark_range MarkerName:0x1234-0x5678`

That suggested format looks good to me, thanks!

Dirbaio:
What does it do if there's several instances of the peripheral at different addrs, and only one is covered by the range in `--mark_range`?

burrbull:
@henkkuli Could you rebase this on current master branch?

burrbull:
> I was thinking that maybe the following would be natural: `--mark_range MarkerName:0x1234-0x5678`

Maybe something like: `--registers 0x1234-0x5678,0x11234-0x15678:Marker1,Marker2,Marker3` ?

PR #577, opened 2022-01-29
Bring documentation on MSP430 up to date.

cr1901:
Crate versions must be `0.3.0` or higher now; minor version `0.2.0` and below is unsupported due to (among other things) bumping to bare-metal `1.0.0` and [using](https://github.com/rust-lang/rust/pull/93219) Rust inline assembly to implement useful operations.


burrbull (dismissed):

cr1901:
Looks like I forgot to update one of the tests... I'll fix this later today...

burrbull (approved):

PR #583, opened 2022-03-18
OFFSET & WIDTH constanst

### LINKED ISSUE #582: The ability to obtain offset values for the modifiable bit fields of registers
To be able to manipulation of peripheral registers through bit-banding, the values of the offset of the modifiable bit fields in the registers are required.

For example, there are functions for bit-banding here: https://docs.rs/stm32f1xx-hal/0.9.0/stm32f1xx_hal/bb/index.html

Implementation variant:
```rust
pub mod perih_name {
   ...
   pub mod reg_name {
      ...
      pub const field_name_OFFSET: u8 = ...;
      ...
   }
   ...
}
```
Example of an implementation variant (stm32f1):
```rust
pub mod rcc {
   ...
   pub mod apb2enr {
      ...
      pub const USART1EN_OFFSET: u8 = 14;
      ...
      impl<'a> USART1EN_W<'a> {
         ...
         pub fn bit(self, value: bool) -> &'a mut W {
            self.w.bits = (self.w.bits & !(0x01 << USART1EN_OFFSET)) | ((value as u32 & 0x01) << USART1EN_OFFSET);
            self.w
         }
      }
      ...
      impl R {
         ...
         pub fn usart1en(&self) -> USART1EN_R {
            USART1EN_R::new(((self.bits >> USART1EN_OFFSET) & 0x01) != 0)
         }
         ...
      }
      ...
   }
   ...
}
```

Usage example:
```rust
unsafe {
   use stm32f1xx_hal::{self as hal, pac};
   hal::bb::set(&(*pac::RCC::PTR).apb2enr, pac::rcc::apb2enr::USART1EN_OFFSET);
}
```

burrbull:
Closes #582

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Constant values of field offsets and width
 - added `dyn` keyword to sanatizer.
+
 ### Changed
 
 - Generate Rust arrays for all register & cluster arrays with sequential_addresses.
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -347,8 +347,8 @@ pub fn fields(
         let mut evs_r = None;
 
         let field_dim = match f {
-            Field::Array(_, de) => {
-                let (first, index) = if let Some(dim_index) = &de.dim_index {
+            Field::Array(info, de) => {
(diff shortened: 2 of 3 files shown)


burrbull:
> would it make sense to actually give the offset in the documentation?

It is already there.
изображение
Or you means something else?

burrbull:
Close in favour of #598

Issue #492, opened 2021-02-21
Support arrays of peripherals


daym:
It would be nice if svd2rust also supported peripheral arrays of SVD 1.3.

See https://arm-software.github.io/CMSIS_5/SVD/html/elem_peripherals.html#elem_peripheral .

>Starting version 1.3 of the SVD specification, arrays of peripherals can be specified. The single peripheral description gets duplicated automatically into an array. The number of array elements is specified by the `<dim>` element. To define arrays, the <name> needs the format `myPeripheral[%s]`. The tag `<dimIncrement>` specifies the address offset between two peripherals. To create copies of a peripheral using different names, you must use the derivedFrom attribute.

duskmoon314:
@daym This feature has been implemented and merged. 😄

Emilgardis:
Fixed in #592

Emilgardis:
For your information @duskmoon314 you can auto-close issues with the following syntax: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

Very useful :D

PR #592, opened 2022-04-22
add code generation support of peripheral arrays

duskmoon314:
## Brief Intro
- generate ArrayProxy of peripherals when const_generic is true
- generate separate struct of peripherals when const_generic is false

close issue: #492

## Design

**Notice:** This is my first try of implement this feature. Please provide suggestions for this PR to better implement this feature.

This PR aims to achieve what issue #492 is asking for. This PR implements a basic try of generating codes depending on the `const_generic` flag.

Let's say we have an svd file with descriptions of peripherals like this:

```XML
<peripheral>
    <dim>2</dim>
    <dimIncrement>0x100</dimIncrement>
    <name>PTEST[%s]</name>
    <groupName>PGroup</groupName>
    <baseAddress>0x20000000</baseAddress>
    <addressBlock>
        <offset>0</offset>
        <size>0x100</size>
        <usage>registers</usage>
    </addressBlock>
    <registers>
        ...
    </registers>
</peripheral>
```

If the `const_generic` is `false`, there is no `ArrayProxy`. So I generate separate structs and only one mod:

```rust
pub struct PTEST0 {
    _marker: PhantomData<*const ()>,
}
...
impl PTEST0 {
    #[doc = r"Pointer to the register block"]
    pub const PTR: *const ptest::RegisterBlock = 0x2000_0000 as *const _;
    ...
}
impl Deref for PTEST0 {
    type Target = ptest::RegisterBlock;
    #[inline(always)]
    fn deref(&self) -> &Self::Target {
        unsafe { &*Self::PTR }
    }
}

pub struct PTEST1 {
    _marker: PhantomData<*const ()>,
}
...
impl PTEST1 {
    #[doc = r"Pointer to the register block"]
    pub const PTR: *const ptest::RegisterBlock = 0x2000_0100 as *const _;
    ...
}
impl Deref for PTEST1 {
    type Target = ptest::RegisterBlock;
    #[inline(always)]
    fn deref(&self) -> &Self::Target {
        unsafe { &*Self::PTR }
    }
}
#[doc = "PTEST"]
pub mod ptest;
```

**NOTICE:** The following design is actually incorrect and has been removed. See the following comment.
<details>
If the `const_generic` is `true`, we can use `ArrayProxy`. So I generate code like this:

```rust
pub struct PTEST {
    _marker: PhantomData<*const ()>,
}
...

pub struct Peripherals {
    #[doc = "PTEST"]
    pub PTEST: crate::ArrayProxy<PTEST, 2usize, 0x0100>,
...
impl Peripherals {
...
    pub unsafe fn steal() -> Self {
        DEVICE_PERIPHERALS = true;
        Peripherals {
            PTEST: crate::ArrayProxy::new(),
```

Since `_array` is a private field of `ArrayProxy`, I add a `new` method to implement the `steal` method of `Peripherals`.
</details>
I also add the support of using `name_of` on `MaybeArray<PeripheralInfo>` to achieve this implementation.

Well, I haven't fully tested this implementation yet. So there may be problems that need to be solved.

### DIFF
--- a/src/generate/array_proxy.rs
+++ b/src/generate/array_proxy.rs
@@ -15,14 +15,6 @@ pub struct ArrayProxy<T, const COUNT: usize, const STRIDE: usize> {
 }
 
 impl<T, const C: usize, const S: usize> ArrayProxy<T, C, S> {
-    /// Create a new ArrayProxy.
-    #[inline(always)]
-    #[allow(unused)]
-    pub(crate) fn new() -> Self {
-        Self {
-            _array: marker::PhantomData,
-        }
-    }
     /// Get a reference from an [ArrayProxy] with no bounds checking.
     pub unsafe fn get_ref(&self, index: usize) -> &T {
         let base = self as *const Self as usize;
--- a/src/generate/array_proxy.rs
+++ b/src/generate/array_proxy.rs
@@ -17,6 +17,7 @@ pub struct ArrayProxy<T, const COUNT: usize, const STRIDE: usize> {
 impl<T, const C: usize, const S: usize> ArrayProxy<T, C, S> {
     /// Create a new ArrayProxy.
     #[inline(always)]
+    #[allow(unused)]
     pub(crate) fn new() -> Self {
         Self {
             _array: marker::PhantomData,
--- a/src/generate/array_proxy.rs
+++ b/src/generate/array_proxy.rs
@@ -1,4 +1,3 @@
-
 /// Access an array of `COUNT` items of type `T` with the items `STRIDE` bytes
 /// apart.  This is a zero-sized-type.  No objects of this type are ever
 /// actually created, it is only a convenience for wrapping pointer arithmetic.
@@ -12,10 +11,17 @@ pub struct ArrayProxy<T, const COUNT: usize, const STRIDE: usize> {
     /// As well as providing a PhantomData, this field is non-public, and
     /// therefore ensures that code outside of this module can never create
     /// an ArrayProxy.
-    _array: marker::PhantomData<T>
+    _array: marker::PhantomData<T>,
 }
 
 impl<T, const C: usize, const S: usize> ArrayProxy<T, C, S> {
+    /// Create a new ArrayProxy.
+    #[inline(always)]
+    pub(crate) fn new() -> Self {
+        Self {
+            _array: marker::PhantomData,
+        }
+    }
     /// Get a reference from an [ArrayProxy] with no bounds checking.
     pub unsafe fn get_ref(&self, index: usize) -> &T {
         let base = self as *const Self as usize;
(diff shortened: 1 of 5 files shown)


src/generate/array_proxy.rs
```diff
@@ -12,10 +11,17 @@ pub struct ArrayProxy<T, const COUNT: usize, const STRIDE: usize> {
...
     /// therefore ensures that code outside of this module can never create
     /// an ArrayProxy.
-    _array: marker::PhantomData<T>
+    _array: marker::PhantomData<T>,
 }
 
 impl<T, const C: usize, const S: usize> ArrayProxy<T, C, S> {
+    /// Create a new ArrayProxy.
+    #[inline(always)]
+    pub(crate) fn new() -> Self {
```
burrbull, line 20:
```suggestion
    #[allow(unused)]
    pub(crate) fn new() -> Self {
```

duskmoon314:
> Need advanced test in CI with SVD containing peripheral array. (possibly already present, need to check)

It seems that CI use SVD in [posborne/cmsis-svd](https://github.com/posborne/cmsis-svd) to test. I try searching `dim` in this repo and cannot find a showcase of a peripheral array. May need to find a real example or build up one for testing.

> [ci failed](https://github.com/rust-embedded/svd2rust/runs/6131289141?check_suite_focus=true#step:4:106): pattern bindings after an `@` are unstable

I use `p @ Peripheral::Array(_, dim)` to extract `dim_element` and keep the `p: MaybeArray<PeripheralInfo>` at the same time. It is stable after 1.56. I will try to find out an alternative.

Emilgardis:
for future reference. If you need to break msrv, do it ;)

duskmoon314:
I think I have misunderstood how `ArrayProxy` works. The current design cannot use `ArrayProxy` to calculate the address since `ArrayProxy` uses these lines:

```rust
let base = self as *const Self as usize;
let address = base + S * index;
&*(address as *const T)
```

But the generated code when `const_generic` is `false` seems to work fine. And the usage in this way is the same as describing peripherals separately in the svd file. (I rearrange [the svd file](https://github.com/duskmoon314/aw-pac/blob/main/svd/d1_unofficial.svd) that I am maintaining locally and test a simple case.)

For now, I think maybe there is no need to provide something like `ArrayProxy` for peripheral arrays. Any suggestions?

burrbull:
> For now, I think maybe there is no need to provide something like `ArrayProxy` for peripheral arrays. Any suggestions?

Instead of `ArrayProxy` you could just create independent instances like if they are just `derivedFrom` first one.

duskmoon314:
> Instead of `ArrayProxy` you could just create independent instances like if they are just `derivedFrom` the first one.

Yes. For now, the generated code is only organized in this way.

Though I think there is no problem, I still want to ask if there are any suggestions.

burrbull:
I don't have suggestions now.
Except missing CI test.

burrbull (approved):

PR #580, opened 2022-02-10
more ci actions

burrbull:

PR #598, opened 2022-04-26
WriteField aliases

burrbull:


burrbull:
@therealprof
This change has 2 sides.
As it uses type aliases generated docs is not ideal, as write proxy related methods are splitted on 2-3 pages.
In other hand code is much cleaner as much less boilerplate.

But much more significant.
`cargo check` on generated pac more then 2 times faster. (for stm32f4)
`cargo build --release` ~ same time, but 10% less memory usage.

Emilgardis (dismissed):

Emilgardis:
reviewing #580 first as this changeset depends on it

burrbull:
I'm only not sure do we need last commit.
All other are as close to ideal as possible.

Emilgardis:
My bad, meant here https://github.com/rust-embedded/svd2rust/blob/bf7c24ef0640eff21b4c782a77972082b6c0865d/src/generate/register.rs#L655, not in the generated code

burrbull:
> My bad, meant here
> 
> https://github.com/rust-embedded/svd2rust/blob/bf7c24ef0640eff21b4c782a77972082b6c0865d/src/generate/register.rs#L655
> , not in the generated code

This is related to last commit I said about.

Emilgardis:
yes exactly, that is what I was commenting about if it was unclear <3

burrbull:
> yes exactly, that is what I was commenting about if it was unclear <3

I removed duplication, but not my doubts.
But I can revert it in future if i find it useful.

Emilgardis:
sorry, I didn't mean remove the whole change, just moving some things out to outside the if let. a lot of the code inside was very similar imo, and I think it could be streamlined

burrbull:
> sorry, I didn't mean remove the whole change, just moving some things out to outside the if let. a lot of the code inside was very similar imo, and I think it could be streamlined

I did not removed commit. I rewrote it as you asked me.

burrbull:
ping @Emilgardis

Emilgardis (approved):

burrbull:
bors r=Emilgardis

PR #606, opened 2022-05-08
derive_more

burrbull:


burrbull:
r? @Emilgardis 
I removed `prebuild` task in favor of independence caches as sometimes it does not work correctly.

Saying about main change, it adds optional deriving of `Deref` and `From` traits which significantly decrease boilerplate, but require advanced dependency and can slow down compilation, especially first one.

src/util.rs
```diff
@@ -45,6 +46,7 @@ impl Default for Config {
             keep_list: false,
             strict: false,
             pascal_enum_values: false,
+            derive_more: false,
```
Emilgardis, line 49: I think the default should be that it's enabled. That probably means the flag needs to be negated, e.g `derive_less` perhaps

Emilgardis (changes requested):
over all looks good, just one concern with defaults

src/util.rs
Emilgardis: `no_derive_more` ?

Emilgardis: I take that back, it should not be defaulted to on

Emilgardis (dismissed):
im fine with this, no concerns

src/util.rs
burrbull: As this requires advanced dependency, it is breaking change for all svd2rust-generated crates.
So I don't think it is good idea to make it default.

burrbull:
conflict. again. fixed

burrbull:
cc @Emilgardis

Emilgardis (approved):
bors r+

PR #608, opened 2022-05-09
ci: add date and changelog to pre-release

duskmoon314:
- Compress binaries for Linux and MacOS
- Extract the corresponding section in CHANGELOG and attach it to the release
- Add the date to the title

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,13 +14,13 @@ jobs:
         include:
           - os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
-            suffix: ""
+            suffix: ".gz"
           - os: macos-latest
             target: x86_64-apple-darwin
-            suffix: ""
+            suffix: ".gz"
           - os: macos-latest
             target: aarch64-apple-darwin
-            suffix: ""
+            suffix: ".gz"
           - os: windows-latest
             target: x86_64-pc-windows-msvc
             suffix: ".exe"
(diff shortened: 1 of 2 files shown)


burrbull:
No need in unzipped versions

.github/workflows/release.yml
```diff
@@ -41,32 +41,74 @@ jobs:
...
+          CHANGELOG="${CHANGELOG//$'\n'/'%0A'}"
+          CHANGELOG="${CHANGELOG//$'\r'/'%0D'}"
+          echo "::set-output name=changelog::$CHANGELOG"
+      - name: Extract released changelog
+        id: extract-released-changelog
+        if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+        run: |
+          CHANGELOG=$(markdown-extract ${{ github.ref_name }} CHANGELOG.md)
+          LATEST_TAG=$(git tag --sort=v:refname |& tail -n 2)
+          CHANGELOG=${CHANGELOG}$'\n\n'"[Unreleased]: https://github.com/rust-embedded/svd2rust/compare/${LATEST_TAG//$'\n'/'...'}"
```
Emilgardis, line 94: shouldn't this be 
```suggestion
          CHANGELOG=${CHANGELOG}$'\n\n'"[${LATEST_TAG}]: https://github.com/rust-embedded/svd2rust/compare/${LATEST_TAG//$'\n'/'...'}"
```

.github/workflows/release.yml
```diff
@@ -41,32 +41,74 @@ jobs:
...
+          CHANGELOG="${CHANGELOG//'%'/'%25'}"
+          CHANGELOG="${CHANGELOG//$'\n'/'%0A'}"
+          CHANGELOG="${CHANGELOG//$'\r'/'%0D'}"
+          echo "::set-output name=changelog::$CHANGELOG"
+      - name: Extract released changelog
+        id: extract-released-changelog
+        if: ${{ startsWith(github.ref, 'refs/tags/v') }}
+        run: |
+          CHANGELOG=$(markdown-extract ${{ github.ref_name }} CHANGELOG.md)
+          LATEST_TAG=$(git tag --sort=v:refname |& tail -n 2)
```
Emilgardis, line 93:
```suggestion
          LATEST_TAG=${GITHUB_REF#refs/tags/}
```

.github/workflows/release.yml
```diff
@@ -41,32 +41,74 @@ jobs:
...
+        id: extract-unreleased-changelog
+        if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
+        run: |
+          CHANGELOG=$(markdown-extract Unreleased CHANGELOG.md)
+          LATEST_TAG=$(git tag --sort=v:refname |& tail -n 1)
+          CHANGELOG="[Latest Commit](https://github.com/rust-embedded/svd2rust/commit/${{ github.sha }})"$'\n\n'${CHANGELOG}$'\n\n'"[Unreleased]: https://github.com/rust-embedded/svd2rust/compare/${LATEST_TAG}...HEAD"
+          CHANGELOG="${CHANGELOG//'%'/'%25'}"
+          CHANGELOG="${CHANGELOG//$'\n'/'%0A'}"
+          CHANGELOG="${CHANGELOG//$'\r'/'%0D'}"
+          echo "::set-output name=changelog::$CHANGELOG"
```
Emilgardis, line 87: I think this overall is wrong, this should be added to the changelog from the start, and not added afterwards in the release

Emilgardis (changes requested):

.github/workflows/release.yml
duskmoon314: I think you mean marking the `Unreleased` section as `vX.Y.Z`, building binaries, and creating release should be in one action? If so, I need to think about how to reimplement this action.

It looks like all the previous releases are done by hand, so I just add action to build binaries and upload them.

Emilgardis: No, this specifically I mean that the changelog should already contain this information, e.g the changeset between this release and the last, and on tag pushes the changelog should use the changelog entry for that specific version.

A release tag should not have its changelog entries under unreleased

duskmoon314: I see. Need time to figure out how to extract that information.

Emilgardis: See for example how it was solved in cross

https://github.com/cross-rs/cross/blob/4deac830a4992a70f7cb070eda38d5ff7cfcbf86/.github/actions/cargo-publish/action.yml#L11

duskmoon314:
Now I use the solution kind of a copy from cross. The body is extracted from `CHANGELOG` and marked as `pre-release` if it is `Unreleased`. But there isn't a link to the comparison page.

Any suggestion on improving this? @Emilgardis 

An example is like below

image

Emilgardis:
In changelog.md, add the tag..head link as the previous version would do but instead add it directly to the file. When we do a release, we'll have to make sure that a new link is added.

One note from me, is the above even needed, the releases page has a Changes link which serves the same purpose

r? burbull

@burrbull you've done the most releases, what do you think about this 😁

Emilgardis (approved):
Lgtm, with or without adding the changes "link"

burrbull:
> @burrbull you've done the most releases, what do you think about this grin

LGTM. I'm planning to release after this is merged.

@duskmoon314 is this still draft?

duskmoon314:
I think if it's ok to not have the link for now, this is not a draft. Currently, I use `changelog-reader-action` and the link is lost.

---

**TLDR**: We need to delete the previous release made by the action which is tagged as `pre-release`

Besides, the action to make the release doesn't support deleting the previous one. So it just updates the one with the same tag. We need to delete the release made by the previous action to remove uncompressed binaries.

Notes: now the action tags the unreleased version with `Unreleased`. So we need to delete the previous one to not confuse people. 😂

Emilgardis:
> Notes: now the action tags the unreleased version with `Unreleased`. So we need to delete the previous one to not confuse people. 😂

I don't understand, how would this cause confusion?

duskmoon314:
> how would this cause confusion?

In my point of view, finding two `pre-release` status releases on the release page is somewhat confusing. Though one is outdated once this is merged.

> why does it not touch the uncompressed binaries?

It seems the action will only update the release if it finds the release existed. Then it updates artifacts with the same name and keeps others. That is, the previous action updated an artifact called `a`, and the new action uploads `a.gz` and keeps `a` still in the release.

My opinion is to delete the previous one which is tagged `pre-release` and let the action create a new one tagged `Unreleased`. This will solve the two problems above.

Emilgardis:
ah right, you mean in the first run of this only, in future runs it won't matter.

Was thinking you meant it will always do this

duskmoon314:
Sorry for not being clear. Yes, I indeed mean the first run.

burrbull:
@duskmoon314 What do you think about similar work for other tools like `svdtools`, `cargo-binutils`?

duskmoon314:
@burrbull You mean adding action like this to other projects? I am definitely willing to help.

PR #612, opened 2022-05-21
Enum typename considering name in enumeratedValues

luojia65:
svd2rust has a good mechanism to parse enumeratedValues into Rust enums. However, current code would pick name of the first field the enum resides, and use `pub type` to re-export it into other names.

SVD files provide `name` field in enumeratedValues, this commit make use of them to name the Rust enums it generated. After this commit, more information of SVD file is considered in output pac crate. Users may also discover a significant drop in amount of types, which will speed up rustdoc generation and crate compilation.

Before this pull request:
<details>

图片
</details>

After this pull request:
<details>

Note that INTERRUPT_ENABLE_A comes from enumeration name InterruptEnable, but not register name interrupt_enable.
图片
</details>

SVD file for this example:
<details>

```xml
            <field>
              <name>fifo_error_enable</name>
              <description>Transmit or receive FIFO error interrupt enable</description>
              <lsb>29</lsb>
              <msb>29</msb>
              <writeConstraint>
                <useEnumeratedValues>true</useEnumeratedValues>
              </writeConstraint>
              <enumeratedValues>
                <name>InterruptEnable</name>
                <enumeratedValue>
                  <name>enable</name>
                  <description>Enable interrupt</description>
                  <value>1</value>
                </enumeratedValue>
                <enumeratedValue>
                  <name>disable</name>
                  <description>Disable interrupt</description>
                  <value>0</value>
                </enumeratedValue>
              </enumeratedValues>
            </field>
            <field>
              <name>arbitrate_lost_enable</name>
              <description>Arbitration lost interrupt enable</description>
              <lsb>28</lsb>
              <msb>28</msb>
              <writeConstraint>
                <useEnumeratedValues>true</useEnumeratedValues>
              </writeConstraint>
              <enumeratedValues derivedFrom="InterruptEnable" />
            </field>
            <field>
              <name>not_acknowledged_enable</name>
              <description>Not-acknowledged response interrupt enable</description>
              <lsb>27</lsb>
              <msb>27</msb>
              <writeConstraint>
                <useEnumeratedValues>true</useEnumeratedValues>
              </writeConstraint>
              <enumeratedValues derivedFrom="InterruptEnable" />
            </field>
<!-- and all other fields in this register -->
```
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Generated enum names now consider `name` field in `enumeratedValues`
+
 ## [v0.24.0] - 2022-05-12
 
 [commits][v0.24.0]
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -9,7 +9,8 @@ use proc_macro2::{Ident, Punct, Spacing, Span, TokenStream};
 use quote::{quote, ToTokens};
 
 use crate::util::{
-    self, Config, ToSanitizedPascalCase, ToSanitizedSnakeCase, ToSanitizedUpperCase, U32Ext,
+    self, Config, ToSanitizedConstantCase, ToSanitizedPascalCase, ToSanitizedSnakeCase,
+    ToSanitizedUpperCase, U32Ext,
 };
 use anyhow::{anyhow, Result};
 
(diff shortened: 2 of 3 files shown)


burrbull:
Some of enums can be exported from other `registers`/`peripherals` (read `modules`)

luojia65:
Looks like MSRV 1.51.0 does not support new style format macro: `format!("{variable}")`

Emilgardis:
> Looks like MSRV 1.51.0 does not support new style format macro: `format!("{variable}")`

it was added in [1.58.0](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings)

Emilgardis:
im not sure I like this change as it currently stands, it is imo less clear what is happening in the docs

luojia65:
There are register spec type parameters in BitWriters, resulting in that write proxies cannot be re-exported properly. Any good opinion?

burrbull:
> cannot be re-exported properly

and should not be

I like idea to use `enumaratedValues` name for enums, but not for readers and writers.

burrbull:
Could you move `ConstantCase`, docs and other code style changes in separate PR?
Possibly it makes sense to collect all case related traits in 1 `ToCase`.

luojia65:
I'll try to split this pull request, thanks! ;)

PR #615, opened 2022-06-10
fix: change to actual support arch in README of svd2rust-regress

duskmoon314:


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Fix the architecture in README.md of svd2rust-regress to those actual
+  supported
+
```
Emilgardis, line 12: don't think this needs to be documented

duskmoon314: Agree, I will remove it now

Emilgardis (approved):
awesome


bors r+

PR #617, opened 2022-06-22
feat: add argument feature_group

### LINKED ISSUE #614: An idea of using `groupName` as cfg feature
# Motivation

Recently I am working on two MCU sharing most of the peripherals. So I searched for issues to find if there were any excellent solutions and found #393. It seems hard to parse multiple SVD files and merge them into one crate automatically. Thus I came up with the idea of using `groupName` as the cfg feature to partly solve this problem. This idea has concerns and may not be a very good idea. But #393 is still open, so I want to share and discuss this idea.

# Design

In the spec of SVD, every peripheral can have a child element named `groupName`:

| child elements | description | type | occurrence |
| -|-|-|-|
|groupName | Define a name under which the System Viewer is showing this peripheral. | xs:Name | 0..1 |

The idea is to add a new argument named `--group_name_as_feature` or something shorter and meaningful and **it is off-by-default**. When the argument is given, then generate a `#[cfg(feature = "<group_name>"]` before the definition of peripheral if it has this child element.

That is if we have an SVD file with the following definition of a uart:

```xml
<peripheral>
  <name>UART0</name>
  <description>Universal Asynchronous Receivers/Transmitters</description>
  <groupName>Interface</groupName>
  <baseAddress>0x96000000</baseAddress>
  ...
```

Then the generated code is:

```rust
#[doc = "Universal Asynchronous Receivers/Transmitters"]
#[cfg(feature = "interface")]
pub struct UART0 {
    _marker: PhantomData<*const ()>,
}
#[cfg(feature = "interface")]
unsafe impl Send for UART0 {}
...
```

Also, use `toml_edit` to add cfg to `Cargo.toml`:

```diff
  [features]
  rt = ["riscv-rt"]
+ interface = []
+ group_1 = []
+ group_2 = []
+ default = ["interface", "group_1", "group_2"]
```

And the maintainer of this crate should add something like the below to the `Cargo.toml`:

```diff
[features]
rt = ["riscv-rt"]
interface = []
default = ["interface"]
+ mcu1 = ["interface", "group_1"]
+ mcu2 = ["interface", "group_2"]
```

# Possible Usecase

Suppose Alice sells two chips with most of the peripherals are same except one has HDMI and other video peripherals supported.

Bob wants to make an SVD file to describe both of them at the same time. With this feature, Bob can give HDMI a groupName such as `video` and get the following `features` in `Cargo.toml`:

```
[features]
video = []
mcu = []
mcu-with-video =["video"]
```

> Maybe this feature can be extended to support multiple arch in the same crate. But for now, I haven't got a good idea of how to implement it elegantly.

# Concern

Just like discussions in #433 and #254, adding cfg may cause usability and composability problems. Maintainers should spend more time maintaining `Cargo.toml` if they enable this feature. And the only benefit I see for now is to generate a crate that supports multiple MCU after someone merge multiple SVD file by hand.

I wonder if there are any other benefits. If the benefits exist, is it worth adopting this design?

Overall, this idea is **very similar to #254** except for using `groupName` as the cfg feature. Thus the advantages and disadvantages are alike.

Note: The implementation may be highly similar to #254, not hard but need more discussion. A try is at [duskmoon314/svd2rust group_name_as_feature](https://github.com/duskmoon314/svd2rust/tree/group_name_as_feature)

duskmoon314:
Add new feature `feature_group`
- When off, do nothing else
- When on
  - generate cfg attribute for every `groupName`
  - generate `features.toml` containing all `groupName`

close #614


duskmoon314:
> perhaps outputting a `features.toml` snippet that crate authors can copy into their Cargo.toml is simpler

I agree. Previously, I just wanted to make it easier to use. But I think it is unexpected behavior for most users when trying to use this feature. I will change it soon.

> maybe `--group-features` or `--feature-group` would be more obvious

Sure. It is indeed long. 🤣

src/util.rs
```diff
@@ -482,3 +484,15 @@ impl FullName for PeripheralInfo {
...
+
+pub fn group_names(d: &Device) -> Vec<String> {
+    let group_names: Vec<_> = d
+        .peripherals
+        .iter()
+        .filter_map(|p| p.group_name.as_ref())
+        .map(|name| name.to_sanitized_snake_case().to_string())
+        .collect::<HashSet<String>>()
+        .into_iter()
+        .collect();
```
burrbull, line 496: What sense of extra allocation?
Just return `HashSet`

burrbull: `HashSet<Cow<str>>`

duskmoon314: Oh, my mistake. Just want to remove duplicates.

src/util.rs
```diff
@@ -482,3 +484,15 @@ impl FullName for PeripheralInfo {
...
+    let group_names: Vec<_> = d
+        .peripherals
+        .iter()
+        .filter_map(|p| p.group_name.as_ref())
+        .map(|name| name.to_sanitized_snake_case())
+        .collect::<HashSet<Cow<str>>>()
+        .into_iter()
+        .collect();
+    group_names
+}
```
burrbull, line 498:
```suggestion
pub fn group_names(d: &Device) -> HashSet<Cow<str>> {
    d.peripherals
        .iter()
        .filter_map(|p| p.group_name.as_ref())
        .map(|name| name.to_sanitized_snake_case())
        .collect()
}
```

burrbull (approved):

PR #619, opened 2022-06-29
Make binary dependencies optional

### LINKED ISSUE #618: Allow to use only as library
When working with FPGA I use this crate as a build dependency library to generate code (because my FPGA configuration changes).
But this also means that I have to pull some extra deps, such as `clap`.
I would like to request to separate executable and generation library.

Shatur:
Users can now disable binary dependencies if they don't need them, closes #618.

I think we could also make JSON and YAML optional and keep them enabled by default. What do you think?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,8 +37,10 @@ path = "src/main.rs"
 required-features = ["bin"]
 
 [features]
-default = ["bin"]
+default = ["bin", "json", "yaml"]
 bin = ["dep:clap", "dep:clap_conf", "dep:env_logger"]
+json = ["dep:serde_json"]
+yaml = ["dep:serde_yaml"]
 
 [dependencies]
 cast = "0.3"
@@ -51,8 +53,8 @@ quote = "1.0"
 proc-macro2 = "1.0"
 anyhow = "1.0"
 thiserror = "1.0"
-serde_json = "1.0.79"
-serde_yaml = "0.8.23"
+serde_json = { version = "1.0.79", optional = true }
+serde_yaml = { version = "0.8.23", optional = true }
 
 [dependencies.svd-parser]
 features = ["expand"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ name = "svd2rust"
 repository = "https://github.com/rust-embedded/svd2rust/"
 version = "0.24.0"
 readme = "README.md"
+rust-version = "1.60"
 
 [package.metadata.deb]
 section = "rust"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -33,12 +33,17 @@ section = "rust"
 doc = false
 name = "svd2rust"
 path = "src/main.rs"
+required-features = ["bin"]
+
+[features]
+default = ["bin"]
+bin = ["clap", "clap_conf", "env_logger"]
```
Emilgardis, line 40: Could we maybe use the new `dep:` syntax introduced in 1.60?

You'll have to bump msrv in [README](https://github.com/rust-embedded/svd2rust#minimum-supported-rust-version-msrv) and [in ci](https://github.com/rust-embedded/svd2rust/blob/298f1f933f516302a9596c53ba89e972cda2a343/.github/workflows/ci.yml#L72)

our msrv policy is to bump when needed.

Shatur: > Does this constitute a breaking change?

I think it's not. But if we will use 1.60 feature it technically could be.

Should I doing so?

Emilgardis: Yes, but it also disables the extra features the optionals make.

without `dep:crate`, you'll have a new feature `crate`

Emilgardis (dismissed):
I like this, with or without `dep:`

asking for one more reviewer before merging

Cargo.toml
Shatur: Makes sense, will do then.

Cargo.toml
```diff
@@ -25,6 +25,7 @@ name = "svd2rust"
 repository = "https://github.com/rust-embedded/svd2rust/"
 version = "0.24.0"
 readme = "README.md"
+rust-version = "1.60"
```
Shatur, line 28: I also added [this](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) newly introduced attribute since we bumped the version to 1.60 which supports it.

Emilgardis:
> What do you think about this suggestion in a separate PR if we merge this?

if it's trivial, I think it'd be good to include it here

Shatur:
> asking for one more reviewer before merging

@Emilgardis sorry for bothering, but maybe ping someone who can left an additional review?

burrbull (approved):

Emilgardis (approved):
Thank you!

bors r=burrbull,emilgardis

Shatur:
@Emilgardis could you please draft a new release after merge?

PR #622, opened 2022-07-04
field writer's always generic

burrbull:
Alternative to #621 

Reverts [760670616ed807041767cade52518291c810177c](https://github.com/rust-embedded/svd2rust/pull/598/commits/760670616ed807041767cade52518291c810177c)

Advantages:
- easier generation logic, identical for single fields and field arrays (less potential bugs)
- easier to integrate #612
- less boilerplate, faster compilation

Disadvantages:
- as writer not attached to offsets, not obvious meaning of writer
- bigger firmware in debug mode? (need check)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Make field writer always generic around bit offset
 - Make binary dependencies optional
 - Make JSON and YAML formats optional
 - Bump MSRV to 1.60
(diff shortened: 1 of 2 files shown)


burrbull:
@newAM Could you test this one also?

newAM (approved):
This one also works

Emilgardis:
I prefer this. Does this increase size considerably in non-opt profile?

newAM:
> I prefer this. Does this increase size considerably in non-opt profile?

Tried one of my test binaries, opt-level 0.  98k with svd2rust 0.24.0, 98k with this PR.  No change on opt-level 1 either.

burrbull:
> 

0.24 is broken. It makes sense to compare only patched versions. But sounds good anyway.

@Emilgardis Let's merge this one then.

Emilgardis (approved):

PR #628, opened 2022-07-11
fix field derive in same register

### LINKED ISSUE #625: svd2rust 0.24.1 generates duplicate definitions which don't compile
Hello,
I'm trying to update from `svd2rust` 0.24.0 to 0.24.1 for https://github.com/gd32-rust/gd32-rs/, but the new version is generating invalid code which doesn't compile. You can see some of the failures at https://github.com/gd32-rust/gd32-rs/runs/7249993930?check_suite_focus=true. For example:

```
error[E0592]: duplicate definitions with name `frozen`
     --> src/gd32f130/mod.rs:63410:13
      |
63410 |             pub fn frozen(self) -> &'a mut W {
      |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ duplicate definitions for `frozen`
...
63659 |             pub fn frozen(self) -> &'a mut W {
      |             -------------------------------- other definition for `frozen`
```

The generated code in question:

```rust
#[doc = "Field `CH1COMCTL` writer - Channel 1 output compare mode"]
pub type CH1COMCTL_W<'a, const O: u8> =
    crate::FieldWriterSafe<'a, u16, CHCTL0_OUTPUT_SPEC, u8, CH1COMCTL_A, 3, O>;
impl<'a, const O: u8> CH1COMCTL_W<'a, O> {
    #[doc = "The comparison between the output compare register CHyCV and the counter CNT has no effect on the outputs"]
    #[inline(always)]
    pub fn frozen(self) -> &'a mut W {
        self.variant(CH1COMCTL_A::FROZEN)
    }
```

```rust
#[doc = "Field `CH0COMCTL` writer - Channel 0 compare output control"]
pub type CH0COMCTL_W<'a, const O: u8> =
    crate::FieldWriterSafe<'a, u16, CHCTL0_OUTPUT_SPEC, u8, CH0COMCTL_A, 3, O>;
impl<'a, const O: u8> CH0COMCTL_W<'a, O> {
    #[doc = "The comparison between the output compare register CHyCV and the counter CNT has no effect on the outputs"]
    #[inline(always)]
    pub fn frozen(self) -> &'a mut W {
        self.variant(CH0COMCTL_A::FROZEN)
    }
```

A similar issue happens in many other places as well, and for other device crates.

burrbull:
Fixes #625

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
         rust: [stable]
 
         # All vendor files we want to test on stable
-        vendor: [Atmel, Freescale, Fujitsu, Holtek, Microchip, Nordic, Nuvoton, NXP, RISC-V, SiliconLabs, Spansion, STMicro, Toshiba]
+        vendor: [Atmel, Freescale, Fujitsu, GD32, Holtek, Microchip, Nordic, Nuvoton, NXP, RISC-V, SiliconLabs, Spansion, STMicro, Toshiba]
 
         # Options are all, none, strict and const
         options: [all, none]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Fix derive field writer when link to same register
 - Use constant case for structure names; internal rearrangements for
   case conversation traits
 - Add new feature `feature_group` which will generate cfg attribute for
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -415,7 +415,7 @@ main() {
         ;;
 
         GD32)
-            test_svd_for_target cortex-m https://q.geek.nz/files/gd32f130.svd.patched
+            #test_svd_for_target cortex-m https://q.geek.nz/files/gd32f130.svd.patched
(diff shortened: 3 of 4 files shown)


burrbull:
сс @qwandor Can you make fixed link to SVDs to add them to CI?

qwandor:
I'm not quite sure what you mean by that, what exactly do you want me to do? The SVD files in https://github.com/gd32-rust/gd32-rs are extensively patched with `svdtools` before running them through `svd2rust`. The file I linked in #625 is with the patches applied.

burrbull:
> what exactly do you want me to do

Just don't delete this file.

Emilgardis:
if it's troublesome to host it, we can upload it as a gist if we're allowed

qwandor:
Okay, I'll keep it up there for now, and let you know if anything changes.

Emilgardis (dismissed):

qwandor:
Is there an advantage of 0.23.1 over 0.24.0 here?

burrbull:
> Is there an advantage of 0.23.1 over 0.24.0 here?

No advantage, but 0.24.0 is broken and generates incorrect code in some cases.

burrbull:
Close in favor of #631

Issue #636, opened 2022-07-25
Identical enum names


burrbull:
#627 made code much clearer in most of cases, but looks like it breaks on some SVDs which have different fields with identical enum names.
cc @luojia65 

For example some NXP SVDs (see https://github.com/posborne/cmsis-svd/blob/master/data/NXP/LPC178x_7x.svd) have many enumeraredValues with name `ENUM`:
![изображение](https://user-images.githubusercontent.com/3072754/180714404-9a1d5f8c-fe41-4fb4-bd8b-565b4d14ce33.png)

Emilgardis:
We need that diff tool :D

luojia65:
@burrbull: I discovered this note on [descriptions of dimArrayIndex](https://www.keil.com/pack/doc/CMSIS/SVD/html/elem_special.html#elem_dimArrayIndex):

> User is responsible for uniqueness across description.

Will this help to solve our issue here?

eddyp:
> Looks like problem is that there are multiple different enums with the same name under one register. I may suggest one of following ideas:
> 
>     1. we generate field name instead if such condition exists, or
> 
>     2. we note that same enumeratedValues name should not exist, and the users should patch their SVD files instead

Can't it be a valid solution that an enum variant with a payload which covers all the values that "map" to the same name is used?

E.g. for the SVD section below, this enum would be generated:

```rust
enum RR_INITMOD {
     mod_63,
     mod_1_63(u8),
```

```xml
            <field>
              <name>RR_INITMOD</name>
              <description>Initialization Delay Modulus</description>
              <bitOffset>16</bitOffset>
              <bitWidth>6</bitWidth>
              <access>read-write</access>
              <enumeratedValues>
                <enumeratedValue>
                  <name>MOD_63</name>
                  <description>63 cycles (same as 111111b)</description>
                  <value>0</value>
                </enumeratedValue>
                <enumeratedValue>
                  <name>MOD_1_63</name>
                  <description>1 to 63 cycles</description>
                  <value>0x1</value>
                </enumeratedValue>
                <enumeratedValue>
                  <name>MOD_1_63</name>
                  <description>1 to 63 cycles</description>
                  <value>0x2</value>
                </enumeratedValue>
                <enumeratedValue>
                  <name>MOD_1_63</name>
                  <description>1 to 63 cycles</description>
                  <value>0x3</value>
                </enumeratedValue>
                <enumeratedValue>
                  <name>MOD_1_63</name>
[...]

PR #645, opened 2022-07-27
type aliases in RegisterBlock

burrbull:


burrbull:
After this PR I'm planning to make release. So need more tests.
cc @newAM  @duskmoon314 @luojia65 
Still not sure about #636

Emilgardis (approved):

newAM (approved):
Works for me, on target tests pass: https://github.com/stm32-rs/stm32wlxx-hal/runs/7547539890?check_suite_focus=true

if anyone else is using nix you can drop into a with this version:

```bash
nix develop github:newAM/nix-shells/svd2rust-pr645#stm32-rs
```

duskmoon314:
LGFM at the first glance

I noticed that in the generated code, the code order of many of the register fields changed. Is this intentional? (Though this doesn't infect anything. I'm just curious.)

duskmoon314 (approved):

luojia65:
Looks good on my projects, it would make register block definition clear, so LGTM.

luojia65 (approved):

burrbull:
> I noticed that in the generated code, the code order of many of the register fields changed. Is this intentional? (Though this doesn't infect anything. I'm just curious.)

The cause should be that fields now sorted by bit offset

burrbull:
bors r=@Emilgardis

PR #649, opened 2022-08-22
max cluster size

burrbull:


adamgreig:
Code changes LGTM, what's the use of the max_cluster_size option?

burrbull:
> Code changes LGTM, what's the use of the max_cluster_size option?

When cluster has hole in the end it can be used to extend it size and collect clusters to array:
изображение

allexoll:
FIY this pr seems to be breaking some things over at [lpc-rs/lpc-pac](https://github.com/lpc-rs/lpc-pac/pull/75), i still need to investigate what's precisely the issue.

PR #651, opened 2022-08-24
Use `critical_section` for `Peripherals::take`.

reitermarkus:
- `cortex_m` https://github.com/rust-embedded/cortex-m/pull/447
- `msp430` https://github.com/rust-embedded/msp430/issues/13
- `riscv` https://github.com/rust-embedded/riscv/pull/110
-  `xtensa_lx` https://github.com/esp-rs/xtensa-lx/issues/20, https://github.com/esp-rs/esp-hal/pull/151
- `mips_mcu` https://github.com/kiffie/pic32-rs/issues/5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Use `critical_section::with` instead of `interrupt::free` for `Peripherals::take`.
+
 ## [v0.25.1] - 2022-08-22
 
 - Fixed parentheses in RegisterBlock field accessors
--- a/src/generate/device.rs
+++ b/src/generate/device.rs
@@ -300,31 +300,6 @@ pub fn render(d: &Device, config: &Config, device_x: &mut String) -> Result<Toke
         }
     }
 
-    let span = Span::call_site();
-    let take = match config.target {
-        Target::CortexM => Some(Ident::new("cortex_m", span)),
-        Target::Msp430 => Some(Ident::new("msp430", span)),
-        Target::RISCV => Some(Ident::new("riscv", span)),
-        Target::XtensaLX => Some(Ident::new("xtensa_lx", span)),
-        Target::Mips => Some(Ident::new("mips_mcu", span)),
-        Target::None => None,
-    }
-    .map(|krate| {
-        quote! {
-            ///Returns all the peripherals *once*
-            #[inline]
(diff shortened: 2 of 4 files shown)


src/generate/device.rs
```diff
@@ -300,48 +300,42 @@ pub fn render(d: &Device, config: &Config, device_x: &mut String) -> Result<Toke
...
+        /// All the peripherals.
         #[allow(non_snake_case)]
         pub struct Peripherals {
             #fields
         }
 
         impl Peripherals {
-            #take
+            /// Returns all the peripherals *once*.
+            #[cfg(feature = "critical-section")]
```
Emilgardis, line 318: This means a new dependency, the crate documentation should be adjusted to account for this.


also, could https://github.com/rust-embedded/svd2rust/blob/a39a61d8ac8928a59d8a2b855c3303a766fd0517/ci/svd2rust-regress/src/svd_test.rs#L8 be updated to include `critical-section`

Emilgardis (dismissed):
lgtm, this is great :D

do we need another reviewer maybe?

adamgreig:
I'm a bit concerned about what this means for platforms that don't have `critical-section` support yet. I know msp430 is working on it but having some codegen size issues, and I don't know anything about `mips_mcu`. It might be worth waiting a little longer for those architectures to get support before releasing svd2rust that uses it?

Having it as an optional dependency/feature seems useful though and does mean those platforms don't need immediate critical-section support, even end-users could add simple interrupt-based implementations to use `take()` if needed.

reitermarkus:
> It might be worth waiting a little longer

I don't think it's worth waiting. Other architectures will have to implement these only once they upgrade `svd2rust` anyways. I have opened issues for the affected architectures to give them a heads up though.

adamgreig:
Oh, thanks, I didn't see the edit to the first comment. SGTM to me then.

Emilgardis (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Use `critical_section::with` instead of `interrupt::free` for `Peripherals::take`.
```
Emilgardis, line 10: A mention to `take` not being provided if `critical-section` is not enabled could probably be added here

therealprof (approved):
LGTM,too. Thanks a lot @reitermarkus for taking this on.

bors r+

PR #656, opened 2022-09-06
clap4 & irx-config

burrbull:


therealprof:
Looks okay in principle although I'm not super-psyched to rely on a crate which had a total of 453 downloads until now, a good fraction of which caused by this CI...

src/util.rs
```diff
@@ -22,23 +22,47 @@ pub const BITS_PER_BYTE: u32 = 8;
 /// that are not valid in Rust ident
 const BLACKLIST_CHARS: &[char] = &['(', ')', '[', ']', '/', ' ', '-'];
 
+#[cfg_attr(feature = "bin", derive(serde::Deserialize))]
```
Emilgardis, line 25: can we cfg this on `serde` instead

burrbull: I've tried to use `serde`, but it doesn't compile. I don't know why.

Emilgardis: it's probably because it's used as `dep:serde`, instead only do `serde` in the `features.bin = [...]`

Emilgardis (dismissed):
looks good to me

The concern with popularity/usage of the `irx-config` is valid, but I think we can go ahead with this.

@therealprof what do you think?

burrbull:
cc @Emilgardis Updated to clap 4.0

therealprof:
@Emilgardis Whoops, sorry, didn't notice you left a comment a while ago. Yes, fine by me.
@burrbull If you could resolve the conflict...

luojia65:
LGTM, maybe I should close my PR then...

burrbull:
rebased

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Update `clap` to 3.2, use `irx-config` instead of `clap_conf`
```
luojia65, line 10: Should this be "update clap to 4.0"? :)

therealprof (approved):
Let's get the party going then...

bors r+

Issue #657, opened 2022-09-27
While compiling got error[E0659]: `core` is ambiguous --> missing self:: when reexporting modules


pellico:
Peripheral could have names that conflict with crates names.

### How to reproduce: 

- Use attached SVD file. [cyt3bb.zip](https://github.com/rust-embedded/svd2rust/files/9657463/cyt3bb.zip)
- Generate using svd2rust 0.25.1 ( )
- `cargo check`

### Error
```bash
error[E0659]: `core` is ambiguous
     --> src\lib.rs:73622:13
      |
73622 |     pub use core::CORE;
      |             ^^^^ ambiguous name
      |
      = note: ambiguous because of multiple potential import sources
      = note: `core` could refer to a built-in crate
      = help: use `::core` to refer to this crate unambiguously
note: `core` could also refer to the module defined here
     --> src\lib.rs:73625:5
      |
73625 | /     pub mod core {
73626 | |         #[doc = r"Register block"]
73627 | |         #[repr(C)]
73628 | |         pub struct CORE {
...     |
81671 | |         }
81672 | |     }
      | |_____^
      = help: use `self::core` to refer to this module unambiguously

For more information about this error, try `rustc --explain E0659`.
error: could not compile `cyt3bb` due to previous error
```
### Workaround
Replace manually `pub use core::CORE;` with `pub use self::core::CORE;`

PR #658, opened 2022-09-28
Fixing compilation error when re-exporting module with same name as a Rust package.

pellico:
Added self:: when referencing submodule to avoid ambiguity fix issue https://github.com/rust-embedded/svd2rust/issues/657


burrbull:
LGTM.
bors r+

burrbull (approved):

burrbull:
bors retry

PR #667, opened 2022-10-12
bin: update dependency `clap` to version 4.0

luojia65:
This pull request updates dependency `clap` svd2rust binary from `2.0` to `4.0`.

It has been years since clap has more user friendly interfaces and more rich ecosystem around it. In this version, we adpot to modern `clap-verbosity-flag` to describe a log level (not `--log_level <LEVEL>`, but `-v`, `-vv` etc., see https://docs.rs/clap-verbosity-flag). The default cases of command lines is changed to kebab-case in modern clap 4.0, other than snake_case in older versions. Other than that, no changes to command line parameters as well as their help messages are applied in this pull request.

The crate `clap_conf` is written back to 2021 and lacks people to maintain. It's not good to use it in frequently used embedded Rust toolchain, I rewrote it with a `toml` and `serde` based toml file parser to adapt to older version toml based toolchain configurations. If this feature should not be used in the future, we may remove this part of code.

The command line helper after this pull request:

图片

No changes to library code is included in this pull request.

For reviewers: if this pull request would include any unacceptible breaking changes in command line, please note here and I will modify code in this pull request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Update dependency `clap` to version 4.0
+
 ## [v0.26.0] - 2022-10-07
 
 - Use edition 2021
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,16 +38,18 @@ required-features = ["bin"]
 
 [features]
 default = ["bin", "json", "yaml"]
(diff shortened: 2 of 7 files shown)


Emilgardis:
see https://github.com/rust-embedded/svd2rust/pull/656 which does the same thing (regarding clap 4.0)

Emilgardis:
with #656 being non breaking, I think that's the solution we should go with

I would personally like to have the flags as both snake and camel case though

therealprof:
@Emilgardis I agree.

PR #676, opened 2022-10-22
test patched STM32

burrbull:


therealprof:
Excuse my ignorance but what are those pached files?

adamgreig:
They're the patched SVDs from stm32-rs (in other words, with all the stm32-rs fixes applied, instead of the raw buggy ST SVDs). Looks like it downloads the current build output from the master branch.

burrbull:
cc @Emilgardis

Emilgardis (approved):
bors r+

PR #677, opened 2022-10-22
inline set/clear_bit

burrbull:
less size in `dev` mode


therealprof (approved):
LGTM, thanks.

bors r+

PR #680, opened 2022-10-24
docs: brief intro to `critical-section`

### LINKED ISSUE #679: The usage of `take()` and `critical-section` should be added
After merging #651, `Peripherals::take()` is not callable unless the user of pac provides an implementation. So we should add how to use it in the document.

---

## Usage I have tried

> The following part is mentioned in the matrix chatting room.

I have tried some ways and found two ways of using `take()`:

### 1 User provide implementation

Users can add implement on their own or use the implementation provided by the target or hal crate. An example will be like this:

```toml
# Cargo.toml
# version and other info are ignored
some-pac = { features = ["critical-section"] }
riscv = { features = ["critical-section-single-hart"] }
```

Then adding a dummy use:

```rust
use rust as _;
```

### 2 Provide implementation in PAC

We can also provide the implementation in PAC through the target crate.

```
// lib.rs
#[cfg(feature = "critical-section-single-hart")]
use riscv as _;
```

```
# Cargo.toml
[features]
critical-section-single-hart = ["critical-section", "riscv/critical-section-single-hart"]
```

Then users can add pac like this:

```
some-pac = { features = ["critical-section-single-hart"] }
```

## Discussion

There may be other usages that I haven't found out. I think it may not be PAC's purpose to provide the implementation. So there should have a section of this in the documentation.

duskmoon314:
May need to add more details

close #679

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Make writing raw bits to a whole register safe if the SVD indicates
   so through the <WriteConstraint> element (see [v0.7.1] too).
 - Remove lint #![deny(const_err)] as it is a hard error in Rust now
+- Add doc of using `critical-section`
 
 ## [v0.26.0] - 2022-10-07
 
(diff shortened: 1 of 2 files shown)


burrbull (approved):

burrbull:
Thanks
bors r+

PR #675, opened 2022-10-21
use_cast

burrbull:
Fixes #671

PR #682, opened 2022-10-25
Not cast if unnecessary

duskmoon314:
In the previous PR #675, some cases of unnecessary casting are eliminated. However, if there is a big field (bigger than 16) with offset, there is still unnecessary generated casting. This PR fixes this case and adjusts the code according to clippy.


burrbull (approved):

burrbull:
update changelog, please

duskmoon314:
I just updated CHANGELOG and squashed commits.

burrbull (approved):

PR #685, opened 2022-10-27

YuhanLiin:
The current generated code for MSP430 atomics is not compatible with the rest of the generated code. This PR fixes this.


burrbull:
could you add ci tests to prevent this issue in future?

also update changelog

burrbull:
@YuhanLiin I've rebased. See https://github.com/rust-embedded/svd2rust/compare/fix-msp430-atomic

PR #687, opened 2022-10-29

YuhanLiin:
Rebased from #685


burrbull:
remove --strict flag

YuhanLiin:
I've moved the PR to #688, which uses a branch from my fork, since I can actually push to it.

PR #688, opened 2022-10-29

YuhanLiin:
Rebased from #685

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,7 @@ jobs:
           - { rust: 1.60.0, vendor: Nordic, options: "" }
           # Use nightly for architectures which don't support stable
           - { rust: nightly, vendor: MSP430, options: "--nightly" }
+          - { rust: nightly, vendor: MSP430, options: "" }
           - { rust: nightly, vendor: Espressif, options: "" }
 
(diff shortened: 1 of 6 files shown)


burrbull:
SVD link 404 error

burrbull (approved):

YuhanLiin:
Note: Now that I've switched the atomic library over to `portable-atomic`, we might be able to port this feature over to other targets.

burrbull:
> Note: Now that I've switched the atomic library over to `portable-atomic`, we might be able to port this feature over to other targets.

it sounds interesting

adamgreig:
I haven't compared them in detail, but for other targets it might make more sense to use [`atomic-polyfill`](https://crates.io/crates/atomic-polyfill) which builds on `critical-section` and is quite widely used in other embedded Rust projects. `portable-atomic` can only use a global disable-interrupts to simulate CAS operations (when enabled with `--cfg portable_atomic_unsafe_assume_single_core`), while `atomic-polyfill` uses `critical-section` which allows pluggable implementations (and is used in svd2rust for `Peripherals::take()`).

At the moment I don't think `atomic-polyfill` supports MSP430 at all, so it's not an option here (yet?).

PR #692, opened 2022-10-30
accessors everywhere (breaking)

burrbull:
All register/cluster access through methods. RegisterBlock fields are private

To access register or cluster instead of:
```rust
per.reg.write(...)
per.regarray[2].write(...)
per.altreg().write(...)
```
use
```rust
per.reg().write(...)
per.regarray(2).write(...)
per.altreg().write(...)
```


duskmoon314:
What will it be like after this PR? From the brief intro, I imagine this:

```rust
// error
uart.thr.write(...);

// correct
uart.thr().write(...);
```

Is this the expected behavior?

burrbull:
Yes

luojia65:
Should we keep register blocks public under some feature gate? Rust kernel developers may still reuse this structure when writing drivers, as in this purpose the base address of peripherals are not fixed as bare metal when MMU comes in.

duskmoon314:
> Should we keep register blocks public under some feature gate? Rust kernel developers may still reuse this structure when writing drivers, as in this purpose the base address of peripherals are not fixed as bare metal when MMU comes in.

It seems this PR only makes the `field`s private and leaves `RegisterBlock` as it is. If so, I think there is no problem to use with a virtual address.

burrbull:
I don't insist on this PR. It is just concept for discuss.

burrbull:
cc @rust-embedded/tools

adamgreig (approved):
It's a big breaking change to the syntax, but I think it should be a fairly easy change for users to make in their code, and importantly it moves us to the syntax we'll need to get rid of memory-mapped structs. I can't see any way to keep the current syntax but not have memory-mapped structs.

I guess if someone needs a pointer to a field, they can still call `as_ptr()` on what the accessor returns?

burrbull:
I guess if someone needs a pointer to a field, they can still call as_ptr() on what the accessor returns?
Should be same as before.

PR #662, opened 2022-10-06
Disjoint arrays

n8tlarsen:
Solution 2 for #660 which pre-parses ERC type names and provides expansion information to the render functions. Works great with a few SVDs from the [Renesas RA DFP](https://www2.renesas.eu/Keil_MDK_Packs/Renesas.RA_DFP.4.0.0.pack) which tend to use disjoint arrays quite frequently.

PR #698, opened 2022-11-21
Compare contents before deriving

### LINKED ISSUE #660: Handling disjoint arrays
I've run into an svd file where a set of registers is allocated linearly in memory, but only specific registers can be used. For example the span 0-31 is allocated for a register type, but only 0-3, 8, 9, 12, 14, 15 and 18 can be used. See ELSR* registers in [R7FA2E1A9.svd](https://github.com/rust-embedded/svd2rust/files/9686850/R7FA2E1A9.svd.txt) from the [Renesas RA DFP](https://www2.renesas.eu/Keil_MDK_Packs/Renesas.RA_DFP.4.0.0.pack). When parsed by svd2rust version 0.25.1, this creates new types for each register grouping, often causing name collisions later when compiling the crate. I can see two ways to handle this:

1. Create a continuous array including the dead addresses. This makes for a simpler API but may lead to undefined behavior when trying to access the dead addresses.
2. Create one register for each live address with a unique name that shares the same type. This would be safer, but the API isn't quite as ergonomic to use.

I lean towards option two for the safety reasons, but am interested in other options for handling this use case.

n8tlarsen:
Closes #660. Compares register properties and fields of regex-matched register names to decide if implicitly deriving is possible. I could use some help with the CI integration for vendor Renesas. See [ra-rs/ra](https://github.com/ra-rs/ra) for SVDs.

### DIFF
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -6,7 +6,10 @@ use svd_parser::expand::{
     derive_cluster, derive_peripheral, derive_register, BlockPath, Index, RegisterPath,
 };
 
-use crate::svd::{array::names, Cluster, ClusterInfo, Peripheral, Register, RegisterCluster};
+use crate::svd::{
+    array::names, Cluster, ClusterInfo, MaybeArray, Peripheral, Register, RegisterCluster,
+    RegisterInfo,
+};
 use log::{debug, trace, warn};
 use proc_macro2::{Ident, Punct, Spacing, Span, TokenStream};
 use quote::{quote, ToTokens};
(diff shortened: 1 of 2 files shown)


src/generate/peripheral.rs
```diff
@@ -864,29 +838,103 @@ fn find_root(dpath: &str, path: &BlockPath, index: &Index) -> Result<RegisterPat
...
+fn is_derivable(
+    source_reg: &MaybeArray<RegisterInfo>,
+    target_reg: &MaybeArray<RegisterInfo>,
+) -> bool {
+    if (source_reg.properties == target_reg.properties) && (source_reg.fields == target_reg.fields)
+    {
+        true
+    } else {
+        false
+    }
```
Emilgardis, line 937:
```suggestion
    source_reg.properties == target_reg.properties) && (source_reg.fields == target_reg.fields
```

burrbull:
I forgot about this PR.
Could you update changelog?

n8tlarsen:
Since there hasn't been a release since #662 should I amend the existing changelog line, or add a new line?

burrbull:
> Since there hasn't been a release since #662 should I amend the existing changelog line, or add a new line?

As you wish.

burrbull (approved):

Emilgardis (approved):
awesome!

bors r=burrbull,emilgardis

PR #711, opened 2023-02-10
Escape <> and & symbol in doc attributes.

### LINKED ISSUE #710: Warning during doc generation: warning: unclosed HTML tag `K`
When generating documentation for generated PAL I get:
`warning: unclosed HTML tag `K`
  --> src\srss\res_cause2.rs:47:5
   |
47 |     #[doc = "Bits 0:15 - Clock supervision logic requested a reset due to loss or frequency violation of a high-frequency clock. Each bit index K corresponds to a HFCLK<K>. Unimplemented clock bits return zero."]`

The reason is clearly are the angle brackets. Looking at CMSIS specs nothing forbid to use angle brackets. Moreover the generated documentation has bad formatting due to embedded tag.

I expect that svd2rust escapes this kind of characters.

svd2rust version 0.28.0
In attachment one of SVD files that triggers the issue.
[cyt2b9.zip](https://github.com/rust-embedded/svd2rust/files/10705495/cyt2b9.zip)

pellico:
Close #710 
The fix escapes only very basic chars <> and & to be as much less intrusive.
The fix could be implement without a new dependencies but more escaping may be required in the near future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,7 @@ jobs:
           - { rust: stable, vendor: Toshiba, options: all }
           - { rust: stable, vendor: Toshiba, options: "" }
           # Test MSRV
-          - { rust: 1.64.0, vendor: Nordic, options: "" }
+          - { rust: 1.65.0, vendor: Nordic, options: "" }
           # Use nightly for architectures which don't support stable
           - { rust: nightly, vendor: MSP430, options: "--atomics" }
           - { rust: nightly, vendor: MSP430, options: "" }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 - Bump MSRV to 1.64
 - Fix dangling implicit derives
+- Fix escaping <> and & characters in doc attributes
 
 ## [v0.28.0] - 2022-12-25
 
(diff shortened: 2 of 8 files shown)


Emilgardis:
Are there cases where we actually want html? Perhaps the description could be using html for nicer formatting

pellico:
> Are there cases where we actually want html? Perhaps the description could be using html for nicer formatting

I  am working for the company that generated this SVD and there is no intention to use html. It is just how HW concept engine described a clock signal "HFCLK<K>". 
In principle we can escape the characters in the SVD comments but this will create a problem for debugger tools. In fact in CMSIS specs html support is not mentioned [CMSIS spec](https://www.keil.com/pack/doc/CMSIS/SVD/html/elem_registers.html#elem_cluster). 
image

So we would like to avoid to have two SVD one for debuggers and one SVD2Rust.

If default escaping is not acceptable, will a command line option be acceptable ?

Emilgardis:
Alright, this is fine then, a string is not html necessarily html :)

The ci error is due to let else being a 1.65 feature, we can bump the minimally supported rust version (MSRV)

pellico:
Thank you.
Let me know if I have to do anything else.

Emilgardis (approved):
Thanks!

bors r+

burrbull:
@Emilgardis 
Looks like this PR causes #849.

PR #712, opened 2023-03-02
Replace cluster suffixes with _

n8tlarsen:
In some devices, a cluster array has registers named the same as the cluster. So when making a Rust array out of the cluster, the PAC failed to compile due to a recursive definition. For example, see [ATSAME51G18A.svd](https://github.com/atsamd-rs/atsamd/blob/master/svd/ATSAME51G18A.svd#L11590), CHANNEL[%s] cluster definition on line 11593.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - Bump MSRV to 1.65
 - Fix dangling implicit derives
 - Fix escaping <> and & characters in doc attributes
+- Replace cluster suffixes with _
 
 ## [v0.28.0] - 2022-12-25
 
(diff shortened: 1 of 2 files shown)


burrbull:
It is awfullest idea I've seen.

n8tlarsen:
1. Please explain your reasoning.
2. Propose a better solution.

burrbull:
> 1. Please explain your reasoning.

Due to several bad designed SVDs you suggest to add **breaking change** which even looks non unaesthetic.

>     2. Propose a better solution.

1. Fix or preprocess SVDs (easiest)
2. Move this change under option
3. Add options for **custom prefixes and suffixes** for names of enums, clusters, etc. (flexible)

n8tlarsen:
I think in this particular case, only one register is affected, so a patch was applied to remedy the SVD name collision. Thanks for your time.

Issue #48, opened 2016-12-27
Add Debug trait implementation


newpavlov:
It would be very helpful to have a `Debug` traits implemented for register structs. Currently if I am debugging some code or interested in how things work I simply add something like [this](https://gist.github.com/anonymous/6da8dd5b9dabce394ce9f8a7939b3c84) to memory map crate. It will be mainly useful for read and memory block structs, but some may find it usefull for write structs too.

japaric:
> It would be very helpful to have a Debug traits implemented for register structs.

Are you serializing/printing the registers through UART/ITM? I wonder if this could be handled with GDB pretty printers, then:

```
(gdb) print block.register
{
    fooen: true,
    baren: false,
    ..
}
```

would Just Work.

In any case, I'm OK with adding `Debug` implementations to `RegisterR`s. The implementation should use the ["formatting structs"](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct) though, then we get pretty printing (`{#?}`) for free.

newpavlov:
I am using ITM. If I am not mistaken this will not work nicely with projects compiled with `--release` (at least with default options). Also it will be friendlier for user to just use several `iprinln!`'s to see how registers change without any additional typing in the console.

"formatting structs" will be a great way to do it.

hannobraun:
I assume this issue refers to the `RegisterBlock` structs. In addition, it would be useful to also have `Debug` implementations for the peripheral structs that dereference to them (the one's with the `ptr` method).

burrbull:
Fixed by #716

PR #716, opened 2023-05-16
implement Debug for readable fields of registers

liebman:
contributes towards #48 

Notes:
- I've only run this with esp (esp32/esp32s3) pacs
- I'd love some feedback on where I've implemented this and if there is a cleaner way

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -230,9 +230,13 @@ pub fn render_register_mod(
     }
 
     let mut r_impl_items = TokenStream::new();
+    let mut r_debug_impl = TokenStream::new();
     let mut w_impl_items = TokenStream::new();
     let mut zero_to_modify_fields_bitmap = 0;
     let mut one_to_modify_fields_bitmap = 0;
+
+    let open = Punct::new('{', Spacing::Alone);
+    let close = Punct::new('}', Spacing::Alone);
 
     if let Some(cur_fields) = register.fields.as_ref() {
         // filter out all reserved fields, as we should not generate code for
@@ -243,6 +247,15 @@ pub fn render_register_mod(
             .collect();
 
         if !cur_fields.is_empty() {
+            if config.impl_debug {
+                r_debug_impl.extend(render_register_mod_debug(
+                    register,
+                    &access,
+                    &cur_fields,
+                    config,
+                ))
+            }
+
             (
                 r_impl_items,
                 w_impl_items,
@@ -261,14 +274,55 @@ pub fn render_register_mod(
                 config,
             )?;
         }
+    } else if !access.can_read() || register.read_action.is_some() {
+        if let Some(feature) = &config.impl_debug_feature {
+            r_debug_impl.extend(quote! {
+                #[cfg(feature=#feature)]
+            });
+        }
+        r_debug_impl.extend(quote! {
+            impl core::fmt::Debug for crate::generic::Reg<#regspec_ident> {
+                fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+                    write!(f, "(not readable)")
+                }
+            }
+        });
+    } else {
+        // no register fields are defined so implement Debug to get entire register value
+        if let Some(feature) = &config.impl_debug_feature {
+            r_debug_impl.extend(quote! {
+                #[cfg(feature=#feature)]
+            });
+        }
+        r_debug_impl.extend(quote! {
+            impl core::fmt::Debug for R {
+                fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+                    write!(f, "{}", self.bits())
+                }
+            }
+        });
+        if let Some(feature) = &config.impl_debug_feature {
+            r_debug_impl.extend(quote! {
+                #[cfg(feature=#feature)]
+            });
+        }
+        r_debug_impl.extend(quote! {
+            impl core::fmt::Debug for crate::generic::Reg<#regspec_ident> {
+                fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+                    self.read().fmt(f)
+                }
+            }
+        });
     }
-
-    let open = Punct::new('{', Spacing::Alone);
-    let close = Punct::new('}', Spacing::Alone);
 
     if can_read && !r_impl_items.is_empty() {
         mod_items.extend(quote! {
             impl R #open #r_impl_items #close
+        });
+    }
+    if !r_debug_impl.is_empty() {
+        mod_items.extend(quote! {
+            #r_debug_impl
         });
     }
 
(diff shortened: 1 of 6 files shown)


src/generate/register.rs
```diff
@@ -243,6 +247,54 @@ pub fn render_register_mod(
...
+                        Some(a) => a,
+                        None => access,
+                    };
+                    let bit_or_bits = if f.bit_width() > 1 {
+                        "bits"
+                    } else {
+                        "bit"
+                    }.to_ascii_lowercase();
+                    let bit_or_bits = syn::Ident::new(&bit_or_bits, span);
+                    log::info!("register={} field={}", reg_name, f.name);
```
burrbull, line 271: maybe `log::debug` of even `trace`

src/generate/register.rs
```diff
@@ -243,6 +247,54 @@ pub fn render_register_mod(
...
+                for &f in cur_fields.iter() {
+                    let field_access = match f.access {
+                        Some(a) => a,
+                        None => access,
+                    };
+                    let bit_or_bits = if f.bit_width() > 1 {
+                        "bits"
+                    } else {
+                        "bit"
+                    }.to_ascii_lowercase();
```
burrbull, line 269: ?

src/generate/register.rs
```diff
@@ -263,13 +315,13 @@ pub fn render_register_mod(
...
-    let open = Punct::new('{', Spacing::Alone);
-    let close = Punct::new('}', Spacing::Alone);
-
     if can_read && !r_impl_items.is_empty() {
         mod_items.extend(quote! {
             impl R #open #r_impl_items #close
         });
+        mod_items.extend(quote!{
+            #r_debug_impl
+        });
```
burrbull, line 324: This should not be added by default. Need new option:
```suggestion
        if config.register_core_debug {
            mod_items.extend(quote!{
                #r_debug_impl
            });
        }
```

liebman: @burrbull  - This adds `impl Debug` for read registers gated by `#[cfg(feature="impl-register-debug")]` in the resulting pac so that users of the pac can choose to have debug or no debut without having to generate the pac themselves.

src/generate/register.rs
liebman: moved to debug

src/generate/register.rs
liebman: removed extraneous `to_ascii_lowercase()`

src/generate/register.rs
burrbull: This significantly grows `pac` codebase with dead code for most of users. So no, this should not be generated by default.

Also we should have possibility to change core `Debug` with `defmt::Format` which looks more usable in this case.

liebman: I've added options to enable Debug and optionally specify a gating feature.

As for `defmt::Format` I've not worked with `defmt` and could use some hints on how you think the data should be formatted?  For Debug I used the `debug_struct` in `core::fmt::Formatter`.

burrbull:
I've missed that you targeting esp32. Do I understand right that you don't have other way to debug except this?
IDE extensions like `probe-rs` or `Cortex-m-Debug` for VSCode?

liebman:
Cortex debug can work easily for newer esp32 family chips (esp32s3, esp32c3, etc) that have builtin usb - for those without builtin usb its only available if the JTAG pins are available. 

One specific use case here is that many of the peripherals (for me this is RTC_CTRL power management) are not well documented.  And since there is a std hal for esp32 that is built on esp-idf one thing I can do is dump peripherals registers on the std hal to help with implementations of the no-std hal by being able to compare what registers are set differently.  After manually building this for esp32 I was starting on esp32s3 and realized this would be much better placed in `svd2rust` and could be automatically generated - thats when I came across #48.

liebman:
What I really want is to be able to dump all registers from the peripheral level as registers (and fields) differ a fair amount between esp32 chip types. (but at least having the registers handled automatically is a huge help.

burrbull:
> What I really want is to be able to dump all registers from the peripheral level as registers (and fields) differ a fair amount between esp32 chip types. (but at least having the registers handled automatically is a huge help.

If you want to read a range of registers, you should also decide what to do with registers which have read side effects (see `readAction`).

burrbull:
Don't do `merge master`. Use `rebase` instread. We want to have linear commit history.

liebman:
I'm already skipping registers that have read side effects.  Currently not even generating the Debug implementation but maybe I should and just print the register name and no value or something short indicating it was not read.

liebman:
> Don't do `merge master`. Use `rebase` instread. We want to have linear commit history.

How can I fix this?

liebman:
I've added debug at the RegisterBlock level so one can print all registers belonging to a peripheral with:
```
    info!("SHA {:#?}", unsafe { &*esp::SHA::ptr() });
```
Write only registers are printed with a note that they are read only and registers that have read actions are also printed with a message.  All readable registers have their fields printed in struct format and if there are no fields then the raw register value is printed.  Register clusters "should" also be handled.

burrbull (approved):

burrbull:
Looks good for me.
@Emilgardis ?

burrbull:
@liebman Could you how some usage example output?

src/util.rs
```diff
@@ -51,6 +51,10 @@ pub struct Config {
     pub feature_peripheral: bool,
     #[cfg_attr(feature = "serde", serde(default))]
     pub max_cluster_size: bool,
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub impl_register_debug: bool,
```
Emilgardis, line 55: I'd like it if this field took a enum instead, which would make it possible to use defmt instead in the future.

Ofcourse, this can be added later with a breaking change

src/main.rs
```diff
@@ -118,6 +118,19 @@ fn run() -> Result<()> {
...
+        .arg(
+            Arg::new("impl_register_debug")
+                .long("impl_register_debug")
+                .action(ArgAction::SetTrue)
+                .help("implement Debug for readable registers"),
+        )
+        .arg(
+            Arg::new("register_debug_feature")
+                .long("register_debug_feature")
+                .help("add feature gating for register debug implementation")
```
Emilgardis, line 130: I think this should be renamed to 

```
impl_debug_feature
```

src/main.rs
```diff
@@ -118,6 +118,19 @@ fn run() -> Result<()> {
                 .action(ArgAction::SetTrue)
                 .help("Use array increment for cluster size"),
         )
+        .arg(
+            Arg::new("impl_register_debug")
+                .long("impl_register_debug")
+                .action(ArgAction::SetTrue)
+                .help("implement Debug for readable registers"),
+        )
+        .arg(
```
Emilgardis, line 127: this also includes more than just registers now (blocks specifically), maybe rename it to

```
impl_debug
```

Emilgardis (approved):
this is good as is, just some improvements on naming and suggestion for the future

liebman:
@burrbull  Here is a (truncated) sample run on an esp32s3 using `println!("RTC_CNT {:#?}", unsafe { &*esp32s3::RTC_CNTL::ptr() });`
```
RTC_CNT RegisterBlock {
    options0: OPTIONS0 {
        sw_stall_appcpu_c0: 0,
        sw_stall_procpu_c0: 0,
        bb_i2c_force_pd: false,
        bb_i2c_force_pu: false,
        bbpll_i2c_force_pd: false,
        bbpll_i2c_force_pu: false,
        bbpll_force_pd: false,
        bbpll_force_pu: false,
        xtl_force_pd: false,
        xtl_force_pu: false,
        xtl_en_wait: 2,
        xtl_force_iso: false,
        pll_force_iso: false,
        analog_force_iso: false,
        xtl_force_noiso: true,
        pll_force_noiso: true,
        analog_force_noiso: true,
        dg_wrap_force_rst: false,
        dg_wrap_force_norst: false,
    },
    slp_timer0: SLP_TIMER0 {
        slp_val_lo: 0,
    },
    slp_timer1: SLP_TIMER1 {
        slp_val_hi: 0,
    },
    time_update: TIME_UPDATE {
        timer_sys_stall: false,
        timer_xtl_off: false,
        timer_sys_rst: false,
    },
    time_low0: TIME_LOW0 {
        timer_value0_low: 46911,
    },
    time_high0: TIME_HIGH0 {
        timer_value0_high: 0,
    },
    state0: STATE0 {
        apb2rtc_bridge_sel: false,
        sdio_active_ind: false,
        slp_wakeup: true,
        slp_reject: false,
        sleep_en: false,
    },
<snip>
```

liebman:
@Emilgardis I like the naming suggestions and will try to update examples and documentation.

src/util.rs
liebman: that or add another bool option for impl_defmt allowing both to be specified.

src/generate/register.rs
```diff
@@ -261,16 +274,70 @@ pub fn render_register_mod(
...
+    } else if register.read_action.is_some() {
+        if let Some(feature) = &config.impl_debug_feature {
+            r_debug_impl.extend(quote! {
+                #[cfg(feature=#feature)]
+            });
+        }
+        r_debug_impl.extend(quote! {
+            impl core::fmt::Debug for crate::generic::Reg<#regspec_ident> {
+                fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+                    write!(f, "(not read/has read action!)")
```
Emilgardis, line 286:
```suggestion
                    write!(f, "(not readable)")
```

src/generate/register.rs
```diff
@@ -386,6 +453,108 @@ pub fn render_register_mod(
...
+        }
+        r_debug_impl.extend(quote! {
+            impl core::fmt::Debug for crate::generic::Reg<#regspec_ident> {
+                fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+                    write!(f, "(write only register)")
+                }
+            }
+        });
+    } else {
+        warn!("not implementing debug!!! {register:#?}");
```
Emilgardis, line 553:
```suggestion
        warn!("not implementing debug for {name}");
```

src/lib.rs
```diff
@@ -495,13 +495,29 @@
...
+//! ## the `--impl_debug` flag
+//!
+//! The `--impl_debug` option will cause svd2rust to generate `core::fmt::Debug` implementations for
+//! all registers and blocks.  If a register is readable and has fields defined then each field value
+//! will be printed - if no fields are defined then the value of the register will be printed. Any
+//! register that has read actions will not be read and printed as `(not read/has read action!)`.
+//! Registers that are not readable will have `(write only register)` printed as the value.
+//!
+//! The `--impl_debug_feature` flad can also be specified to compile debug implementations conditionally
+//! based on the supplied feature name.
```
Emilgardis, line 507:
```suggestion
//! The `--impl_debug_feature` flad can also be specified to include debug implementations conditionally
//! behind the supplied feature name 
```

Emilgardis (changes requested):

src/generate/register.rs
Emilgardis: I'd like it better if the register didn't include a debug impl if it never could be printed, but that would mean having to do a custom debug impl for blocks

liebman: Yea - and thats the way I started off but then its not consistant for the developer who maybe tries to print a register contents and its different for write-only vs readable.  Main reason was making the logic work in peripherals.rs so that it exactly matched what is in registers.rs for registers/arrays/readable/read-action/write-only so I punted and just implemented Debug for all - it was much easier.

liebman: I'm not sure about `(not readable)` as thats not correct, how about `(has read action)`

liebman: or `has read side effects`

liebman: maybe even just `(side effects)`?

liebman: actually - what if we dropped one of the else if conditions and just printed "not readable" for write only or read action?

Emilgardis: im ok with any of these 👍🏼

liebman: Done

Emilgardis (approved):
awesome! thank you :)

I think we can merge this, I'll leave the final r+ for @burrbull

Emilgardis:
Might be good to squash some of the commits btw!

liebman:
squashed

burrbull (approved):

liebman:
Is there a date planned for the next release?

PR #718, opened 2023-05-19
feat: configurable link section attribute for irqs

sethp:
This change introduces a new config field that allows `svd2rust` to target which linker sections get assigned to the `__INTERRUPTS` static, with reasonable defaults.

Previously on RISC-V, the choice was always left up to the compiler, and it seemed to always pick `.rodata`. Unfortunately, in my context, that meant placing the LUT in a memory range that had a lot of highly variable latency, which cost not just time but predictability in servicing interrupts.

With this change in place, I'm able to target a particular section (e.g. `.data`, or `.trap.rodata`) for the placement of the static, which grants more granular control over the ultimate loaded memory address.

For the full details about the problem, please see: https://github.com/esp-rs/esp-hal/pull/534/commits/e29f3d547dc210e1b73313be6053a2122239a467

### DIFF
--- a/src/generate/interrupt.rs
+++ b/src/generate/interrupt.rs
@@ -188,13 +188,11 @@ pub fn render(
                 writeln!(device_x, "PROVIDE({name} = DefaultHandler);")?;
             }
 
-            let link_section_attr = if let Some(section) = &config.interrupt_link_section {
+            let link_section_attr = config.interrupt_link_section.as_ref().map(|section| {
                 quote! {
                     #[link_section = #section]
                 }
-            } else {
-                quote! {}
-            };
+            });
 
             root.extend(quote! {
                 #[cfg(feature = "rt")]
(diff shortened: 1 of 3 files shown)


src/generate/interrupt.rs
```diff
@@ -172,6 +188,14 @@ pub fn render(
                 writeln!(device_x, "PROVIDE({name} = DefaultHandler);")?;
             }
 
+            let link_section_attr = if let Some(section) = &config.interrupt_link_section {
+                quote! {
+                    #[link_section = #section]
+                }
+            } else {
+                quote! {}
+            };
```
burrbull, line 197:
```suggestion
            let link_section_attr = config.interrupt_link_section.as_ref().map(|section| {
                quote! {
                    #[link_section = #section]
                }
            });
```
Same in 225.

burrbull:
I hope  this is backward compatible?

burrbull (dismissed):

sethp:
Oh, great point: the code generator changes are in the sense that passing `None` for that parameter (or not setting a value, which should default to `None`) doesn't change the generated code.

Using the new parameter may not be backwards compatible, depending on what link section is being set: as long as the new section gets output in some part of the resulting binary it sure could be, but it's somewhat easy to rely on a particular layout, either accidentally or on purpose.

src/generate/interrupt.rs
sethp: Ah, much nicer. Thank you!

burrbull (approved):

PR #720, opened 2023-05-29
default Field generics

burrbull:


Emilgardis (commented):
nice, good ergonomic change!

bors r+

PR #722, opened 2023-05-29
FieldSpec

burrbull:


burrbull:
@Emilgardis What do you think about this?

src/generate/generic.rs
```diff
@@ -47,6 +50,12 @@ pub trait RegisterSpec {
     type Ux: RawReg;
 }
 
+/// Raw field type
+pub trait FieldSpec: Sized {
+    /// Raw field type (`u8`, `u16`, `u32`, ...).
+    type Ux: Copy + PartialEq + From<Self>;
+}
+
```
Emilgardis, line 58: why not make this a supertrait?

```suggestion
/// Raw field type
pub trait FieldSpec: Copy + PartialEq + From<Self> + Sized {}

```

burrbull: `Self` and `Ux` are different types if field has enumeratedValues.

burrbull:
@Emilgardis So what your judgment?

Emilgardis (dismissed):
This is good, no concerns from me except maybe the name choice of `Ux`, but we're already using that so it's fine.

feel free to r=me

burrbull:
bors r=@Emilgardis

burrbull:
bors retry

burrbull:
bors r=@Emilgardis

Emilgardis:
bors retry

Emilgardis (approved):

PR #723, opened 2023-06-01
Add steal() for each peripheral.

JalonWong:
This is more convenient than the `Peripherals::steal()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - removed register writer & reader wrappers, generic `REG` in field writers (#731)
 - Updated syn to version 2 (#732)
 - Let readable field fetch doc from svd description (#734)
+- Add `steal()` for each peripheral
 
 ## [v0.29.0] - 2023-06-05
 
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -53,6 +53,25 @@ pub fn render(p_original: &Peripheral, index: &Index, config: &Config) -> Result
         feature_attribute.extend(quote! { #[cfg(feature = #feature_name)] });
     };
 
+    let steal_fn = quote! {
+        /// Steal an instance of this peripheral
+        ///
+        /// # Safety
+        ///
+        /// Ensure that the new instance of the peripheral cannot be used in a way
+        /// that may race with any existing instances, for example by only
+        /// accessing read-only or write-only registers, or by consuming the
+        /// original peripheral and using critical sections to coordinate
+        /// access between multiple new instances.
+        ///
+        /// Additionally, other software such as HALs may rely on only one
+        /// peripheral instance existing to ensure memory safety; ensure
+        /// no stolen instances are passed to such software.
+        pub unsafe fn steal() -> Self {
+            Self { _marker: PhantomData }
+        }
+    };
+
     match &p {
         Peripheral::Array(p, dim) => {
             let names: Vec<Cow<str>> = names(p, dim).map(|n| n.into()).collect();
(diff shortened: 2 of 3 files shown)


thejpster:
Is it sufficient to jus tcall the existing `FOO::ptr()` method to get a `Foo::RegisterBlock` instead? 

I'm not sure I'd want to make it easier to *steal* peripherals, because a stolen peripheral bypasses the ownership system (`p.FOO` is supposed to be a singleton and indicate exclusive access to the *Foo* peripheral).

JalonWong:
@thejpster 
The `FOO::ptr()` is not sufficient. For instance, I have five UARTs, but only two different `Foo::RegisterBlock` types. This makes it challenging for me to treat each UART differently. Moreover, using UART types also enhances code readability.

Indeed, a stolen peripheral bypasses the ownership system. So I implemented the `steal()` as an unsafe method. And there are indeed practical needs for it, such as when I want to split a UART into three instances: a sender, a receiver and a interrupt handler. I found it difficult to achieve this without the `steal()` method.

I believe that anyone who uses unsafe methods should take responsibility for their choices. However, it can be challenging without these unsafe methods when working on low-level MCU development.

adamgreig:
We discussed this PR a bit more in today's meeting, I think it's fair to say there is interest in merging it but we're not sure yet, and in particular wonder if the new function would need to set the global flag that `Peripherals::steal()` sets. Hopefully not, but we'll try and find out.

JalonWong:
> We discussed this PR a bit more in today's meeting, I think it's fair to say there is interest in merging it but we're not sure yet, and in particular wonder if the new function would need to set the global flag that `Peripherals::steal()` sets. Hopefully not, but we'll try and find out.

No need to set the global flag, because before calling the `steal()` of each peripheral, `Peripherals::take()` or `Peripherals::steal()` must be called first.

reitermarkus:
> No need to set the global flag, because before calling the `steal()` of each peripheral, `Peripherals::take()` or `Peripherals::steal()` must be called first.

Given the peripherals are already stolen/taken, maybe this should be named something like `clone_unchecked`.

JalonWong:
> But you don't have a `RegisterBlock`, you have a `*mut RegisterBlock` - the value of the pointer indicates which of the multiple UART peripherals you are currently dealing with. I seem to recall that only thing the singleton objects really do is hide this from you by automatically conjuring up a pointer whenever it is required, and implementing `trait Deref` so accesses to it are transparent.

Implementing traits and writing generics for `RegisterBlock` always feels a bit weird to me.
So currently, I'm doing it like this:
```rust
pub trait UartReg {
    fn ptr() -> *mut Self;
    // or
    unsafe fn mut_ref() -> &'static mut Self;
    // or
    unsafe fn steal_mut(&self) -> &'static mut Self;
}

impl UartReg for USART1 { 
    fn ptr() -> *mut Self {
        (1_usize as *mut Self)
    }
    //...
}
impl UartReg for UART4 { //... }
```
They are very useful. However, they are also weird because the pointer points to an invalid address.

adamgreig:
Thanks for the PR! We discussed it again in today's meeting and we think it might be best if these new methods were associated functions (i.e., like `fn steal() -> Self` without requiring `&self`).

They'd still be unsafe and do the same thing, but you don't need to already have the peripheral before you can make a new copy. With an associated `steal()` there's also no need for a `clone_unchecked()`, since you can just `steal()` directly. We would separately make it so that `Peripherals::steal()` does not set the global flag, so that doesn't need to be a problem for this PR.

What do you think?

JalonWong:
@adamgreig 
> Thanks for the PR! We discussed it again in today's meeting and we think it might be best if these new methods were associated functions (i.e., like `fn steal() -> Self` without requiring `&self`).

It is sufficient for me.
But this approach makes it easier to steal peripheral instances than my current approach. If you think it's appropriate, I have no objections.

adamgreig:
I think this is ready to merge, @rust-embedded/tools any last thoughts?

thejpster:
Won't clippy complain if an `unsafe fn` doesn't have a `# Safety` in the docs?

burrbull:
@adamgreig Maybe you fix docs yourself?

burrbull:
cc @thejpster

thejpster:
Sorry I'm on vacation at the moment.

burrbull (approved):

Issue #724, opened 2023-06-04
update CI


burrbull:
* [ ] replace `bors` with `merge queue`, rust-embedded/wg#671
* [ ] replace `actions-rs` with alternative

cc @rust-embedded/tools

Emilgardis:
I can do this, should be relativly simple, should do the same on rust-embedded/svd also

the wg has not unified on a guideline for how to do this transition, but riscv has transitioned, we can also do so

PR #726, opened 2023-06-04
update ci

Emilgardis:
part of #724

remove all uses of actions-rs which use deprecated gha functions and is not maintained
removes specific clippy-check in favor of running clippy directly


.github/workflows/ci.yml
```diff
@@ -10,12 +10,13 @@ jobs:
     name: CI
     runs-on: ubuntu-latest
     needs: [check, ci-linux, ci-clippy, ci-serde]
+    if: always()
     steps:
       - name: Done
-        run: exit 0
+        run: jq --exit-status 'all(.result == "success")' <<< '${{ toJson(needs) }}'
```
Emilgardis, line 16: preparation for merge queue

Emilgardis (commented):
one thing we lose with this is a [gha problem matcher](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md).

We can add one easily, but I feel like dtolnay/rust-toolchain should do that, but I know the author wants to keep it minimal

burrbull (approved):

burrbull:
Do we need to add `merge_group`?

burrbull:
I messed up svd2rust and svd PRs.

Emilgardis:
this does not implement merge queue, did you enable it here @burrbull ? Going to do that separately

Emilgardis:
#727 implements merge_group

PR #728, opened 2023-06-04
improve regress

### LINKED ISSUE #355: svd2rust-regress usability woes
The interface of `svd2rust-regress` is a bit iffy. On one hand internally every manufacturer and chip name is converted to a "nice" lower-case name with no special characters and this is used everywhere, also for the output of the `--list` option. On the other hand all options like `--manufacturer` and `--chip` expect to receive the unchanged original name.

Also wildcards are not supported which would be really nice to have.

Next problem is that bad tests are only listed or can be executed when the `--bad-tests` option is supplied however I'd expect to see bad tests in the listing as well (maybe with some additional marker) to know that they exists and also to execute them without having to explicitly specify that it's in the "bad tests" category when I specifically ask for a certain chip.

Emilgardis:
Beginning of revamp for our CI

This pr adds the following new features

* [x] Makes it possible to specifically single out a chip and have tests/diff always be run for it
* [x] Allow passing any command to svd2rust via `svd2rust-regress tests -- --command`
* [x] Add alias `cargo regress` -> `cargo run -p svd2rust-regress --` 
* [x] Allow diffing two versions of the output
  This allows for diffing between svd2rust versions and command inputs
  ```sh
  # diff the output between master branch without any extra args and this pr with --atomics
  svd2rust-regress diff --baseline "" --current "--atomics" --chip my-chip
  # diff the output of master branch and this pr 
  svd2rust-regress diff --baseline "@master" --current "@pr" --chip my-chip
  # shorthand for above with a well-known default chip
  svd2rust-regress diff pr
  ```
  where `@pr` = version from current pr, `@master` = version from master branch
* [x] Add a `on_comment` hook to allow diffing prs
  ```text
  /ci diff pr
  ```
  ```text
  /ci diff -c achipwewanttodiff -- --atomics
  ```
  this should output the result in a `<details>` or gist

resolves #355


burrbull:
Nice!

burrbull:
we need to add retry for curl
https://stackoverflow.com/questions/42873285/curl-retry-mechanism

src/generate/peripheral.rs
```diff
@@ -1393,21 +1388,15 @@ fn new_syn_field(ident: Ident, ty: syn::Type) -> syn::Field {
...
 
 fn new_syn_array(ty: syn::Type, len: u32) -> syn::Type {
     let span = Span::call_site();
-    syn::Type::Array(syn::TypeArray {
-        bracket_token: syn::token::Bracket { span },
-        elem: ty.into(),
-        semi_token: Token![;](span),
-        len: new_syn_u32(len, span),
-    })
+    syn::parse_quote_spanned!( span => [#ty; #len] )
```
burrbull, line 1401: Move `syn` update in independent PR, please.
```suggestion
    let len = unsuffixed(len as _);
    syn::parse_quote_spanned!( span => [#ty; #len] )
```

Emilgardis: done in #732

burrbull:
What status of this?

Emilgardis:
https://github.com/Emilgardis/svd2rust/actions/runs/7079947413/job/19267177832?pr=1#step:7:45 feedback appreciated!

burrbull:
What advances of using json here?
Yaml looks more clear:
изображение

burrbull:
> Any opinions on the diff? I think I want it in another way, and having it be saved elsewhere other than action logs would be nice, but I don't think a comment works for it and there's no way to create a gist using an organization account (we'd need a `rust-embedded` bot account)

Need someone more powerful. cc @therealprof @adamgreig

burrbull:
@Emilgardis I've made several fixes. See #782

.cargo/config.toml
```diff
@@ -1,2 +1,5 @@
 [target.aarch64-unknown-linux-gnu]
 linker = "aarch64-linux-gnu-gcc"
+
+[alias]
+regress = "run -p svd2rust-regress --"
```
Emilgardis, line 5: newline needed

.github/workflows/diff.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - run: cargo regress diff ${{ matrix.command }} --use-pager-directly
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GITHUB_PR: ${{ matrix.pr }}
+          GIT_PAGER: delta --hunk-header-style omit
+  summary:
+    runs-on: ubuntu-latest
+    needs: [diff]
+    if: always()
```
Emilgardis, line 63: this needs to check that any generate.outputs.diffs was not empty, otherwise a comment will always be done: https://github.com/Emilgardis/svd2rust/pull/1#issuecomment-1840483908

ci/svd2rust-regress/src/svd_test.rs
```diff
@@ -216,11 +216,11 @@ impl TestCase {
         let svd_url = &self.svd_url();
         let svd = reqwest::blocking::get(svd_url)
             .with_context(|| format!("Failed to get svd URL: {svd_url}"))?
+            .error_for_status()
+            .with_context(|| anyhow!("Response is not ok for svd url"))?
```
burrbull, line 220: Print url here, please, or move upper.

Emilgardis: url is printed with the status error

```
Caused by:
    0: Response is not ok for svd url
    1: HTTP status client error (404 Not Found) for url (https://raw.githubusercontent.com/cmsis-svd/cmsis-svd-data/main/data/Atmel/AT91SAM9G09.svd)
    at ci/svd2rust-regress/src/main.rs:199
```

burrbull: Oh. Sorry then.

.github/workflows/ci.yml
```diff
@@ -171,3 +176,40 @@ jobs:
         uses: Swatinem/rust-cache@v2
 
       - run: cargo fmt --all -- --check
+
+  artifact:
```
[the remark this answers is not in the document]
Emilgardis: artifact should only happen in pull requests thanks to https://github.com/rust-embedded/svd2rust/blob/008983847030e96c455ac9f2f17d5bfe1fa376ae/.github/workflows/ci.yml#L182

so an artifact is only uploaded on pull_request, a release is made on commits to master

burrbull (approved):

ci/svd2rust-regress/src/diff.rs
```diff
@@ -0,0 +1,311 @@
...
+                            .matches(&t.chip.to_ascii_lowercase())
+                    })
+                }
+            })
+            .collect::<Vec<_>>();
+        let test = match (tests.len(), self.sub) {
+            (1, _) => tests[0],
+            (1.., Some(DiffingMode::Pr)) => tests
+                .iter()
+                .find(|t| t.chip == "STM32F401")
```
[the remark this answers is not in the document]
Emilgardis: I basically want to make it easy to generate a diff with

/ci diff pr

that would show the diff of the pr

so instead of always having to do

/ci diff -c AGoodChip

burrbull (dismissed):

ci/svd2rust-regress/src/diff.rs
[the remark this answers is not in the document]
Emilgardis: ok, should we use https://stm32-rs.github.io/stm32-rs/stm32f103.svd.patched then maybe?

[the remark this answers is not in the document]
Emilgardis: fixed!

Emilgardis:
managed to sneak in a bug during the rebase and implementing `regress pr`, should be all good now.

burrbull (approved):

Emilgardis:
🥳

Emilgardis:
/ci diff pr

lets try it :)

Emilgardis:
well it works, but it doesn't like the command , will fix

PR #731, opened 2023-06-05
generic REG in field writers

### LINKED ISSUE #705: Field writer has no impl when enumeratedValue is derived across peripherals
Hello, 
I'm currently implementing a PAC/HAL for the GD32E103 device and set the SVD to derive TIMER*.CTL0.CEN registers to TIMER0.CTL0.CEN register:
```
            <field>
              <name>CEN</name>
              <description>Counter enable</description>
              <bitOffset>0</bitOffset>
              <bitWidth>1</bitWidth>
              <enumeratedValues derivedFrom="TIMER0.CTL0.CEN.CEN"/>
            </field>
```

The generated code reuses the CEN_A and CEN_R types from TIMER0 to every timer but not the CEN_W. The issue is that is defines the type (`pub type CEN_W<'a, const O: u8> = crate::BitWriter<'a, u32, CTL0_SPEC, CEN_A, O>;`) but it does not provides its implementation (to set the variants) as it does on TIMER0.

This issue appeared between svd2rust 0.24.0 and 0.25.0.

burrbull:
should fix #705

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,7 +75,7 @@ jobs:
           - { rust: stable, vendor: Spansion, options: "--atomics" }
           - { rust: stable, vendor: STMicro, options: "" }
           - { rust: stable, vendor: STMicro, options: "--atomics" }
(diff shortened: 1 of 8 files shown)


burrbull:
cc @kaizensparc

therealprof (approved):
LGTM

bors r+

PR #734, opened 2023-06-09
let readable field fetch doc from svd description

### LINKED ISSUE #733: if a field have doc on `read` function, then it should be put it in rust doc
https://github.com/rust-embedded/svd2rust/blob/ef56dd1daeaf0048184567d306ca3d5c9d7c0208/src/generate/register.rs#LL808C89-L808C89

Currently, only doc in `write` function or the `write` side of `read-write` functions has been put into rust `#[doc = ...]`

eZioPan:
will close #733


burrbull:
Update changelog, please.

burrbull (approved):

PR #735, opened 2023-06-13
No longer set DEVICE_PERIPHERALS in steal()

adamgreig:
We've discussed this a bit in the weekly meetings, I propose removing setting `DEVICE_PERIPHERALS` from `steal()` and leave it only in `take()`. This shouldn't meaningfully change the safety conditions or the use of either API, but does make stealing zero-cost. It was always possible to unsafely `transmute()` to obtain `Peripherals` without setting `DEVICE_PERIPHERALS`, so I'm interested if anyone can remember why this existed in the first place or if it is indeed still needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - removed register writer & reader wrappers, generic `REG` in field writers (#731)
 - Updated syn to version 2 (#732)
 - Let readable field fetch doc from svd description (#734)
+- No longer set `DEVICE_PERIPHERALS` in `Peripherals::steal()`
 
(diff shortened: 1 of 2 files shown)


adamgreig:
Closing this. We discovered why this is present: frameworks like RTIC want to `take()` the peripherals but know it cannot fail as they are running before any other code, so can use `steal()` to avoid the overhead associated with `unwrap()` while still setting the `DEVICE_PERIPHERALS` flag to ensure subsequent user code cannot safely end up with duplicate peripherals.

We should add this to the documentation ideally, but it doesn't seem worth changing this behaviour. Instead we can add the new per-peripheral `steal()` in #723 which needn't set `DEVICE_PERIPHERALS`, and that can be the zero-overhead way of unsafely creating a peripheral for new code.

Issue #708, opened 2023-01-27
Implement the ability to work with `W` outside of `write'


AndreySmirnov81:
I want to instead
```rust
use stm32f1::stm32f103 as pac;
let periph = pac::Peripherals::take().unwrap();
let usart1 = periph.USART1;
// ...
usart1.cr1.write(|w| {
    w.ue().enabled();
    w.m().m9();
    w.pce().enabled();
    w.ps().even();
    w.te().enabled();
    w.tcie().enabled();
    w.re().disabled(); // <- First write
    w
});
// ...
usart1.cr1.write(|w| {
    w.ue().enabled();
    w.m().m9();
    w.pce().enabled();
    w.ps().even();
    w.te().enabled();
    w.tcie().enabled();
    w.re().enabled(); // <- Modified
    w
});
// ...
usart1.cr1.write(|w| {
    w.ue().enabled();
    w.m().m9();
    w.pce().enabled();
    w.ps().even();
    w.te().enabled();
    w.tcie().enabled();
    w.re().disabled(); // <- Modified
    w
});
``` 
to do 
```rust
use stm32f1::stm32f103 as pac;
let periph = pac::Peripherals::take().unwrap();
let usart1 = periph.USART1;
// ...
const USART_BASE_CFG: pac::usart1::cr1::W = pac::usart1::cr1::RESET_VALUE
    .ue().enabled()
    .m().m9()
    .pce().enabled()
    .ps().even()
    .te().enabled()
    .tcie().enabled()
    .re().disabled();
// ...
usart1.cr1.write(USART_BASE_CFG);
// ...
usart1.cr1.write(USART_BASE_CFG.re().enabled());
// ...
usart1.cr1.write(USART_BASE_CFG.re().disabled());
```

adamgreig:
Thanks for suggesting this!

We discussed it at the end of today's meeting ([here](https://libera.irclog.whitequark.org/rust-embedded/2023-01-31#1675194291-1675195486;)) though didn't really come to any conclusions yet.

I think one problem is that even if we added a const `W` like your `INITIAL_VALUE` (which is fairly easy to do, though may affect compile times a bit), it's not clear how you could pass it in. The `write()` method takes a closure that returns a reference to the `W`, so for example something like this doesn't work:

```rust
let mut w: stm32f405::tim2::cr1::W  = unsafe { core::mem::transmute(0) };
w.opm().enabled();
tim2.cr1.write(|_| w.cen().enabled());
```

```
[...]

PR #738, opened 2023-07-06
make register reader/writer copiable, return writer

burrbull:
cc @rust-embedded/tools 
Do anyone see disadvantage of using value instead of mutable reference for writer?

Related to #708 

Big breaking change.

This does not let you create writer outside of `write` as suggested by #708, but lets to reuse of writer.

for example
this
```rust
apb.rstr().modify(|_, w| w.$perrst().set_bit());
apb.rstr().modify(|_, w| w.$perrst().clear_bit());
```
can be optimized to
```rust
let w = apb.rstr().modify(|_, w| w.$perrst().set_bit());
apb.rstr().write(|_| w.$perrst().clear_bit());
```

`R<REG>` -> `W<REG>` conversion could be added also. Unsafe?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+- make register reader/writer copiable, return writer from write & modify
 - removed register writer & reader wrappers, generic `REG` in field writers (#731)
 - Updated syn to version 2 (#732)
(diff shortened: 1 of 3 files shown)


thejpster:
That seems like you could get yourself into trouble (accidentally using a w from the function scope not the closure argument, or saving a w and not using for a call, then using it again after and losing your changes).

Is there any evidence that what we have is inefficient?

burrbull:
Closing in favor of #873.

PR #740, opened 2023-08-02
fix atomics feature

burrbull:
Fixes #729


burrbull:
cc @rust-embedded/tools

adamgreig (approved):

Issue #399, opened 2019-11-16
Check in Cargo.lock


hugwijst:
With the stabilization of Cargo.lock file publishing in rust-lang/cargo#7026, lets restart the discussion from #148 about adding `Cargo.lock` to the repo.

My two cents: not having `Cargo.lock` checked in is a huge road block to using `svd2rust` in production.

Backgroud: we are running `svd2rust` on every CI build. By not locking `svd2rust`, there is a chance that the final artifact is not reproducible if a package dependent on by `svd2rust` fails to adhere to semver. More realistically, such an update breaks the build, which is completely unacceptable.

hugwijst:
The first stable version of Cargo that supports publishing lock-files is [1.37](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-137-2019-08-15).

therealprof:
I disagree with your reasoning: Semver has proven to provide pretty good stability that API breakage does pretty much never happens.

On the other hand people are continuously improving their crates in a compatible manner, ranging from dependency cutting over optimisation all the way to fixing bugs detected by newer Rust compiler versions or forward compatibility lints that it actually is **very desirable** to automatically receive those updates without having to intervene manually, which is a cumbersome and very annoying maintenance task.

Indeed as you mentioned the purpose of Cargo.lock is to provide reproducible builds however I don't see much value in rebuilding `svd2rust` continuously. If you want to have a defined and stable version of `svd2rust`, why not compile a specific version, execute your tests and then re-use that pre-built artefact as long as you like. Alternatively you could also use the new cargo ability to download all dependencies and do an offline build, then you'll also get the exact same dependencies every time.

IMHO the far bigger hazard here is that `svd2rust` changes in an incompatible way...

hugwijst:
I agree that semver generally works quite well, and that picking up the latest version of compatible crates is generally desirable *for end users*. Note that this won't change when including `Cargo.lock` files in published releases: Cargo will normally ignore the lock file, unless `install --locked` is used.

Regardless of if we would rebuild `svd2rust` on every build or if we would distribute an artifact, we would want to be able to 1. rebuild the artifact to prove it wasn't tempered with and 2. be able to apply fixes without having to trace back versions of dependencies that allowed the project to build.

[...]

PR #741, opened 2023-08-09
more bagdes, bump deps

burrbull:


burrbull:
cc @rust-embedded/tools

thejpster:
In the absence of a Cargo.lock file, the package bumps won't make any difference because the resolver will pick the highest one anyway. Technically we should specify the lowest we can get away with, but usually it's whatever was the latest at the point the dep was added.

So I would either add the `Cargo.lock` file (and I'll want a second opinion from @rust-embedded/tools on that), or I wouldn't bump the dependencies and I wouldn't show the dependency check badge.

I lean towards adding the lock file myself.

adamgreig:
Bumping the `Cargo.toml` will still ensure everyone has _at least_ that version, but yea since this is typically installed as a binary with `cargo install` you'd expect everyone to get the latest version anyway.

Unless someone runs `cargo install --locked svd2rust` I don't think Cargo will pay attention to the lockfile if we did add one, so I'm not sure there's much point adding it. It might make some difference to people working on it from a Git checkout, but I think that's just developers. I'd expect mostly it's best to use the latest available version of dependencies anyway, and the CI will be checking that too.

Still it shouldn't _hurt_ to bump the deps here to whatever the latest version we want to use is. I wouldn't bother with the dependency check badge myself but I don't mind having it. I think @burrbull wants to do a release ASAP, so I think let's merge this now and we can revisit a lockfile after the release.

adamgreig:
Just adding on to the last comment, Cargo does recommend binaries like svd2rust publish a lockfile so someone could use `--locked` if they wanted to, so perhaps we should be including it here too, even though it won't be used by default.

adamgreig (approved):

thejpster:
Yeah I think we should. Further I think we spiff encourage the use of --locked in the docs/README and also ship binaries to use with cargo binstall which were built with locked dependencies.

Emilgardis:
I'm happily surprised about the inclusion of a Cargo.lock, we've had discussions about this before and ended up not doing it.

cc https://github.com/rust-embedded/svd2rust/issues/399

PR #751, opened 2023-10-18
writer offset as a struct field

burrbull:
This is retry of one of my old experiments.

It does 2 things:
1. Moves `offset` from field writer generic to struct field.
2. Always index field array from 0 (slightly confusing but much easier  to work).

Generated code is much simpler.

The main advantage is you can now simply use `.write(|w| w.fname(3).set_bit()` for write/modify field arrays (previously only reading was accepted) same as `.write(|w| w.fname3().set_bit()`.

The main disadvantage is bigger firmware size in debug mode.

cc @kossnikita, @therealprof  and anyone who want to help.

I tried several examples from `stm32f4xx-hal`:
```
                   before PR vs after PR
blinky-timer-irq  41984/1172 vs 42628/1172
pwm-sinus         49672/8628 vs 50608/8632
rtc-alarm         63704/4680 vs 65648/4680
rtic-usart-shell 88024/17548 vs 88904/17548
```

So it is 2-3% bigger on debug mode for me.

`cargo build --release -j1 --timings`. time of `stm32f4`(f411) compilation on my machine, dependencies not included and peak memory usage during `rtic-usart-shell` compilation:
```
0.14.0    4.73s   758860k
0.15.1    3.00s   530916k
before PR 2.67s   476400k
after PR  2.47s   450996k
```

### DIFF
--- a/src/generate/generic.rs
+++ b/src/generate/generic.rs
@@ -548,7 +548,7 @@ macro_rules! bit_proxy {
             /// Field width
             #[inline(always)]
             pub const fn width(&self) -> u8 {
-                1
+                Self::WIDTH
             }
         
             /// Field offset
--- a/src/generate/generic.rs
+++ b/src/generate/generic.rs
@@ -304,16 +304,17 @@ pub mod raw {
         }
     }
 
-    pub struct FieldWriter<'a, REG, const WI: u8, const O: u8, FI = u8, Safety = Unsafe>
+    pub struct FieldWriter<'a, REG, const WI: u8, FI = u8, Safety = Unsafe>
     where
         REG: Writable + RegisterSpec,
         FI: FieldSpec,
     {
         pub(crate) w: &'a mut W<REG>,
+        pub(crate) o: u8,
         _field: marker::PhantomData<(FI, Safety)>,
     }
 
-    impl<'a, REG, const WI: u8, const O: u8, FI, Safety> FieldWriter<'a, REG, WI, O, FI, Safety>
+    impl<'a, REG, const WI: u8, FI, Safety> FieldWriter<'a, REG, WI, FI, Safety>
     where
         REG: Writable + RegisterSpec,
         FI: FieldSpec,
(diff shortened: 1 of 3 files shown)


kossnikita:
My crate is very similar to yours, and the PAC is also similar. So I don't think there will be a big difference. I don't understand how I can measure the compile time. Just the example build or HAL and PAС?

```
                   before PR vs after PR
qei               29656/9504 vs 30268/9504
pwm            35004/9524 vs 35612/9524
```
Sorry, I haven't finished the more complex examples yet.

burrbull:
> I don't understand how I can measure the compile time

You just need to `cargo clean` and then
```
cargo build --timings
```
(+ `--release`). After finish cargo generates html report and prints a link to it. Doesn't matter what you are building example or HAL. It show compilation time of each dependency.

kossnikita:
Got it. I mean what time do you need? HAL compile or PAC + HAL?

burrbull:
I've measured just a PAC compilation time.
Sure a full time is also interesting, as this PR may also affect on linking time. But I don't see good way to estimate difference.

kossnikita:
```
cargo build --release -j1 --timings --example qei --features at32f415
```

PAC compile time:

6.7±0.15s vs 6.5±0.05s

src/generate/generic.rs
```diff
@@ -509,21 +535,33 @@ macro_rules! bit_proxy {
...
             REG: Writable + RegisterSpec,
             bool: From<FI>,
         {
             /// Field width
             pub const WIDTH: u8 = 1;
 
+            /// Field width
+            #[inline(always)]
+            pub const fn width(&self) -> u8 {
+                1
```
Emilgardis, line 551: this could use the above const

Emilgardis (commented):
i'm fine with this new interface and zero indexing, but I think we need to signal the break a little better, not sure how though

src/generate/generic.rs
burrbull: is this better?

Emilgardis (approved):
lgtm! I'd like it if someone else on @rust-embedded/tools could also review

therealprof (approved):
I don't quite understand why we want to do this but I'm fine in general.

PR #752, opened 2023-10-20
shorter names for constants

burrbull:
~~Due to less lines in `write` and `modify` this potentially leads to smaller size of debug builds.~~

cc @kossnikita Test this also, please.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


kossnikita:
Tested with #751 and #751 + #752

```
                   before PR vs after PR
qei               30268/9504 vs 29656/9504
pwm               35612/9524 vs 35004/9524
```

burrbull:
cc @rust-embedded/tools

Emilgardis (changes requested):
I'm not in favor of this change, makes it more obfuscated imo

since it's `const` there shouldn't be any difference

PR #757, opened 2023-11-01
reexport features

burrbull:


Emilgardis:
why haven't we already done this? :3

adamgreig:
> why haven't we already done this? :3

Right now, the PACs must re-export the `interrupt` macro, but we hope to make this not required soon by changing how the macro works in cortex-m-rt. Re-exporting core peripherals was never required though, so making this optional helps reduce external dependencies - the hope is to make it so PACs don't depend on cortex-m or cortex-m-rt at all.

burrbull:
@Emilgardis next one, please

Emilgardis (approved):

Emilgardis:
docs might need to be updated also

PR #766, opened 2023-11-20
remove deny lints

### LINKED ISSUE #763: Look over usage of #[deny(...)]
Should we use deny lints in the generated crates? 

See discussion in https://github.com/rust-embedded/svd2rust/issues/760

burrbull:
closes #763


burrbull:
cc @Emilgardis what this error could mean?
```
ci/script.sh: line 66: RUST_TOOLCHAIN: unbound variable
```

thejpster:
I think it means that bash variable didn't have a value. It's on the line where you check if it's a nightly Rust version. But you also set the variable in CI before you call the script, so I'm confused?

Emilgardis:
the error is due to `set -u`, an easy fix is to use `"${RUST_TOOLCHAIN:-}"`

Emilgardis (approved):

PR #767, opened 2023-11-21
support of default enum value

### LINKED ISSUE #172: When `isDefault` is present in an `enumeratedValue`, map all non-specified reads to that variant
`isDefault` is _supposed_ to be an alternate to `value` within an `enumeratedValue`. Realistically, it's probably useful to allow both `value` and `isDefault` to be specified.

Current svd2rust appears to ignore it. It would likely be more correct to adjust the `_from` impls on the read proxies to translate all non-specified values into the enumeration marked with `isDefault`. Right now, it appears those will be mapped to a `_Reserved` enumeration.

burrbull:
Closes #172

This is usefull for (reading) fields like this:
изображение

For such fields changes look like:
изображение
изображение

### DIFF
--- a/src/generate/register.rs
+++ b/src/generate/register.rs
@@ -1,14 +1,14 @@
 use crate::svd::{
-    self, Access, BitRange, DimElement, EnumeratedValues, Field, MaybeArray, ModifiedWriteValues,
-    ReadAction, Register, RegisterProperties, Usage, WriteConstraint,
+    self, Access, BitRange, DimElement, EnumeratedValue, EnumeratedValues, Field, MaybeArray,
+    ModifiedWriteValues, ReadAction, Register, RegisterProperties, Usage, WriteConstraint,
 };
 use core::u64;
 use log::warn;
 use proc_macro2::{Ident, Punct, Spacing, Span, TokenStream};
 use quote::{quote, ToTokens};
-use std::borrow::Cow;
 use std::collections::HashSet;
 use std::fmt::Write;
+use std::{borrow::Cow, collections::BTreeMap};
 use svd_parser::expand::{
     derive_enumerated_values, derive_field, BlockPath, EnumPath, FieldPath, Index, RegisterPath,
 };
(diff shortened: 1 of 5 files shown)


burrbull:
cc @Emilgardis

src/generate/register.rs
```diff
@@ -1400,3 +1474,83 @@ fn lookup_filter(
...
+                #pcv,
+            },
+            quote! {
+                #pc::#pcv => #pcval,
+            },
+        )
+    }) {
+        vars.extend(v);
+        casts.extend(c);
+    }
```
Emilgardis, line 1525: this seems wrong, the repr is wrong now when compared to the actual value on chip

```suggestion
    for (v, c) in variants.iter().chain(std::iter::once(default)).map(|v| {
        let desc = util::escape_special_chars(&util::respace(&format!("{}: {}", v.value, v.doc)));
        let pcv = &v.pc;
        let pcval = &util::unsuffixed(v.value);
        (
            quote! {
                #[doc = #desc]
                #pcv = #pcval,
            },
            quote! {
                #pc::#pcv => #pcval,
            },
        )
    }) {
        vars.extend(v);
        casts.extend(c);
    }
```

src/generate/register.rs
```diff
@@ -1400,3 +1474,83 @@ fn lookup_filter(
...
+        }
+        impl From<#pc> for #fty {
+            #[inline(always)]
+            fn from(variant: #pc) -> Self {
+                match variant {
+                    #casts
+                }
+            }
+        }
+    });
```
Emilgardis, line 1548: this could also be a normal cast when the repr is fixed, or if this can't be fixed we should not use `#[repr(u8)]` and provide a `Into` impl

Emilgardis (changes requested):

src/generate/register.rs
burrbull: I've removed `repr`, but I don't see what is wrong with `From`.

Emilgardis: you're right, I mentioned an Into in error

Emilgardis: is there a reason repr couldn't be used?

burrbull: when I started I thought it is impossible, but after refactoring and when you've asked the question it looks simple.
Now changes much smaller:
изображение

Emilgardis: Awesome!

Emilgardis (approved):

PR #769, opened 2023-11-23
mode Config to config module

burrbull:


burrbull:
cc @Emilgardis

src/config.rs
```diff
@@ -0,0 +1,150 @@
...
+    pub source_type: SourceType,
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub log_level: Option<String>,
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub interrupt_link_section: Option<String>,
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub reexport_core_peripherals: bool,
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub reexport_interrupt: bool,
+}
```
Emilgardis, line 49: all of these `#[cfg_attr(feature = "serde", serde(default))]` (and the one with the function call) could be replaced with a `#[cfg_attr(feature = "serde", serde(default))]` on the container I think

https://serde.rs/container-attrs.html#default

```suggestion
#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
#[derive(Clone, PartialEq, Eq, Debug)]
#[cfg_attr(feature = "serde", serde(default))]
pub struct Config {
    pub target: Target,
    pub atomics: bool,
    pub atomics_feature: Option<String>,
    pub generic_mod: bool,
    pub make_mod: bool,
    pub ignore_groups: bool,
    pub keep_list: bool,
    pub strict: bool,
    pub pascal_enum_values: bool,
    pub feature_group: bool,
    pub feature_peripheral: bool,
    pub max_cluster_size: bool,
    pub impl_debug: bool,
    pub impl_debug_feature: Option<String>,
    pub output_dir: PathBuf,
    pub input: Option<PathBuf>,
    pub source_type: SourceType,
    pub log_level: Option<String>,
    pub interrupt_link_section: Option<String>,
    pub reexport_core_peripherals: bool,
    pub reexport_interrupt: bool,
}
```

Emilgardis: this would also bring another benefit, because I think it expands to

```rs
Config {
    <fields that were deserialized>,
    ..Config::default()
}
```
meaning the deserialization of json `{}` = `Config::default()` always.

Emilgardis (dismissed):

src/config.rs
burrbull: What about `output_dir`?

Emilgardis: it should be fine, since it'd use the current_dir call in `<Config as Default>::default()`

burrbull: I've changed `output_dir` to `Option<PathBuf>`.

Emilgardis (approved):

PR #770, opened 2023-11-23
generate defmt::Format for enums

burrbull:
depends on #769

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -133,7 +133,7 @@ fn run() -> Result<()> {
         )
         .arg(
             Arg::new("impl_defmt")
-                .long("impl_defmt")
+                .long("impl-defmt")
                 .help("Add automatic defmt implementation for enumerated values")
                 .action(ArgAction::Set)
                 .value_name("FEATURE"),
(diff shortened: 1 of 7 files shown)


therealprof (dismissed):
LGTM, thanks!

Emilgardis:
this depends on #769

burrbull:
rebased

src/main.rs
```diff
@@ -127,7 +127,14 @@ fn run() -> Result<()> {
...
             Arg::new("impl_debug_feature")
                 .long("impl_debug_feature")
-                .help("add feature gating for block and register debug implementation")
+                .help("Add feature gating for block and register debug implementation")
+                .action(ArgAction::Set)
+                .value_name("FEATURE"),
+        )
+        .arg(
+            Arg::new("impl_defmt")
+                .long("impl-defmt")
```
Emilgardis, line 136: this is not consistent with the other flags, unless we also bring in #771

burrbull: do you suggest to support both `-` and `_` for all flags, or only for old?

Emilgardis: If we go with #771, I think it makes sense to only alias snake case for old flags.
Thus, if we do include #771 then an alias on `impl-defmt` isn't needed, but maybe it's nice to include for familiarity.

burrbull: added alias in last commit

Emilgardis (approved):

PR #771, opened 2023-11-23
dash in flags 

burrbull:


Emilgardis (commented):
nice, I wonder if we could use `alias` also to make this change backwards compatible.

For example, we could 

```rs
trait ArgExt {
    // FIXME: better name
    pub fn long_with_snake_alias(&mut self, arg: &'static str) -> Self;
}
impl ArgExt for clap::Arg {
    pub fn long_with_snake_alias(&mut self, arg: &'static str) -> Self {
        self.long(arg).alias(arg.to_snake_case())
    }
}
````

burrbull:
> nice, I wonder if we could use `alias` also to make this change backwards compatible.

Yes, clap supports aliases
https://docs.rs/clap/latest/clap/struct.Arg.html#method.alias

Emilgardis (dismissed):
I don't like the repetition and would prefer the `long_with_snake_alias` approach but it's not a hard requirement.

I'd add that the reason we ended up using snake_case for multi word flags instead of kebab-case was just because we ended up doing it that way in #362

I think we've had discussions about this before but I can't find those discussions. Anyway, LGTM!

Emilgardis:
sorry! changelog line 12 should be changed

Emilgardis (approved):

PR #783, opened 2023-12-06
fix regress ci problems

Emilgardis:
follow up to #728


burrbull (approved):

burrbull:
Looks like `diff` downloads SVD twice for `baseline` and `current`. Does it?

PR #788, opened 2023-12-23
generic safe/unsafe W::bits

burrbull:
**breaking change**
mark `Writable` register with `Safety` marker.

makes `W::bits` always `unsafe` and always present

add safe `W::set` (other name?) synonym for `W::bits`.


burrbull:
cc @rust-embedded/tools what do you think?

burrbull:
@Emilgardis let's merge this.

Emilgardis (approved):

PR #789, opened 2023-12-23
move in R::field_iter

burrbull:


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - Add `svd::Device` validation after parsing by `serde`
 - Add `skip-crate-attributes` config flag
 - Better display parsing errors
+- move in R::field_iter
```
Emilgardis, line 13: I think this can be worded better to say where this change is applied

Emilgardis (approved):

thejpster:
A description with the rationale for the change would be useful. This can also go in the git commit and the changelog. I assume this reduces generated code size?

burrbull:
> A description with the rationale for the change would be useful. This can also go in the git commit and the changelog. I assume this reduces generated code size?

Not sure about code size. This just fixes compilation:
```
error[E0373]: closure may outlive the current function, but it borrows `self`, which is owned by the current function
   --> src/gpa/pmd.rs:121:21
    |
121 |         (0..16).map(|n| PMD_R::new(((self.bits >> (n * 2)) & 3) as u8))
    |                     ^^^              --------- `self` is borrowed here
    |                     |
    |                     may outlive borrowed value `self`
    |
```
Not sure why it occurs as `bits` is `Copy` type.

burrbull:
What is going with CI?

burrbull:
@Emilgardis

Emilgardis:
should we file an issue on irx-config about the break?

Emilgardis (approved):

burrbull:
> should we file an issue on irx-config about the break?

https://github.com/abakay/irx-config/pull/20

PR #791, opened 2023-12-25
Add `base-address-shift` configuration flag

gephaistos:
Hello!
We are working with the FPGA-based project and unlike with micro-controllers all peripherals are dynamic.
In order to swap different devices (described within several SVD files) in a single board we need to be able to move their addresses.
That's why in this PR we added a flag to shift all base addresses provided in SVD.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -200,6 +200,17 @@ fn run() -> Result<()> {
                 .help("Reexport interrupt macro from cortex-m-rt like crates"),
         )
         .arg(
+            Arg::new("base_address_shift")
+                .short('b')
+                .long("base-address-shift")
+                .alias("base_address_shift")
+                .action(ArgAction::Set)
+                .help("Add offset to all base addresses on all peripherals in the SVD file.")
+                .long_help("Add offset to all base addresses on all peripherals in the SVD file.
+Useful for soft-cores where the peripheral address range isn't necessarily fixed.
+Ignore this option if you are not building your own FPGA based soft-cores."),
+        )
+        .arg(
             Arg::new("log_level")
                 .long("log")
                 .short('l')
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+- Add `base-address-shift` config flag
+
 ## [v0.31.5] - 2024-01-04
 
 - `move` in `RegisterBlock::reg_iter` implementation (iterator of register/cluster array)
--- a/src/config.rs
+++ b/src/config.rs
@@ -28,6 +28,7 @@ pub struct Config {
     pub reexport_core_peripherals: bool,
     pub reexport_interrupt: bool,
     pub ident_formats: IdentFormats,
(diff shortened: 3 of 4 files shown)


burrbull (approved):

burrbull:
Would not it better to generate some global constant and use it calculating `PTR`?

Emilgardis:
I don't see the reasoning for this, why is the base address in the svd not shifted already?

gephaistos:
The address might depend on other cores.
For example, if we want to add our core to [these](https://github.com/riscv-steel/riscv-steel) cores the base address should have one value and in the [other case](https://github.com/chipsalliance/caliptra-rtl) it would require another value.
It would be more convenient to be able to change base address value dynamically instead of changing this value in SVD file each time.

burrbull:
> It would be more convenient to be able to change base address value dynamically instead of changing this value in SVD file each time.

What do you mean under "dynamically" in rust terms?
Is it `const BASE_ADDRESS` or `static mut BASE_ADDRESS`. In second case it can lead to significant downgrade in code optimization.

gephaistos:
I meant to dynamically shift it from SVD file during generation, not on the Rust side. The PR does exactly this functionality.

thejpster:
It seems like a niche use-case, but the PR is low-overhead and most people can just ignore the configuration option. Perhaps the help text should be something like:

> This offset will be added to all base addresses on all peripherals in the SVD file. Useful for soft-cores where the peripheral address range isn't necessarily fixed. Ignore this option if you are not building your own FPGA based soft-cores.

src/main.rs
```diff
@@ -198,6 +198,16 @@ fn run() -> Result<()> {
...
+        .arg(
+            Arg::new("base_address_shift")
+                .short('b')
+                .long("base-address-shift")
+                .alias("base_address_shift")
+                .action(ArgAction::Set)
+                .help("Add offset to all base addresses on all peripherals in the SVD file.
+                Useful for soft-cores where the peripheral address range isn't necessarily fixed.
+                Ignore this option if you are not building your own FPGA based soft-cores."),
+        )
```
Emilgardis, line 210:
```suggestion
                .help("Add offset to all base addresses on all peripherals in the SVD file.
Useful for soft-cores where the peripheral address range isn't necessarily fixed.
Ignore this option if you are not building your own FPGA based soft-cores."),
        )
```

Emilgardis (changes requested):

src/main.rs
Emilgardis: Even better would be to use `.long_help("...")` for the additional info

Emilgardis:
```suggestion
.help("Add offset to all base addresses on all peripherals in the SVD file.")
                .long_help("Add offset to all base addresses on all peripherals in the SVD file.
Useful for soft-cores where the peripheral address range isn't necessarily fixed.
Ignore this option if you are not building your own FPGA based soft-cores."),
        )
```

burrbull (dismissed):

Emilgardis (dismissed):

burrbull:
need rebase

burrbull (approved):

PR #796, opened 2024-01-04
cargo doc constants generation

burrbull:
Before:
изображение

After:
изображение

cc @Emilgardis


Emilgardis:
Why did the diff jobs get skipped?
image

/ci diff pr

Emilgardis:
ugh, must've been something wrong with the runner

I like this change, however I feel like it's a shame rustdoc shows it this way. If it'd be possible to render it as `0x0u32` I'd like it even more

Emilgardis (approved):

PR #802, opened 2024-01-23
write_raw

burrbull:


src/generate/generic.rs
```diff
@@ -203,6 +203,18 @@ impl<REG: Writable> Reg<REG> {
...
+impl<REG: Writable> Reg<REG> {
+    /// Writes raw value to register.
+    ///
+    /// # Safety
+    ///
+    /// Unsafe as it passes value without checks.
+    #[inline(always)]
+    pub unsafe fn write_raw(&self, bits: REG::Ux) {
+        self.register.set(bits);
+    }
```
Emilgardis, line 215: hmm, I'm not certain about the name here. Maybe `set_bits_unchecked`

Emilgardis: or just `set_bits`

PR #806, opened 2024-02-03
regress improvements

Emilgardis:


Emilgardis:
keeping delta would be nice, it provides much better context in the headers , maybe just disabling colors would be enough

Cargo.toml
```diff
@@ -73,4 +73,10 @@ features = ["full","extra-traits"]
...
 members = ["ci/svd2rust-regress"]
 default-members = ["."]
-exclude = ["output"]
+exclude = [
+    "output",
+    # workaround for https://github.com/rust-lang/cargo/pull/12779, doesn't work for output though
+    # see https://github.com/rust-lang/cargo/issues/6009#issuecomment-1925445245
+    "output/baseline/**",
+    "output/current/**"
+]
```
Emilgardis, line 82: this actually doesnt work, at all, it looked like it did work but then it started failing again. https://github.com/rust-lang/cargo/issues/6009#issuecomment-1925445245

either we need to have this be fixed, or we initialize in a temporary folder then move that folder

.github/workflows/diff.yml
```diff
@@ -51,15 +53,11 @@ jobs:
...
-          tool: git-delta
-
-      - run: cargo regress diff --use-pager-directly ${{ matrix.command }}
+      - run: cargo regress diff ${{ matrix.command }}
+      - run: cargo regress diff ${{ matrix.command }}
         env:
           GH_TOKEN: ${{ github.token }}
           GITHUB_PR: ${{ matrix.pr }}
-          GIT_PAGER: delta --hunk-header-style omit
   summary:
```
Emilgardis, line 61: I think we can use `--raw` for this to keep the context delta provides but get rid of the faulty coloring github actions does

Emilgardis: image

burrbull (approved):

PR #805, opened 2024-02-03
change case defaults

burrbull:
Closes #803 

cc @jonathanpallant, @Emilgardis

PR #812, opened 2024-02-15
add ident-format option, fix identificator bugs

burrbull:


src/util.rs
```diff
@@ -21,6 +24,41 @@ pub const BITS_PER_BYTE: u32 = 8;
 /// that are not valid in Rust ident
 const BLACKLIST_CHARS: &[char] = &['(', ')', '[', ']', '/', ' ', '-'];
 
+fn to_pascal_case(s: &str) -> String {
+    if !s.contains('_') {
+        s.to_pascal_case()
+    } else {
+        let mut string = String::new();
+        let mut parts = s.split('_').peekable();
```
Emilgardis, line 32: Should this maybe first convert to UPPER_CASE, and then convert to this "special" pascal case?

Emilgardis (changes requested):

src/util.rs
Emilgardis: Sorry, missed this in #805

burrbull: For `_1token` case? Maybe.

burrbull: No. Same result.

burrbull: I don't see other issues except this. And this name does not look correct for ident.
It's better to ping someone from ESP to fix SVD.

Emilgardis: ah, I was mistaken about what the code does, this looks fine

Emilgardis (approved):

PR #819, opened 2024-02-26
DimSuffix trait

burrbull:
cc @Emilgardis

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -184,18 +184,12 @@ pub fn escape_special_chars(s: &str) -> String {
     escape_brackets(&html_escaped)
 }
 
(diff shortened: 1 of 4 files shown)


src/util.rs
```diff
@@ -433,6 +427,27 @@ pub fn build_rs() -> TokenStream {
     }
 }
 
+pub trait Suffix {
+    fn replace_suffix(&self, suffix: &str) -> Cow<str>;
+    fn clean_suffix(&self) -> Cow<str> {
+        self.replace_suffix("")
+    }
+}
```
Emilgardis, line 435: would it maybe make more sense to call these `expand_dim`, `remove_dim`?

burrbull: Ok. Done.

Emilgardis (approved):

PR #829, opened 2024-03-19
add IsEnum & rm FieldWriterSafe, add set

burrbull:
instead of #825 

This blocks using `variant` method as "safe" replacement for `bits` on fields without `enumeratedValues`.
Also makes `bits` always `unsafe`. Instead adds `set` method for `Safe` fields. (same as was done in 0.32 for register writer).


burrbull:
cc @Emilgardis

Emilgardis (approved):
Lgtm!

PR #836, opened 2024-05-03
Debug impl cleanups

burrbull:
[DebugStruct::field](https://doc.rust-lang.org/std/fmt/struct.DebugStruct.html#method.field) takes `&dyn Debug`. For primitive types it should give same result as `Display`.


burrbull:
cc @Emilgardis

Emilgardis (approved):

PR #838, opened 2024-05-09
use ConstZero instead of Default

burrbull:
Advantages:
- more `const`s
- `zero` is more semantically correct then `default`
- cleaner code

Disadvantages:
- `rustdoc` still not good with `const defaults`. Prints 
изображение

instead of `const ZERO_TO_MODIFY_FIELDS_BITMAP: u32 = 0` for default value.

cc @Emilgardis

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: master
+    branches: [master]
   pull_request:
   merge_group:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 ## [Unreleased]
 
 - Generic `Periph<RB, A>`
+- use `ConstZero::ZERO` instead of `Default::default()` to force const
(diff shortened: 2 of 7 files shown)


burrbull:
@Emilgardis @adamgreig  what do you think about this change?

Emilgardis:
Not sure i understand why num-traits is needed for this, could you elaborate?

burrbull:
> Not sure i understand why num-traits is needed for this, could you elaborate?

We can introduce our own trait if you are saying about this.

Emilgardis (approved):

burrbull:
Still issues in CI with
```
119158 | #[no_mangle]
       |   ^^^^^^^^^ usage of unsafe attribute
       |
help: wrap the attribute in `unsafe(...)`
       |
119158 | #[unsafe(no_mangle)]
```
cc @Emilgardis @robamu 

Need add "edition 2021" or increase MSRV.

burrbull:
cc @Emilgardis rebased

burrbull:
cc @Emilgardis  let's merge

Emilgardis:
Semver break should be fine, implementations of RawReg shouldnt be done by other crates

PR #841, opened 2024-05-16
raw_access

burrbull:
Related to https://github.com/rust-embedded/wg/issues/791

cc @adamgreig

### DIFF
--- a/src/generate/generic_reg_raw.rs
+++ b/src/generate/generic_reg_raw.rs
@@ -1,27 +1,6 @@
 /// Generic peripheral accessor
 pub struct Periph<PER: PeripheralSpec> {
-    _marker: marker::PhantomData<PER>,
-}
-
-impl<PER: PeripheralSpec> Periph<PER> {
-    /// Steal an instance of this peripheral
-    ///
-    /// # Safety
-    ///
-    /// Ensure that the new instance of the peripheral cannot be used in a way
-    /// that may race with any existing instances, for example by only
-    /// accessing read-only or write-only registers, or by consuming the
-    /// original peripheral and using critical sections to coordinate
-    /// access between multiple new instances.
-    ///
-    /// Additionally, other software such as HALs may rely on only one
-    /// peripheral instance existing to ensure memory safety; ensure
-    /// no stolen instances are passed to such software.
-    pub unsafe fn steal() -> Self {
-        Self {
-            _marker: marker::PhantomData,
-        }
-    }
+    rb: PER::RB,
 }
 
 impl<PER: PeripheralSpec> core::ops::Deref for Periph<PER> {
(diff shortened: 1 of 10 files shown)


adamgreig:
I'm not sure if there's any advantage to skipping volatile_cell but still having the registers be themselves memory-mapped (with &self cast to a pointer). We still have the problem that Rust can dereference &self and perform an invalid read. Probably we have to do something similar to chiptool where the peripheral struct owns a pointer and the register methods offset it - the total memory cost is the same because instead of having reference-to-a-ZST like now, we would have an owned-pointer.

It might be hard to do this and keep vcell too, but maybe we can just remove vcell support at the same time?

burrbull:
> It might be hard to do this and keep vcell too, but maybe we can just remove vcell support at the same time?

It is not hard. `svd2rust` code is flexible enough.

Review yet one time, please.

burrbull:
rebased on #847

romancardenas:
Any news on this? May I help somehow?

burrbull:
@romancardenas I've rebased but I'm not sure there is nothing missed.

src/generate/peripheral/accessor.rs
```diff
@@ -62,6 +65,21 @@ impl ToTokens for AccessType {
...
+                doc,
+                name,
+                ty,
+                offset,
+            })) => {
+                let offset = (*offset != 0).then(|| unsuffixed(*offset)).map(|o| quote!(.add(#o)));
+                quote! {
+                    #[doc = #doc]
+                    #[inline(always)]
+                    pub const fn #name(&self) -> #ty {
```
Emilgardis, line 78: Do we not want &#ty here? Same for the other accessors

romancardenas: I think that, with raw access, peripherals return pointers to registers instead of a reference to one of the fields of their register block. Thus, instead of borrowing access, users are "owners" of these registers.

jonathanpallant:
Could we add some expansions of a sample (small) SVD file? Maybe as test cases? It would be easier to review what the output looks like than to review the codegen code.

burrbull:
@jonathanpallant you can you `ci diff` tool to check how PR changes output.

See
https://github.com/rust-embedded/svd2rust/pull/841#issuecomment-2723692172

burrbull:
@jonathanpallant rebased. See https://github.com/rust-embedded/svd2rust/actions/runs/21062390532/job/60571560472?pr=841#step:7:50

I don't plan to work on this PR myself. But you can use it as start point.

burrbull:
@Emilgardis I suggest to merge this. If someone want to improve this he could do it in new PR as changes are under the option flag.

romancardenas:
It would be useful to have this merged in master, even if marked as experimental. I might spend some time on Easter and open a PR with some suggestions.

burrbull:
cc @adamgreig

src/generate/generic_reg_raw.rs
```diff
@@ -0,0 +1,329 @@
...
+pub struct Reg<REG: RegisterSpec> {
+    ptr: *mut u8,
+    _marker: marker::PhantomData<REG>,
+}
+
+unsafe impl<REG: RegisterSpec> Send for Reg<REG> where REG::Ux: Send {}
+
+impl<REG: RegisterSpec> Reg<REG> {
+    #[inline(always)]
+    pub const fn new(ptr: *mut u8) -> Self {
```
romancardenas, line 25: What about making this function `unsafe`?

```suggestion
    /// # Safety: ptr must point to a valid register address
    pub const unsafe fn new(ptr: *mut u8) -> Self {
```

src/generate/generic_reg_raw.rs
```diff
@@ -0,0 +1,329 @@
...
+    /// let bits = periph.reg.read().bits();
+    /// ```
+    /// or get the content of a particular field of a register:
+    /// ```ignore
+    /// let reader = periph.reg.read();
+    /// let bits = reader.field1().bits();
+    /// let flag = reader.field2().bit_is_set();
+    /// ```
+    #[inline(always)]
+    pub unsafe fn read(&self) -> R<REG> {
```
romancardenas, line 56: Following with my previous suggestion, I think `read` would be `safe`:
```suggestion
    pub fn read(&self) -> R<REG> {
```

src/generate/generic_reg_raw.rs
```diff
@@ -0,0 +1,329 @@
...
+        }
+    }
+}
+
+impl<REG: Resettable + Writable> Reg<REG> {
+    /// Writes the reset value to `Writable` register.
+    ///
+    /// Resets the register to its initial state.
+    #[inline(always)]
+    pub unsafe fn reset(&self) {
```
romancardenas, line 69: `write*` methods are more subtle to issues, so I'd leave them `unsafe` by default

src/generate/peripheral.rs
```diff
@@ -609,6 +633,18 @@ fn register_or_cluster_block(
         Ident::new("RegisterBlock", span)
     };
 
+    if config.raw_access {
+        accessors.extend(quote! {
+            pub(crate) const fn new(ptr: *mut u8) -> Self {
```
romancardenas, line 638:
```suggestion
            pub(crate) const unsafe fn new(ptr: *mut u8) -> Self {
```

PR #845, opened 2024-05-29
little cleanups

burrbull:


src/generate/register.rs
```diff
@@ -158,21 +156,21 @@ fn api_docs(
...
             doc,
-            "You can {} this register and get [`{module}::R`]{}. ",
+            "You can {} this register and get [`{module}::R`]{}.",
             method("read"),
             if inmodule { "(R)" } else { "" },
         )?;
 
         if let Some(action) = read_action {
-            doc.push_str("WARN: ");
+            doc.push_str(" WARN: ");
```
Emilgardis, line 173: We could make this a little better with https://github.com/rust-lang/rust/pull/106561 now :D

burrbull: It's interesting. I'll test. The question is should we up MSRV or not?

Emilgardis: since it's just css it doesn't break when the class doesn't exist. No msrv bump needed

burrbull: Seems like done.
Do you have any other ideas on improving documentation?

Emilgardis (approved):

Issue #786, opened 2023-12-14
RISC-V: adoption of `riscv-pac`


romancardenas:
The @rust-embedded/risc-v team is preparing a few changes for the RISC-V ecosystems.
One of these changes is the [`riscv-pac`](https://github.com/rust-embedded/riscv/tree/master/riscv-pac) crate, which will hold all the interrupt-related traits any RISC-V PAC should implement.

I don't know currently how to integrate this in `svd2rust`, but I would be happy to help :)

romancardenas:
We are trying to make the `riscv` ecosystem more flexible by allowing custom, non-standard interrupt sources in [this PR](https://github.com/rust-embedded/riscv/pull/211). There is still some work to be done, but I would like to elaborate on our new approach and how this affects `svd2rust`.

# The `riscv-pac` crate and interrupt types

The [`riscv-pac`](https://github.com/rust-embedded/riscv/blob/riscv-pac/riscv-pac/src/lib.rs) crate defines two types of interrupt sources:

- [`CoreInterruptNumber`](https://github.com/rust-embedded/riscv/blob/fc4f314cc569745be8f25bbbf50470a5b0182c9d/riscv-pac/src/lib.rs#L68): interrupt sources that has a core interrupt number associated to them (e.g., `MachineExternal` in the RISC-V standard).
- [`ExternalInterruptNumber`](https://github.com/rust-embedded/riscv/blob/fc4f314cc569745be8f25bbbf50470a5b0182c9d/riscv-pac/src/lib.rs#L80): interrupt sources assigned to external stimuli (e.g., GPIOs) that are **not** handled by the core directly. Instead, the core interacts with an external interrupt peripheral (e.g., PLIC) to deal with them. This peripheral can be seen as a way to demultiplexer that triggers a single core interrupt that is shared among all the external interrupt sources.

The [`riscv-pac`](https://github.com/rust-embedded/riscv/blob/riscv-pac/riscv-pac/macros/src/lib.rs) crate provides a set of macros to ease the implementation of these traits. The idea is that `svd2rust` can use these macros to automatize trait implementations.

# We need special code for core interrupts

So, core interrupts need additional code that currently is not supported by `svd2rust`. Namely, we need to:

[...]

PR #856, opened 2024-07-01
Use `riscv` section for RISC-V targets

romancardenas:
This PR uses [my proposed `riscv` SVD section](https://github.com/rust-embedded/svd/pull/265) to generate RISC-V PACs using the [new set of traits in `riscv-pac`](https://github.com/rust-embedded/riscv/pull/211).

In this PR, the `mod interrupt` in RISC-V targets is handled differently, and includes:

- `ExternalInterrupt`s (very similar to current `Interrupt`s)
- `CoreInterrupt`s (this is new, previously it just assumed the peripheral follows the RISC-V standard)
- `Priority` levels (necessary for interfacing standard peripherals with trait bounds)
- `HartId`s (necessary for interfacing standard peripherals with trait bounds).

Also, the PAC does not directly generate all the handlers array and so on. Instead, it relies on macros exposed in `riscv-pac`.

I still need some time to pollish everything, but feedback is more than welcome :)

Solves #786

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+- Compatibility with `riscv` 0.12 and `riscv-rt` 0.13
+- Add `riscv_config` section in `settings.yaml`
+  It uses `riscv-pac` traits and standard `riscv-peripheral` peripherals.
+- Add `settings.yaml` file for target-specific settings.
 - Add warning about indexing register arrays
 - Skip generating `.add(0)` and `1 *` in accessors
 - Bump MSRV of generated code to 1.76
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ clap = { version = "4.0", optional = true }
 irx-config = { version = "=3.3.0", features = [
     "cmd",
     "toml-parser",
+    "yaml",
 ], optional = true }
 env_logger = { version = "0.11", optional = true }
 inflections = "1.1"
--- a/src/config.rs
+++ b/src/config.rs
@@ -36,6 +36,8 @@ pub struct Config {
     pub field_names_for_enums: bool,
     pub base_address_shift: u64,
     pub html_url: Option<url::Url>,
+    /// Path to YAML file with chip-specific settings
+    pub settings: Option<PathBuf>,
 }
 
 #[allow(clippy::upper_case_acronyms)]
@@ -312,3 +314,14 @@ impl DerefMut for IdentFormats {
 pub enum IdentFormatsTheme {
     Legacy,
 }
+
+#[cfg_attr(feature = "serde", derive(serde::Deserialize), serde(default))]
+#[derive(Clone, PartialEq, Eq, Debug, Default)]
+#[non_exhaustive]
+/// Chip-specific settings
+pub struct Settings {
+    /// RISC-V specific settings
+    pub riscv_config: Option<riscv::RiscvConfig>,
+}
+
+pub mod riscv;
--- a/src/generate/device.rs
+++ b/src/generate/device.rs
@@ -7,11 +7,11 @@ use std::fs::File;
 use std::io::Write;
 use std::path::Path;
 
-use crate::config::{Config, Target};
+use crate::config::{Config, Settings, Target};
 use crate::util::{self, ident};
 use anyhow::{Context, Result};
 
-use crate::generate::{interrupt, peripheral};
+use crate::generate::{interrupt, peripheral, riscv};
 
 /// Whole device generation
 pub fn render(d: &Device, config: &Config, device_x: &mut String) -> Result<TokenStream> {
@@ -26,6 +26,14 @@ pub fn render(d: &Device, config: &Config, device_x: &mut String) -> Result<Toke
         } else {
             tmp
         }
+    };
+
+    let settings = match config.settings.as_ref() {
+        Some(settings) => {
+            let file = std::fs::read_to_string(settings).context("could not read settings file")?;
+            serde_yaml::from_str(&file).context("could not parse settings file")?
+        }
+        None => Settings::default(),
     };
 
     if config.target == Target::Msp430 {
@@ -187,13 +195,21 @@ pub fn render(d: &Device, config: &Config, device_x: &mut String) -> Result<Toke
         });
     }
 
-    debug!("Rendering interrupts");
-    out.extend(interrupt::render(
-        config.target,
-        &d.peripherals,
-        device_x,
-        config,
-    )?);
+    match config.target {
+        Target::RISCV => {
+            debug!("Rendering RISC-V specific code");
+            out.extend(riscv::render(&d.peripherals, device_x, &settings)?);
+        }
+        _ => {
+            debug!("Rendering interrupts");
+            out.extend(interrupt::render(
+                config.target,
+                &d.peripherals,
+                device_x,
+                config,
+            )?);
+        }
+    }
 
     let feature_format = config.ident_formats.get("peripheral_feature").unwrap();
     for p in &d.peripherals {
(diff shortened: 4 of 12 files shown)


romancardenas:
[This PR](https://github.com/greenlsi/e310x/pull/1) shows how an updated PAC would look like over e310x chips

romancardenas:
Added the `unstable-riscv` feature to optionally activate the proposed new functionality. I still have to include support for `riscv-peripheral` implementations depending on the name of the peripheral.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+- Implement `riscv-pac` traits via experimental `riscv` element in SVD.
+  You must use the `unstable-riscv` feature to use this.
```
Emilgardis, line 11:
```suggestion
  You must compile `svd2rust` with the `unstable-riscv` feature to use this.
```

Emilgardis (commented):
good start! should probably be marked as a draft, I see a bunch of todos

romancardenas:
Yes! I'm currently working on a bunch of changes to use standard peripherals when possible, so it will take a few more iterations until it is ready to merge

romancardenas:
As shown in [this PR](https://github.com/greenlsi/e310x/pull/1), now it is possible to use `riscv-peripheral` standard implementations (i.e., CLINT and PLIC) instead of the ones generated by `svd2rust`.  [This](https://github.com/greenlsi/e310x/blob/3dcafe269c84c35c93bb2a6a5518bfc225624740/src/lib.rs#L14) shows the new outcome. Note that it only takes one macro call per standard peripheral:

```rust
riscv_peripheral :: clint_codegen ! (base 0x2000000 , mtimecmps [mtimecmp0 = (crate :: interrupt :: Hart :: H0 , "[0](crate::interrupt::Hart::H0)")] , msips [msip0 = (crate :: interrupt :: Hart :: H0 , "[0](crate::interrupt::Hart::H0)")] ,);
riscv_peripheral :: plic_codegen ! (base 0xC000000 , ctxs [ctx0 = (crate :: interrupt :: Hart :: H0 , "[0](crate::interrupt::Hart::H0)")] ,);
```

# To do list

- [ ] Configuration parameter to select whether to use the standard macros or the old-school way.
- [ ] A mechanism to specify the CLINT's clock frequency to benefit from its `embedded-hal` implementation.
- [ ] Configuration parameter to activate the use of `async_delay` flag for CLINT.

Regarding the second point: what do you think about providing the CLINT clock frequency via a configuration parameter?

romancardenas:
I started with the configuration parameters... and decided to explore using configuration parameters only, leaving the SVD file untouched.

For the E310x chip, the configuration file would look like [this](https://github.com/greenlsi/e310x/blob/c314fc9a53d640fad356666b364b3e7dcecfcebe/svd2rust.toml).
What do you think about this alternative? I personally like it more.

burrbull:
> I started with the configuration parameters... and decided to explore using configuration parameters only, leaving the SVD file untouched.
> 
> For the E310x chip, the configuration file would look like [this](https://github.com/greenlsi/e310x/blob/c314fc9a53d640fad356666b364b3e7dcecfcebe/svd2rust.toml).
> What do you think about this alternative? I personally like it more.

Makes sense. But why a TOML? YAML should looks much cleaner for such things.

Also what should we do with svd-rs changes and released crates?

burrbull:
I think this should not be in `svd2rust` config file. It's better to use another `--riscv-config file.yaml` option as you may want to use different configs for different input SVDs but with the same `svd2rust.toml`.

romancardenas:
I addressed your suggestions. I'm not sure if my independent YAML file implementation is the best approach, let me know what you think.

You can see how I use it for E310x chips [here](https://github.com/greenlsi/e310x/blob/5f39a74e5ff22d9e9d4e26aa953b40b29b1c637c/riscv_config.yaml)

romancardenas:
- PAC using this PR: [`e310x`](https://github.com/greenlsi/e310x/pull/1)
- HAL using the result PAC: [`e310x-hal`](https://github.com/greenlsi/e310x-hal)
- BSP using the resulting HAL: [`hifive1`](https://github.com/greenlsi/hifive1)
    - It contains an example using the PLIC and CLINT peripherals. You can emulate the CLINT example with QEMU

burrbull:
@romancardenas still a draft?

romancardenas:
Not really, I don't expect to add any changes. I'm waiting to release the next version of the `riscv` crates in crates.io, though.

src/config.rs
```diff
@@ -312,3 +314,57 @@ impl DerefMut for IdentFormats {
...
+
+#[cfg(feature = "unstable-riscv")]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize), serde(default))]
+#[derive(Clone, PartialEq, Eq, Debug, Default)]
+#[non_exhaustive]
+pub struct RiscvConfig {
+    pub core_interrupts: Option<Vec<RiscvEnumItem>>,
+    pub exceptions: Option<Vec<RiscvEnumItem>>,
+    pub priorities: Option<Vec<RiscvEnumItem>>,
+    pub harts: Option<Vec<RiscvEnumItem>>,
```
burrbull, line 326: Is `None` different from `Some(Vec::empty)` here?

romancardenas: Currently, if you define `core_interrupts` in your YAML file, `svd2rust` will generate a custom `CoreInterrupt` struct for your target, regardless of whether it is empty or not. Thus, an empty vector would probably be meaningless. I can change this behavior following any of your design decisions.

burrbull: I want to say that you can leave just `Vec<T>` and check `vec.is_empty()` instead of `option.is_some`.

Possibly you should also mark a field with `serde(default)`. Not sure.

romancardenas: Will take a look to your suggestion!

burrbull:
> I think this should not be in `svd2rust` config file. It's better to use another `--riscv-config file.yaml` option as you may want to use different configs for different input SVDs but with the same `svd2rust.toml`.

What about this?

I'd want to see something like this:
```rust
/// Common svd2rust options
struct Config {
   ...
   /// Path to YAML file with chip related settings
   pub settings: Option<PathBuf>,
}

#[non_exhaustive]
struct Settings {
    pub riscv_config: Option<RiscvConfig>,
    
    // other settings (for example #576)
}
```

romancardenas:
> I'd want to see something like this:
> 
> ```rust
> /// Common svd2rust options
> struct Config {
>    ...
>    /// Path to YAML file with chip related settings
>    pub settings: Option<PathBuf>,
> }
> 
> #[non_exhaustive]
> struct Settings {
>     pub riscv_config: Option<RiscvConfig>,
>     
>     // other settings (for example #576)
> }
> ```

Oh, I see! I added a `riscv_config` parameter to the general configuration, but maybe it will be more scalable having a separate `settings.yml` file that includes the RISC-V-specific configuration (together with potentially new parameters such as Zaamo-only registers).

romancardenas:
@burrbull Let me know if the new version looks as you expected

src/generate/device.rs
```diff
@@ -30,6 +30,14 @@ pub fn render(d: &Device, config: &Config, device_x: &mut String) -> Result<Toke
         }
     };
 
+    let _settings = match config.settings.as_ref() {
```
burrbull, line 33: Why is it underscored?

Cargo.toml
```diff
@@ -35,13 +35,14 @@ default = ["bin", "json", "yaml"]
 bin = ["dep:clap", "dep:env_logger", "serde", "dep:irx-config"]
 json = ["dep:serde_json"]
 yaml = ["dep:serde_yaml"]
-unstable-riscv = ["irx-config/yaml"]
+unstable-riscv = []
```
burrbull, line 38: Is it unstable yet?

burrbull:
Squash commits, please.

romancardenas:
Done

burrbull (approved):

burrbull:
Please, revert commits from `svd-rs`.

PR #868, opened 2024-10-14
Restrict `proc-macro2` to Version 1.0.86

AdinAck:
I noticed that `proc-macro2` version `1.0.87` introduced character validation for the `Punct` struct, resulting in `svd2rust` panicking like so:

```
thread 'main' panicked at {me}/.cargo/registry/src/index.crates.io-6f17d22bba15001f/proc-macro2-1.0.87/src/lib.rs:847:13:
unsupported proc macro punctuation character '{'
```

Temporary solution is holding back `proc-macro2` at version `1.0.86`. Better solution would probably be representing the `{` character with a different struct than `Punct`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -893,9 +893,9 @@ dependencies = [
 
 [[package]]
 name = "proc-macro2"
-version = "1.0.71"
+version = "1.0.86"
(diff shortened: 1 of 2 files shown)


burrbull:
Fixed in #864

PR #870, opened 2024-10-16
move must_use from methods to generic type

burrbull:
~~need to test first if~~ it works as expected


burrbull:
cc @Emilgardis

Emilgardis (approved):
Makes sense!

PR #873, opened 2024-10-20
return raw value from write/modify

burrbull:
This looks not so beautiful like #738, but almost not breaking and allows to transfer data between writes with unsafe.
```rust
    let bits = rcc.apb1rstr().modify(|_, w| w.usart2rst().clear_bit());
    rcc.apb1rstr().write(|w| unsafe { // save on 1 read operation
        w.bits(bits).usart2rst().set_bit()
    });
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - Skip generating `.add(0)` and `1 *` in accessors
 - Bump MSRV of generated code to 1.76
 - move `must_use` from methods to generic type
-- Add `write_and`, `write_with_zero_and`, and `modify_and` register modifiers
+- *breaking change* Return raw writtened value
+- Add `from_write`, `from_write_with_zero`, and `from_modify` register modifiers
+  with generic return value
 
 ## [v0.33.5] - 2024-10-12
 
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -589,7 +589,7 @@ main() {
             test_patched_stm32 stm32f7x3
             test_patched_stm32 stm32g070
             test_patched_stm32 stm32g473
-            test_patched_stm32 stm32h753
+            test_patched_stm32 stm32h743
             test_patched_stm32 stm32l0x3
             test_patched_stm32 stm32l162
             test_patched_stm32 stm32l4x6
--- a/src/generate/generic_reg_vcell.rs
+++ b/src/generate/generic_reg_vcell.rs
@@ -74,18 +74,18 @@ impl<REG: Resettable + Writable> Reg<REG> {
     /// ```
     /// In the latter case, other fields will be set to their reset value.
     #[inline(always)]
-    pub fn write<F>(&self, f: F)
+    pub fn write<F>(&self, f: F) -> REG::Ux
     where
         F: FnOnce(&mut W<REG>) -> &mut W<REG>,
     {
-        self.register.set(
-            f(&mut W {
-                bits: REG::RESET_VALUE & !REG::ONE_TO_MODIFY_FIELDS_BITMAP
-                    | REG::ZERO_TO_MODIFY_FIELDS_BITMAP,
-                _reg: marker::PhantomData,
-            })
-            .bits,
-        );
+        let value = f(&mut W {
+            bits: REG::RESET_VALUE & !REG::ONE_TO_MODIFY_FIELDS_BITMAP
+                | REG::ZERO_TO_MODIFY_FIELDS_BITMAP,
+            _reg: marker::PhantomData,
+        })
+        .bits;
+        self.register.set(value);
+        value
     }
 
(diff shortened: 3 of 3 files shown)


AdinAck:
This is certainly similar to my idea but not quite the same.

This does not permit arbitrary return types.

I have been using the ability to return arbitrary return types to enforce type-state validity for HALs like so:

```rust
let state = reg.modify(|_, w| {
    State::set(w.some_field())
});
```

[Full example](https://github.com/AdinAck/stm32g4xx-hal/blob/cfb5874c8cf0b766321e523a2b80e12aadf9eaad/src/cordic.rs#L1066-L1074)

burrbull:
Ok. After you've wrote I understand even less.
What is `State::set`? Constructor of `State`?
What it takes? Value of register or value of field?
What does `some_field()` do? Does it change a field somehow?

I cannot to chain field more?

AdinAck:
> What is `State::set`? Constructor of `State`?

Yes, a type-state corresponds to the value of a bitfield.

> What it takes? Value of register or value of field?

`set` takes a mutable reference to the exact bitfield of a register it represents the state of.

> What does some_field() do? Does it change a field somehow?

`some_field` was a placeholder name for some imaginary bitfield in some register.

Imagine a hypothetical peripheral which can only be enabled or disabled:
```rust
struct ImaginaryPeripheral<State: PowerState> {
    rb: IMGPERIPH,
    state: State,
}
```

If I wanted a `ImaginaryPeripheral<Enabled>` I would need to convert `ImaginaryPeripheral<Disabled>` into that. (Just like conversion of pin modes: `PB3<Analog>` into `PB3<Input<PullUp>>`)

So I could write a function for conversion:

```rust
impl<State: PowerState> ImaginaryPeripheral<State> {
    fn freeze<NewState: PowerState>(self) -> ImaginaryPeripheral<NewState> {
        let new_state = self.rb.reg().modify(|_, w| { // peripheral has a single register called "reg"
            NewState::set(w.power()) // power type-state only needs the power bitfield of this register
        });

        ImaginaryPeripheral {
            rb: self.rb,
            state: new_state,
        }
    }
}
```

The peripheral type holds instances of type-states to prove that the type-state was in fact `set`.

The definition of `set` for `Disabled` would say, set the bit to `0`, and the definition of `set` for `Enabled` would set the bit to `1`.

All of this is just an example of *my reason* for wanting to return arbitrary values from the register write/modify closure. It is conceivable other use-cases exist, but this is mine.

AdinAck:
Looking at your example, I think I see what you're going for, and that's definitely a different problem than the one I am trying to solve.

burrbull:
Traditional solution for your example.

Stable Rust:
```rust
struct ImaginaryPeripheral<State: PowerState> {
    rb: IMGPERIPH,
    state: State, // or even PhantomData<State>
}
trait PowerState {
   const STATE: StateEnum; // generated by svd2rust
}
struct Enabled;
impl PowerState for Enabled {
   const STATE: StateEnum = StateEnum::Enabled;
}

self.rb.reg().modify(|_, w| w.power().variant(NewState::STATE));
ImaginaryPeripheral {
    rb: self.rb,
    state: NewState,
}
```

Nightly Rust:
```rust
#![feature(adt_const_params)]

struct ImaginaryPeripheral<const State: StateEnum> {
    rb: IMGPERIPH,
}

self.rb.reg().modify(|_, w| w.power().variant(State));
ImaginaryPeripheral::<NewState> {
    rb: self.rb,
}
```

burrbull:
Anyway I suggest you to discuss this in Tuesday meeting.

AdinAck:
Yeah that is how HALs primarily look right now, and in fact the version of the example I linked earlier that made it into stm32g4xx-hal looks like this.

The problem I have with it is that the line:

```rust
self.rb.reg().modify(|_, w| w.power().variant(NewState::STATE));
```

could be commented out and the program would still compile no problem.

More complex peripherals have unintuitive type-state relations where errors can slip through easily. I'm trying to eliminate these kinds of errors.

I like knowing that any and all undesired behavior is limited to `set` implementations, which are trivial to implement.

AdinAck:
> Anyway I suggest you to discuss this in Tuesday meeting.

Ok I still don't know what the procedure is for this but I will try. Thanks!

burrbull:
> could be commented out and the program would still compile no problem.
> More complex peripherals have unintuitive type-state relations where errors can slip through easily. I'm trying to eliminate these kinds of errors.

Ok. It's clearer now.

Emilgardis:
Isn't this a huge change?

burrbull:
@Emilgardis let's merge this one too.

Emilgardis:
Hmm, I guess it shouldn't break to much, since people should have closed the expression with `;`

burrbull:
> Hmm, I guess it shouldn't break to much, since people should have closed the expression with `;`

Compared to #738 this does not affect on expressions inside `write`, only outside.
For example in `f1xx-hal` there is only 18 places requires adding `;`.

Issue #859, opened 2024-08-26
Register Write Limitation


AdinAck:
*Moved from `stm32-rs-nightlies`*

I ran into an issue when writing to a register where I wanted to return a value from the write, like this:

```rust
let val = reg.write(|w| {
    T::set(w.field())
});
```

But, the signature of `write` forbids this:

```rust
pub fn write<F, T>(&self, f: F)
    where
        F: FnOnce(&mut REG::Writer) -> &mut W<REG>,
```

So I'm curious, why not change the signature to:

```rust
pub fn write<F, T>(&self, f: F) -> T
    where
        F: FnOnce(&mut REG::Writer) -> T
```

By implementing it like this:

```rust
pub fn write<F, T>(&self, f: F) -> T
where
    F: FnOnce(&mut W<REG>) -> T,
{
    let mut writer = W {
        bits: REG::RESET_VALUE & !REG::ONE_TO_MODIFY_FIELDS_BITMAP
            | REG::ZERO_TO_MODIFY_FIELDS_BITMAP,
        _reg: marker::PhantomData,
    };
    let result = f(&mut writer);

    self.register.set(writer.bits);

    result
}
```

Is there a drawback to this?

burrbull:
https://github.com/rust-embedded/svd2rust/pull/738#issuecomment-1624819353

AdinAck:
I don't believe my proposal is related to that. I don't want to return the writer, in fact, the proposed body of `write` forbids `T` ever being `W` or `&mut W` since `bits` is moved out of the writer for `register.set`, no?

AdinAck:
Ok yes I believe I set up an analogous setup in rust playground:

```rust
#[derive(Debug)]
struct Bits(u32);

#[derive(Debug)]
struct W {
    bits: Bits,
    _reg: (),
}

fn consume_bits(Bits(num): Bits) {
    println!("bits: {}", num);
}

fn write<F, T>(f: F) -> T
where
    F: FnOnce(&mut W) -> T,
{
    let mut writer = W { bits: Bits(0), _reg: () };
    
    let result = f(&mut writer);
    
    consume_bits(writer.bits);
    
    result
}

fn main() {
    let outside = write(|_w| 0xaa);
    // let outside = write(|w| w); <- does not compile
    println!("outisde: {:?}", outside);
}
```

So I don't believe this introduces any complications as mentioned in #738.

burrbull:
[...]

PR #874, opened 2024-10-22
Arbitrary Return Types for Register Modification Closures

AdinAck:
Discussed in #859.

## Added
- `write_and`
- `write_with_zero_and`
- `modify_and`

These functions return a `T`. The writer cannot be returned outside of the closure because the lifetime is constrained to remain within the function body.

## Motivation

My reasons for this change are to allow strong type-state validity:

```rust
let state = rb.reg().write_and(|w| State::set(w.field()));
```

Thanks @burrbull!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - Skip generating `.add(0)` and `1 *` in accessors
 - Bump MSRV of generated code to 1.76
 - move `must_use` from methods to generic type
+- Add `write_and`, `write_with_zero_and`, and `modify_and` register modifiers
 
 ## [v0.33.5] - 2024-10-12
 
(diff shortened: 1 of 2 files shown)


burrbull:
> These functions return a `T`

Return a **generic** `T` which corresponds to type returned by passed closure.

Just a note.

burrbull:
Instead of suffix it can be prefix.

What about `external_write`? `ext_write`

AdinAck:
Hey great minds think alike, I was just thinking the same thing.

I was thinking `from_write` (but that's a bit close to `From`) and `with_write`.

burrbull:
`from_write` sounds good

AdinAck:
Ok well I guess we can tentatively go with `from_*` then. Thanks!

AdinAck:
On a side note, to reduce duplicate code, I *think* the original methods could invoke the new ones with the generic T filled in. Will look into it later.

burrbull (approved):

burrbull:
cc @Emilgardis any comments?

Emilgardis:
Would it make sense to pit this behind a setting? I think not, but somethibg to maybe consider

PR #876, opened 2024-10-23
fix and check compilation with no features

### LINKED ISSUE #875: Current `master` branch does not build with `--no-default-features`
I've updated the workflow in my fork to include this option, failing run can be seen here:  
https://github.com/jessebraham/svd2rust/actions/runs/11481198244

Enabling the `yaml` feature also does not resolve this:
https://github.com/jessebraham/svd2rust/actions/runs/11481238404

Emilgardis:
resolves #875

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -37,6 +37,7 @@ pub struct Config {
     pub base_address_shift: u64,
     pub html_url: Option<url::Url>,
     /// Path to YAML file with chip-specific settings
+    #[cfg(feature = "yaml")]
     pub settings: Option<PathBuf>,
 }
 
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -37,6 +37,7 @@ pub struct Config {
     pub base_address_shift: u64,
     pub html_url: Option<url::Url>,
     /// Path to YAML file with chip-specific settings
+    #[cfg(feature = "yaml")]
     pub settings: Option<PathBuf>,
```
jannic, line 41: I think this would just ignore the settings file, even if it is configured, correct?
Wouldn't it be better to report an error instead, like I implemented in #877?

PR #879, opened 2024-10-25
[test] disable respace

burrbull:


burrbull:
@Emilgardis Let's merge this too. This should be done in `svdtools`, not in `svd2rust`.

burrbull:
cc @thejpster

Emilgardis:
why was this made into a draft again?

Emilgardis:
I dont enjoy the output :/ doesn't look good

burrbull:
> I dont enjoy the output :/ doesn't look good

What do you suggest?

Emilgardis:
Hmm, I guess the problem I have is how the source material is formatted. is there a issue/pr for improving the descriptions in the patched svds?

burrbull:
> is there a issue/pr for improving the descriptions in the patched svds?

For adequate support `svdtools` requires regex replace.

But I still not sure about syntax:
1) https://github.com/rust-embedded/svdtools/pull/252: easiest, but support descriptions only
2) https://github.com/rust-embedded/svdtools/compare/master...regex-replace more universal, but more complex code and patch rules
3) allow for each string entry of `svd_rs::XxBuilder` to take "replace string". Something similar to `sed` maybe.

4.) always process this special case: replace `\n\s+` with ` `.

Emilgardis:
Could we reverse the information about where the text in the description starts, then apply that spacing and then do a u"nindent"?

burrbull:
> Could we reverse the information about where the text in the description starts, then apply that spacing and then do a "unindent"?

Could you show example? I don't understand what you mean.

burrbull:
> instead, I think since we can, via [`text_pos_at`](https://docs.rs/roxmltree/0.18.1/roxmltree/struct.Document.html#method.text_pos_at) get how much of that is indentation to make the xml look nice and how much is indentation for formatting, we then use the same logic `unindent` uses to make the doc string

You mean to check that text indentation is equal to tag indentation?

Emilgardis:
Yes, essentially.

I'm fine with merging this however, but I'm a bit suspicious about if we're really handling this in a good way or not.

Emilgardis (approved):

PR #891, opened 2024-12-11
unsafe no-mangle

burrbull:
Fix compilation on nightly, but bumps Rust to 1.82

cc @Emilgardis


burrbull:
Alternative is to just force edition 2021 in CI

Emilgardis:
I think specifying 2021 is the way to go for now

burrbull:
> I think specifying 2021 is the way to go for now

#892

burrbull:
cc @Emilgardis includes #933

Issue #855, opened 2024-06-25
Add (optional) `doc_auto_cfg` attribute to generated library


robamu:
In my crates, I currently use the following code snippet at the top of `lib.rs`:

```rs
#![cfg_attr(docsrs, feature(doc_auto_cfg))]
```

For documentation built by docs.rs, this explicitly displays required features for modules/functions etc. without requiring additional attributes.
What do you think about adding this to the generated library?

In all my libraries using this, I use the following snippets in `Cargo.toml` to use this for the docs-rs build:

```rs
[package.metadata.docs.rs]
rustdoc-args = ["--generate-link-to-definition"]
```

robamu:
Actually, I have issue building this locally without warnings with current nightly.

Adding the following snippet to `Cargo.toml`:

```toml
[lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = ['cfg(docs_rs)'] }
```

fixes the warning ,but the documentation still is not built properly.. In any case https://docs.rs/va108xx/0.3.0/va108xx/ shows the end result , with the documentation displaying where the `rt` feature is required.

robamu:
After some more research:

It is sufficient to use 

```rs
#![cfg_attr(docsrs, feature(doc_auto_cfg))]
```

in the `lib.rs`. The `docsrs` configuration will be set by docs-rs automatically.
In the `Cargo.toml`

```toml
[package.metadata.docs.rs]
rustdoc-args = ["--generate-link-to-definition"]
```

is sufficient now.

The old way I used to build this locally does not work anymore. I tested this locally using

```sh
export RUSTDOCFLAGS="--cfg docsrs"
cargo +nightly doc --all-features  --open
```

the old way

```
cargo +nightly doc --all-features --config 'build.rustdocflags=["--cfg", "docsrs", "--generate-link-to-definition"]' --open
```

does not work anymore for some reason.

Issue #689, opened 2022-10-29
Why doesn't the CI use `svd2rust-regress`?


YuhanLiin:
The CI script seems to do the exact same thing as `svd2rust-regress`, so I'm wondering why they're separated.

Emilgardis:
no real reason but that we haven't gotten to it :)

Some previous discussions on this 

https://github.com/rust-embedded/svd2rust/pull/542#issuecomment-934170012

I've previously also mention the possibility to use something like `bors try --regress` to catch changes in prs and print them as comments, with the new github cli I think we can do this easy and improve the experience.

burrbull:
cc @Emilgardis should CI be replaced with `svd2rust-regress test`?

Emilgardis:
I think it could and probably should

PR #899, opened 2025-01-15
improve docs of generated library

robamu:
This adds feature gate displays, for example like the ones in the [syn docs](https://docs.rs/syn/latest/syn/).
The generated library only has a few of those, but I find them to be very useful. Fixes https://github.com/rust-embedded/svd2rust/issues/855

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - Other internal clippy fixes for `clippy::manual_div_ceil`, `clippy::nonminimal_bool` and
   `clippy::needless_lifetimes`
 - Update `svd-rs` to 0.14.11
+- Added `#![cfg_attr(docsrs, feature(doc_auto_cfg))]` to the generated library code. This
+  adds a display of the feature gates in the documentation of the generated library
 
 ## [v0.35.0] - 2024-11-12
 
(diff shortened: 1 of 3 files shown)


burrbull:
It is nightly feature, is not it?

robamu:
The additional feature gate displays require building docs with nightly, but it is still possible to generate docs without nightly without issues.

burrbull:
We possibly need a "doc workflow" (on stable & nightly). It can be independent job or added to one of existent. Could you work on this?

robamu:
You mean building the docs for the PACs generated as part of the CI? I could look into this.. I would probably have to adapt the `ci/script.sh` for this? Maybe it would be a good idea to move the CI from the shell script to the regress tool as a pre-requisite?

burrbull:
> Maybe it would be a good idea to move the CI from the shell script to the regress tool as a pre-requisite?

Sure. https://github.com/rust-embedded/svd2rust/issues/689
But better in separate PR.

robamu:
https://github.com/rust-embedded/svd2rust/pull/905 is WIP and I already was thinking of how to best add this workflow. I saw that the bash script is used for the clippy check as well. If this is done by the regress tool, I am wondering what the best way would be to do the CI. The regress tool already takes care of creating new projects for each chip. One way would be to also run `clippy` and `doc` for each job directly after the check. The alternative would be to run clippy and docs only for a small selection of chips in s separate job.. what would be the better/preferred way here? @burrbull @Emilgardis

Emilgardis:
For checking if generated doc comments is correct it'd be enough to just run it with the regular per-target tests imo.

burrbull:
> run clippy and docs only for a small selection of chips in s separate job

This is enough

burrbull:
Rebase, please.

burrbull (approved):

PR #905, opened 2025-02-03
Update cargo regress tool and move CI

robamu:
- CI now uses the svd2rust-regress tool instead of a bash script
- Fixed a bug where the workspace TOML file was still updated and changed
  when running the regression tool
- Upgrade `portable-atomic` version for CI checks to 1.
- Introduce MSRV v1.82.0 for the regress tool.
- Allow passing multiple passthrough options to the regress tool.
- Allow passing options using the test YAML file using the `opts` key
- Allow specifying a suffix for test cases inside the YAML file using the `suffix` key.
  This can be used to resolve name duplications.
- Allow skipping the `cargo check` for test cases with a `skip_check` key
  inside the YAML file


robamu:
I don't know if I can enable CI for my own fork, that would be useful for testing..

.github/workflows/ci.yml
```diff
@@ -108,14 +106,10 @@ jobs:
       - name: Self install
         run: |
           cargo install svd2rust --path .
+          cargo install svd2rust-regress --path ci/svd2rust-regress
```
Emilgardis, line 109: this is uneccessary, can just call the alias `cargo regress`

.github/workflows/ci.yml
```diff
@@ -108,14 +106,10 @@ jobs:
...
 
-      - name: Run CI script for `${{ matrix.vendor }}` under rust `${{ matrix.rust }}` with options=`${{ matrix.options }}`
-        env:
-          VENDOR: ${{ matrix.vendor }}
-          OPTIONS: ${{ matrix.options }}
-          COMMAND: check
-          RUST_TOOLCHAIN: ${{ matrix.rust }}
-        run: bash ci/script.sh
+      - name: Run regression tool
+        run: svd2rust-regress tests --test-cases ci/svd2rust-regress/tests.yml -m ${{ matrix.vendor }} ${{ matrix.options }}
```
Emilgardis, line 112:
```suggestion
        run: cargo regress tests --test-cases ci/svd2rust-regress/tests.yml -m ${{ matrix.vendor }} ${{ matrix.options }}
```

ci/svd2rust-regress/src/github.rs
```diff
@@ -93,7 +93,7 @@ fn find_executable(dir: &Path, begins: &str) -> Result<Option<PathBuf>, anyhow::
                     .path()
                     .extension()
                     .is_some_and(|s| s == std::env::consts::EXE_EXTENSION))
-            && entry.path().extension().is_none_or(|s| s != "gz")
+            && entry.path().extension().map_or(true, |s| s != "gz")
```
Emilgardis, line 96: why this change?

src/util.rs
```diff
@@ -72,13 +72,16 @@ impl Case {
         }
     }
     pub fn sanitize<'a>(&self, s: &'a str) -> Cow<'a, str> {
-        let s = sanitize(s);
+        self.cow_to_case(s.into())
+    }
```
Emilgardis, line 76: doesn't this make sanitize not sanitize anymore?

.github/workflows/ci.yml
```diff
@@ -54,46 +54,44 @@ jobs:
     runs-on: ubuntu-latest
     needs: [check]
     strategy:
+      fail-fast: false
       matrix:
-        # Options are all, none, strict and const
         include:
```
Emilgardis, line 59: we could use a

```yaml
outputs:
    matrix: ${{ steps.generate-matrix.outputs.matrix }}
```

job

which would convert the matrix in this file to

```
strategy:
  fail-fast: false
    include: ${{ fromJson(needs.generate-matrix.outputs.matrix) }}
```

however, that wouldn't test different command flags. So maybe we shouldn't do that and instead just have a single job to test all defined test targets in `svd2rust-regress`

ci/svd2rust-regress/src/svd_test.rs
```diff
@@ -386,3 +446,30 @@ fn visit_dirs(dir: &Path, cb: &mut dyn FnMut(&fs::DirEntry)) -> std::io::Result<
...
+        // Calling cargo init will add the chip test directory to the workspace members, regardless
+        // of the exclude list. The unmodified manifest path is stored here and written
+        // back after cargo init.
+        let manifest_path = crate::get_cargo_workspace().join("Cargo.toml");
+        let unmodified_toml = fs::read(&manifest_path).unwrap();
+        Self {
+            unmodified_toml,
+            manifest_path,
+        }
+    }
```
Emilgardis, line 468: im not sure this is needed anymore, due to us adding `[workspace]` to the manifest

Emilgardis (changes requested):

.github/workflows/ci.yml
robamu: I thought it might be fast if it is installed first before running it, but I don´t think it is..

ci/svd2rust-regress/src/github.rs
robamu: this API is not available in the MSRV version of Rust. But I also need to investigate why this wasn´t an issue with the old CI

src/util.rs
robamu: yes :+1:  did not see that, I just did a quickfix here because I received foldernames like espressifesp32

ci/svd2rust-regress/src/github.rs
robamu: I think what's required here is that the regress tool still needs to be built with stable (or at least v1.82), and then only use v1.76 for checking the crate. I think this will be difficult with the matrix options. However, what we might be able to do (and I already did some prep-work for this) is that this MSRV rust version is specified in a toolchain TOML file in the generated project.

robamu:
I think this is getting to a good state now. The only problematic thing is that I had to disable the checks for MSP430. When I try to do a `cargo check` (locally, but same problem in CI)  here, I get errors that some external C function specified in some dependency is not valid for the target.. I don´t know why this wasn´t an issue with the previous CI.

.github/workflows/ci.yml
Emilgardis: I do this in cross-rs, where I have a targets.toml that populate the CI automatically.

The input basically expands to a list of targets, not just a single target in a list

Emilgardis:
no changelog is needed for this

robamu:
Fixes #689

.github/workflows/ci.yml
robamu: Would this be for a different PR?

burrbull (approved):

PR #907, opened 2025-02-07
update CI to run clippy and docs

robamu:


burrbull:
At least one `doc` job must be run on stable.

burrbull:
NRF ci fails with name collision. You should change one of suffixes https://docs.rs/svd2rust/latest/svd2rust/#the---ident-format-and---ident-formats-theme-flags

burrbull:
quotes

burrbull:
squash after error fixing, please.

robamu:
Yes. One CPU fails in clippy due to some formatting issues with doc comments. That disappears when I run `cargo fmt` on the file. Not sure how to best continue here. I think most PACs use the `form` tool or at least run `cargo fmt`, so maybe it would be okay to run the formatter before running clippy and generating docs?

burrbull:
SiliconLabs-SIM3L1x8_SVD fails.

By the way why CI does not show SVD-file URL?

burrbull:
Looks like issue with escaping newlines in some cases:
```rust
            #[doc = "If apb_pbyte_en\\[3:0\\]
= 1, a single byte is written/read to/from the TX/RX fifos.nElse, If apb_pbyte_en\\[3:0\\]
= 2, two bytes are written/read to/from the TX/RX fifos.nElse, If apb_pbyte_en\\[3:0\\]
= 3, three bytes are written/read to/from the TX/RX fifos.nElse, four bytes are written/read to/from the TX/RX fifos."]
```

robamu:
SiliconLabs-SIM3L1x8_SVD fails.

> By the way why CI does not show SVD-file URL?

See https://github.com/rust-embedded/svd2rust/pull/914

burrbull (approved):

PR #909, opened 2025-02-08
Add `mtvec_align` for RISC-V

### LINKED ISSUE #901: RISC-V Settings file: add `mtvec-align` field
Motivated by https://github.com/rust-embedded/riscv/pull/259

The RISC-V standard states that, in vector mode, the value in `mtvec` must be at least 4-byte aligned. However, targets may impose more strict alignment constraints (e.g., 256-byte aligned in Ibex).

The linked PR adds a `RISCV_MTVEC_ALIGN ` environment variable that PACs should set in their build script to the desired value. I propose using `svd2rust`'s config file to allow PACs to use this feature.

romancardenas:
PACs can now define the required vector table alignment in vector mode. You can check the resulting generated code [here](https://github.com/riscv-rust/e310x/commit/67907eab4137603f3af145708b1e42c6bd82e8d6) for the E310x chip.

This feature has not yet been published in `riscv-rt`'s `crates.io` release but will be soon. Adding it to `svd2rust` before `riscv-rt` should not lead to any malfunctioning.

Closes #901


src/util.rs
```diff
@@ -399,7 +399,9 @@ impl U32Ext for u32 {
     }
 }
 
-pub fn build_rs() -> TokenStream {
+pub fn build_rs(settings: &crate::Settings) -> TokenStream {
```
burrbull, line 402: maybe it makes sense to pass entire `config` to avoid breaking changes in future?

romancardenas: Done!

src/config.rs
```diff
@@ -345,6 +351,13 @@ impl Settings {
             self.riscv_config = source.riscv_config;
         }
     }
+
+    pub fn extra_build(&self) -> Option<TokenStream> {
+        match self.riscv_config.as_ref() {
```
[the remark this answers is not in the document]
romancardenas: Not exactly `map`, but I changed it to `and_then`

burrbull (approved):

Issue #902, opened 2025-01-28
proposal: separate crates for peripherals


rmsyn:
# Split peripheral modules into crates

This proposal is to add a CLI option to split peripheral modules into separate crates in a project workspace.

## Motivation

For large PAC crates, compilation on machines with less-powerful cores is excruciatingly slow. The bottleneck for any projects depending on the PAC crate is obviously the PAC, because one core is pinged to full usage, while all other cores are idle.

By splitting the PAC into separate crates for each peripheral, each core could work on a separate peripheral crate. The separate crates could also improve incremental compilation, since a change in one of the peripheral crates would not necessarily require recompilation of other peripheral crates.

## Challenges

Much of the plumbing is already there for creating separate crates for each peripheral.

Each peripheral already generates its own module.

Peripheral versioning would inherit from the top-level `Cargo.toml` version.

The main challenge will be making minimal changes to enable the code generation, without adding too much to maintainability burden.

## Example

An example project structure might look like:

```
- Cargo.toml
- <soc>-generic
  - Cargo.toml
  - src
    - lib.rs
- <soc>-i2c
  - Cargo.toml
  - src
    - lib.rs
    - i2c0
    - i2c1
    - ...
    - i2c<N>
- <soc>-spi
  - Cargo.toml
  - src
    - lib.rs
    - spi0
    - spi1
    - ...
    - spi<N>
```

## Alternative approaches

Once the [parallel front-end](https://github.com/rust-lang/rust-project-goals/issues/121) is stable in `rustc`, the main motivation for this proposal may be less relevant.

However, there will still be users stuck on older versions of `rustc` where this proposal will be useful.

burrbull:
I like the idea. I think it is last chunk of the puzzle.
But it requires more changes then an option.

1. generic` part must be moved to independent module. (what name to use?). This also makes many other things easier.
2. Finish this https://github.com/rust-embedded/svd2rust/pull/430
 
Other notes:
1. `groupName` should be used for crate names and existing "feature-group" option.
2. For folder names hash values could be used instead of peripheral names.
 With `#[path="hash"] mod spi1` attribute. Like in https://github.com/djmcgill/form/pull/24. Advantages:
- Allows to reuse identical peripherals across different chips.
- You can skip writting (or even generating if hash is calculated from SVD directly) peripheral if folder already exists.

burrbull:
So I've tried to implement 1 step but understood that I don't know how to do this. There are 2 issues:
[...]

PR #922, opened 2025-03-11
[#experiment ]Associated Reader/Writer

burrbull:
Related to #902

@rmsyn This is required step to make register write methods generic over crates: https://github.com/burrbull/reg

But it is slower to compile:
For example stm32f4 crate on old PC:
Before PR: 6.92 s
After PR: 10.23 s

### DIFF
--- a/src/generate/generic.rs
+++ b/src/generate/generic.rs
@@ -104,10 +104,23 @@ pub trait FieldSpec: Sized {
 /// Marker for fields with fixed values
 pub trait IsEnum: FieldSpec {}
 
+#[doc(hidden)]
+pub trait FromBits<U> {
+    unsafe fn from_bits(b: U) -> Self;
+}
+
+#[doc(hidden)]
+pub trait ToBits<U> {
+    fn to_bits(&self) -> U;
+}
+
 /// Trait implemented by readable registers to enable the `read` method.
 ///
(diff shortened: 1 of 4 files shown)


src/generate/generic.rs
```diff
@@ -104,17 +104,33 @@ pub trait FieldSpec: Sized {
...
 
+#[doc(hidden)]
+pub trait FromBits<U> {
+    unsafe fn from_bits(b: U) -> Self;
+}
+
+#[doc(hidden)]
+pub trait ToBits<U> {
+    fn to_bits(&self) -> U;
+}
```
rmsyn, line 115: If these traits are not intended to be used externally, maybe we can implement `Readable` without the extra layer of trait resolution?

I'm thinking along the lines of creating a helper macro to implement the `Readable` trait on the needed types.

rmsyn (commented):
> But it is slower to compile:
> For example stm32f4 crate on old PC:
> Before PR: 6.92 s
> After PR: 10.23 s

Not completely sure what created the drastic increase in compile times. However, maybe removing the extra trait resolution will bring it back down?

src/generate/generic.rs
burrbull: I forgot about `inline`.
Now it is 7.94 s

PR #933, opened 2025-05-14
Various attributes and extern "C" must be marked "unsafe" to compile generated code

Plaba:
tested against [RA4M1 SVD](https://github.com/nguyentri/ra-pac/blob/main/svd/R7FA4M1AB.svd)

tested on rustc 1.86.0 svd2rust 0.36.1

### DIFF
--- a/src/generate/device.rs
+++ b/src/generate/device.rs
@@ -291,10 +291,10 @@ pub fn render(d: &Device, config: &Config, device_x: &mut String) -> Result<Toke
     }
 
     out.extend(quote! {
-        // NOTE `no_mangle` is used here to prevent linking different minor versions of the device
+        // NOTE `unsafe(no_mangle)` is used here to prevent linking different minor versions of the device
         // crate as that would let you `take` the device peripherals more than once (one per minor
         // version)
-        #[no_mangle]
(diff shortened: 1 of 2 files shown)


Emilgardis:
This is due to edition change, use 2021 edition in Cargo.toml to resolve. I remember an issue where we discussed how to solve this but I can't find it anymore

burrbull:
> but I can't find it anymore

https://github.com/rust-embedded/svd2rust/pull/891

1.82.0 Released on: 17 October, 2024

jannic:
> Should we bump rust to 1.82 and fix this or stay on `edition 2021`

Edition 2024 needs rust 1.85.

burrbull:
Ok. Another suggestion. Support 2 rust versions.

What do you think about support by default last stable version of Rust?
But also add `--rust-1.76` option which allows to generate code acceptable by old compilers?

BartMassey:
@burrbull We did discuss adding a `--edition=2024` flag to `svd2rust` (with some earlier Rust version as default); also adding `cfg(edition="2024")` flags to the generated code. The consensus was that things are working OK as-is as far as we know, so we chose not  to put in the work yet.

PR #940, opened 2025-06-25
Make the interrupt table no_mangle on Xtensa

MabezDev:
This was an oversight discovered in https://github.com/esp-rs/esp-hal/issues/2589


burrbull:
update changelog, please

PR #947, opened 2025-09-20
Add unsafe block around `DEVICE_PERIPHERALS = true` in `Peripherals::steal()`

kurtjd:
In edition 2024, `#[warn(unsafe_op_in_unsafe_fn)]` is [enabled by default](https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html).

This PR adds the unsafe block around mutating `DEVICE_PERIPHERALS` in the generated `Peripherals::steal()` function. I had considered only adding this if the `--edition=2024` flag was passed, but there is no harm in this being here even for earlier editions and is IMO clearer regardless. Though if folks disagree I can make that change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+- Change `feature(doc_auto_cfg)` to `feature(doc_cfg)` to allow nightly docs to build.
 - Add unsafe block around `DEVICE_PERIPHERALS = true` in `Peripherals::steal()`
   to support Rust 2024 edition.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+- Add unsafe block around `DEVICE_PERIPHERALS = true` in `Peripherals::steal()`
+  to support Rust 2024 edition.
+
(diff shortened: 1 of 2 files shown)


burrbull:
> I can make that change

Make this, please.

kurtjd:
> > I can make that change
> 
> Make this, please.

Done!

burrbull (dismissed):

kurtjd:
Hm CI appears to be failing for reasons unrelated to these changes. Would you like me to make the fix as part of this PR anyway?

burrbull:
Yes, try to fix it, please.

kurtjd:
> Yes, try to fix it, please.

Should be fixed now. Looks like this PR: https://github.com/rust-lang/rust/pull/138907 merged `doc_auto_cfg` into `doc_cfg` and nightly no longer supports the old name, so I updated this to generate PACs with the new name. This allows docs to be built with nightly, now allowing CI to pass.

burrbull (approved):

kurtjd:
So svd2rust-regress completed locally for me with no errors, but during CI looks like it's failing because clippy is not installed. I'll see if I can update the workflow to ensure clippy is available during CI.

burrbull:
cc @Emilgardis

kurtjd:
Looks like CI is passing now! Please let me know if there are any other changes you would like me to make.

burrbull:
> Please let me know if there are any other changes you would like me to make.

You could prepare PR for `0.37.1` if you want faster release.

PR #952, opened 2026-01-09
Fix cluster info size calculation.

wt:
This fixes generation of a pac for the LPC55S16 chip. Before, trying to generate this board's pac would result in the following error:

```
ERROR svd2rust] Error rendering device
    
    Caused by:
        0: can't render peripheral 'I2S0', group 'I2S'
        1: Could not expand register or cluster block
        2: can't expand cluster 'SECCHANNEL[%s]'
        3: Cluster SECCHANNEL[%s] has size 24928 bits that is more then array increment 256 bits
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+- Fix the cluster info size calculation.
+
 ## [v0.37.1] - 2025-10-17
 
 - Change `feature(doc_auto_cfg)` to `feature(doc_cfg)` to allow nightly docs to build.
--- a/src/generate/peripheral.rs
+++ b/src/generate/peripheral.rs
@@ -913,7 +913,7 @@ fn cluster_info_size_in_bits(info: &ClusterInfo, path: &BlockPath, config: &Conf
                     .map(|rbf| rbf.size)
                     .sum();
 
-                (reg.address_offset * BITS_PER_BYTE) + reg_size
+                reg_size
             }
             RegisterCluster::Cluster(clust) => {
                 (clust.address_offset * BITS_PER_BYTE) + cluster_size_in_bits(clust, path, config)?


Emilgardis:
this definitely results in wrong output. Can you share the SVD file?

image

wt:
Here's the SVD xml file: 
[LPC55S16.xml](https://github.com/user-attachments/files/24535531/LPC55S16.xml)

Emilgardis:
The `addressOffset`s are wrong, seems to be 0xC0 too large

Emilgardis:
going to close this PR as it's not correct and seems like the input is the source of the problem experienced.

wt:
Oh, I think I see what you are saying. Thanks for the help.

Emilgardis:
to clarify :) this seems to work, not sure it's correct though as I was unable to find register descriptions

```diff
diff --git "a" "b"
index 698e1ca..dd099bd 100644
--- "a"
+++ "b"
@@ -54251,7 +54251,7 @@ SPDX-License-Identifier: BSD-3-Clause
           <register>
             <name>PCFG1</name>
             <description>Configuration register 1 for channel pair</description>
-            <addressOffset>0xC20</addressOffset>
+            <addressOffset>0x0</addressOffset>
             <size>32</size>
             <access>read-write</access>
             <resetValue>0</resetValue>
@@ -54300,7 +54300,7 @@ SPDX-License-Identifier: BSD-3-Clause
           <register>
             <name>PCFG2</name>
             <description>Configuration register 2 for channel pair</description>
-            <addressOffset>0xC24</addressOffset>
+            <addressOffset>0x4</addressOffset>
             <size>32</size>
             <access>read-write</access>
             <resetValue>0</resetValue>
@@ -54318,7 +54318,7 @@ SPDX-License-Identifier: BSD-3-Clause
           <register>
             <name>PSTAT</name>
             <description>Status register for channel pair</description>
-            <addressOffset>0xC28</addressOffset>
+            <addressOffset>0x8</addressOffset>
             <size>32</size>
             <access>read-write</access>
             <resetValue>0</resetValue>
```

PR #954, opened 2026-04-17
Add `--skip-peripherals-struct` flag

MabezDev:
## Summary

Adds a new CLI flag `--skip-peripherals-struct` that skips generation of:

- the `Peripherals` struct (and its `#[allow(non_snake_case)]` fields),
- its `take`/`steal` impls,
- the `DEVICE_PERIPHERALS` static (and its `#[no_mangle]` attribute).

## Motivation

Some downstream PAC consumers manage their own `Peripherals` struct / singleton instead of using the one emitted by svd2rust. In that case the svd2rust-generated `DEVICE_PERIPHERALS` symbol is not just dead weight — because it's `#[no_mangle]`, it's a global symbol that can collide across crates and prevents Cargo from linking two minor versions of a PAC in the same binary (which is the whole reason the `#[no_mangle]` is there).

Concrete example: esp-pacs manages its own `Peripherals` in esp-hal and ran into exactly this problem. See [esp-rs/esp-pacs#427](https://github.com/esp-rs/esp-pacs/pull/427), which deletes ~1700 lines of hand-maintained overrides (custom `lib.rs` shims per chip) once svd2rust can be told not to emit the struct at all.

## AI Usage

Generative AI assisted the development of this PR, however I have carefully vetted and amended the code and believe it to be of decent quality. Review welcome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+- Add `--skip-peripherals-struct` flag to skip generating the `Peripherals`
+  struct, its `take`/`steal` impl and the `DEVICE_PERIPHERALS` static
+
 ## [v0.37.1] - 2025-10-17
 
 - Change `feature(doc_auto_cfg)` to `feature(doc_cfg)` to allow nightly docs to build.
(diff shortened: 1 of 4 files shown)


burrbull:
rebase, please

MabezDev:
Superseded by #955, which is based off current master rather than the 0.37.1 release tag.

PR #956, opened 2026-04-18
Reg::prepare

burrbull:
Related discussion https://github.com/rust-embedded/svd2rust/issues/708#issuecomment-2833412665

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,11 +126,11 @@ jobs:
           cargo install svd2rust --path .
 
       # Install the MSRV toolchain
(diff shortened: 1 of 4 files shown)


jannic:
I like this. In the past, I did crimes like https://github.com/jannic/rp2040-pio-logic-analyzer/blob/af9fe29443b91858835bd76785c78b9d1ee82a28/src/lib.rs#L238 to get the same result. Avoiding the transmute would be nice.

jannic (dismissed):

burrbull:
@jannic changelog conflict. Need reapprove.

jannic (approved):

adamgreig (approved):

PR #957, opened 2026-04-21
Fix escape char

Shriyans406:
This PR ensures that characters like &, <, and > are escaped or avoided in SVD descriptions so that they don't break generated Rust docs

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -15,6 +15,10 @@ use syn::{
     punctuated::Punctuated, token::PathSep, Lit, LitInt, PathArguments, PathSegment, Type, TypePath,
 };
 
+use regex::Regex;
+use std::sync::OnceLock;
+
+
 use anyhow::{anyhow, Result};
 
 pub const BITS_PER_BYTE: u32 = 8;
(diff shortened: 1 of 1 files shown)


src/util.rs
```diff
@@ -179,11 +179,24 @@ pub fn escape_brackets(s: &str) -> String {
...
-        s.into()
+    if !s.contains('[') && !s.contains('&') && !s.contains('<') && !s.contains('>'){
+        return s.into();
+    }
+    let mut escaped = s.to_string();
+    if escaped.contains('&') {
+        escaped = escaped.replace('&', "&amp;");
+    }
+    if escaped.contains('<') {
+        escaped = escaped.replace('<', "&lt;");
```
burrbull: svd string can contain valid html tags. How do you propose to process such cases?

Shriyans406:
Thank you for the feedback! Escaping of < and > would break HTML formatting
in the SVD.

I've updated the implementation to be more selective, it now uses a regex
to identify and preserve HTML tags (like <b>) and entities (like &amp;),
while still escaping characters (like & and < that aren't part of a tag) to
prevent them from being misinterpreted.

I've also added unit tests covering these cases.
Is it fine?

burrbull:
I don't see any changes in this PR.

Shriyans406:
I think I made a mistake, so please check it sir.

Regards

PR #958, opened 2026-04-24
Ci/add wch riscv test

Shriyans406:
This PR expands risc-v test area by adding the WCH CH32V30x chip to the regression suite. It introduced the WCH manufacturer to the test harness and also includes a new test case in tests.yml, which has been successfully verified locally through the regression tool.


burrbull:
Separate this PR from #957, please.

PR #959, opened 2026-04-25
use cortex-m-types crate for the InterruptNumber trait

robamu:
this is a  part of the effort to make updating `cortex-m` easier. PACs can now depend on `cortex-m-types` instead of `cortex-m`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,11 +125,11 @@ jobs:
           cargo install svd2rust --path .
 
       # Install the MSRV toolchain
-      - uses: dtolnay/rust-toolchain@1.76.0
+      - uses: dtolnay/rust-toolchain@1.81.0
       - name: Run reression tool with MSRV
         # The MSRV only applies to the generated crate. The regress tool should still be run with
         # stable.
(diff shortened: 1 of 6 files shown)


burrbull:
Makes sense. Could you add discussion link in first message, please?

robamu:
which discussion link? this PR?

robamu:
Also had to bump the MSRV.

burrbull:
> which discussion link?

Where `cortex-m-types` was introduced.

burrbull (approved):

robamu:
I can not find it. maybe because `cortex-m` was not always a mono-repo?

adamgreig:
Might be worth a note that PACs generated using `cortex-m-types` won't export an `InterruptNumber` that works with current `cortex-m`'s NVIC (https://docs.rs/cortex-m/latest/cortex_m/peripheral/struct.NVIC.html). I think we can publish a new cortex-m 0.7.x that has a blanket impl of its `InterruptNumber` trait for the one from `cortex-m-types` that would maintain compatibility, but need to check that actually works.

PR #960, opened 2026-04-25
Strict safe access

Shriyans406:
WRT to refinement suggested by the other contributor.


burrbull:
Please, leave commits that related only to PR. Other commits remove or move to appropriate PRs.

PR #962, opened 2026-05-03
add back optional cortex-m interrupt number impl generation

### LINKED ISSUE #961: Add back generation of cortex_m InterruptTypes
I changed the generated PACs to only implement `cortex_m_types::InterruptNumber` in https://github.com/rust-embedded/svd2rust/pull/959.

This will become a problem once PACs are generated because they will not be compatible with various other important crates in the ecosystem, which still use the `cortex_m` `InterruptNumber` trait. I will prepare a PR which adds back the generation of the old trait, so both are implemented for now. I will also add an option to suppress the generation of this old trait.

Once `cortex-m` v0.8/v1 hits, we could omit the implementation by default and change, the option to instead add the old implementation. maybe there is an even better way?

robamu:
Fixes #961

### DIFF
--- a/svd2rust-regress/src/svd_test.rs
+++ b/svd2rust-regress/src/svd_test.rs
@@ -18,6 +18,7 @@ const CRATES_ALL: &[&str] = &[
 const CRATES_MSP430: &[&str] = &["msp430 = \"0.4.0\"", "msp430-rt = \"0.4.0\""];
 const CRATES_ATOMICS: &[&str] =
     &["portable-atomic = { version = \"1\", default-features = false }"];
+const CRATES_CORTEX_M_LEGACY_INT_NUM: &[&str] = &["cortex-m = { version = \"0.7\" }"];
 const CRATES_CORTEX_M: &[&str] = &["cortex-m-types = \"0.1\"", "cortex-m-rt = \"0.7\""];
 const CRATES_RISCV: &[&str] = &["riscv = \"0.12.1\"", "riscv-rt = \"0.13.0\""];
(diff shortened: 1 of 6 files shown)


burrbull:
As I understand `cortex_m::interrupt::InterruptNumber` is deprecated? Is not it?

In this case generation should be disabled by default to prevent yet one breaking change in future.

robamu:
See https://github.com/rust-embedded/svd2rust/issues/961 . The old trait will only be deprecated when v0.8 of cortex-m is released. What we could do however is to already make the generation optional and additive, so use a `add-cortex-m-int-num` instead of `skip-cortex-m-int-num`

burrbull:
> use a `add-cortex-m-int-num` instead of `skip-cortex-m-int-num`

Yes, this is what I meant.

You should also check this feature in CI and optionally add the dependency.

burrbull (dismissed):

svd2rust-regress/src/svd_test.rs
```diff
@@ -432,6 +433,8 @@ impl TestCase {
             .chain(if let Some(opts) = opts {
                 if opts.iter().any(|v| v.contains("atomics")) {
                     CRATES_ATOMICS.iter()
+                } else if opts.iter().any(|v| v.contains("add-cortex-m-int-num")) {
```
burrbull: It is better another `chain` instead of `else if` here.

burrbull (approved):

burrbull:
Could you also describe flag in `lib.rs` docs? Forgot about this.

Issue #966, opened 2026-08-23
Odd (and incorrect) code generation from SVD


MsFormula:
I was in the process of updating a project that I am working on to new SVD's and when I generated the PAC I got some strange code. This first block is and excerpt from the relevant part of the SVD. The full thing can be found [here](https://github.com/adafruit/broadcom-peripherals/blob/4c890dbdc8260066fab6a18739b3c6ca67630a45/svd/peripherals/bcm_aux.svd.jinja#L316):
```
<field>
    <name>CTS_ASSERT</name>
    <description>CTS assert level</description>
    <bitOffset>7</bitOffset>
    <bitWidth>1</bitWidth>
    <enumeratedValues>
        <name>ASSERT_LEVEL</name>
        <enumeratedValue>
            <name>HIGH</name>
            <description>Assert high</description>
            <value>0</value>
        </enumeratedValue>
        <enumeratedValue>
            <name>LOW</name>
            <description>Assert low</description>
            <value>1</value>
        </enumeratedValue>
    </enumeratedValues>
</field>
<field>
    <name>RTS_ASSERT</name>
    <description>RTS assert level</description>
    <bitOffset>6</bitOffset>
    <bitWidth>1</bitWidth>
    <enumeratedValues derivedFrom="UART1.CNTL.CTS_ASSERT.ASSERT_LEVEL"></enumeratedValues>
</field>
```

Generated code is as follows:
```
#[doc = "RTS assert level"]
pub use AssertLevel;
#[doc = "Field `RTS_ASSERT` reader - RTS assert level"]
pub use CtsAssertR as RtsAssertR;
#[doc = "Field `RTS_ASSERT` writer - RTS assert level"]
pub use CtsAssertW as RtsAssertW;
#[doc = "CTS assert level\n\nValue on reset: 0"]
#[cfg_attr(feature = "defmt", derive(defmt::Format))]
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub enum AssertLevel {
    #[doc = "0: Assert high"]
    High = 0,
    #[doc = "1: Assert low"]
    Low = 1,
}
impl From<AssertLevel> for bool {
    #[inline(always)]
    fn from(variant: AssertLevel) -> Self {
        variant as u8 != 0
    }
}
#[doc = "Field `CTS_ASSERT` reader - CTS assert level"]
pub type CtsAssertR = crate::BitReader<AssertLevel>;
impl CtsAssertR {
[...]

PR #967, opened 2026-08-23
sort-fields flag

burrbull:
cc #966 @MsFormula

Try this build with `--sort-fields=reverse` flag.


MsFormula:
@burrbull I don't know if you saw my response on my issue but this worked great. Thank you.

